PR #102, opened 2013-08-29
Add missing NULL reset after free()

kou:
If the next `GRN_MALLOC()` fails, `dv[0].data` has dangling pointer. It should be reset after `free()`.


kou:
Reject because the pointer is reassigned immediately. (only when `GRN_MALLOC` is succeeded...)

PR #152, opened 2014-02-04
Use grn_obj_get_value() in grn_expr_exec()

kou:
grn_expr_exec() uses grn_obj_get_value_() that returns a pointer to
the value and value size. So we can't get value that is not owned by
object with grn_obj_get_value_(). GRN_COLUMN_INDEX returns estimate
size as value and it is not owned value. In other words, we can't use
grn_obj_get_value_() for GRN_COLUMN_INDEX.

grn_obj_get_value() writes value to the passed grn_obj. So
grn_obj_get_value() can returns not owned value.

grn_obj_get_value_() is lighter API rather than grn_obj_get_value()
because grn_obj_get_value_() is just returns a pointer. It doesn't
write the value. Maybe, it is the reason why grn_expr_exec() uses
grn_obj_get_value_().

But grn_expr_exec() always writes value returned from
grn_obj_get_value_(). There is no speed merit. So we can use
grn_obj_get_value() rather than grn_obj_get_value_() without speed
down.

We can get value of GRN_COLUMN_INDEX by using grn_obj_get_value().

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -5003,11 +5003,13 @@ grn_accessor_get_value(grn_ctx *ctx, grn_accessor *a, grn_id id, grn_obj *value)
   uint32_t vs = 0;
   uint32_t size0;
   void *vp = NULL;
+  grn_id original_domain;
   if (!value) {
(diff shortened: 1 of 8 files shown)


kou:
Reject because it's out of date.

PR #195, opened 2014-08-21
Make grn_obj_cast() for table with nonexistent key failure

kou:
The current behavior doesn't report any error and not set any value.
Users should check both return value of grn_obj_cast() and
GRN_BULK_VSIE(dest) is increased.

With this change, users only check return value of grn_obj_cast().


kou:
Reject because the same change is already included.

PR #276, opened 2015-01-30
Make sure to use the header files of onigmo in vendor.

knu:
Groonga currently includes their header files with <>, and lists the -I flags too late, so the build shall fail if the system has an incompatible version of onigmo/oniguruma or mruby headers installed.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -11,12 +11,12 @@ AM_CFLAGS =					\
 	$(GRN_CFLAGS)				\
 	$(MESSAGE_PACK_CFLAGS)			\
 	$(MRUBY_CFLAGS)				\
-	$(ONIGMO_CFLAGS)			\
 	$(LIBLZ4_CFLAGS)
 
 DEFAULT_INCLUDES =				\
 	-I$(top_builddir)			\
-	-I$(top_srcdir)/include
+	-I$(top_srcdir)/include			\
+	$(ONIGMO_CFLAGS)
 
 DEFS += -D_REENTRANT $(GRN_DEFS) -DGRN_DAT_EXPORT
 


kou:
Thanks for your report!

But I can't reproduce it by creating `/usr/include/mruby.h` with the following content:

```c
#error xxx
```

Should I confirm it on FreeBSD? I confirmed it on Debian GNU/Linux.

Could you show output of `make V=1`?

knu:
I've experienced a failure on a host where Oniguruma 4 is installed in /usr/local when building groonga with --prefix=/usr/local.
```
libtool: compile:  cc -DHAVE_CONFIG_H -D_REENTRANT -DGRN_DLL_FILENAME=L\"libgroonga-0.dll\" -DGRN_PLUGINS_DIR=\"/usr/local/lib/groonga/plugins\" -DGRN_RELATIVE_PLUGINS_DIR=\"lib/groonga/plugins\" -DGRN_RUBY_SCRIPTS_DIR=\"/usr/local/lib/groonga/scripts/ruby\" -DGRN_RELATIVE_RUBY_SCRIPTS_DIR=\"lib/groonga/scripts/ruby\" -DGRN_LOG_PATH=\"/var/log/groonga.log\" -DGRN_DEFAULT_DOCUMENT_ROOT=\"/usr/local/share/groonga/html/admin\" -DGRN_DEFAULT_RELATIVE_DOCUMENT_ROOT=\"share/groonga/html/admin\" -DGRN_DAT_EXPORT -I.. -I../include -I/usr/local/include -fno-strict-aliasing -I/usr/local/include -I../vendor/onigmo-source -O2 -pipe -g -fno-omit-frame-pointer -march=core2 -fstack-protector -fno-strict-aliasing -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-pointer-sign -Wno-missing-field-initializers -Wformat=2 -Wstrict-aliasing=2 -Wdisabled-optimization -Wfloat-equal -Wpointer-arith -Wbad-function-cast -Wcast-align -Wredundant-decls -Wwrite-strings -MT util.lo -MD -MP -MF .deps/util.Tpo -c util.c -o util.o >/dev/null 2>&1
...
ii.c:6025:9: error: use of undeclared identifier 'OnigPosition'
        OnigPosition position;
        ^
ii.c:6039:9: error: use of undeclared identifier 'position'
        position = onig_search(regex,
        ^
ii.c:6046:13: error: use of undeclared identifier 'position'
        if (position != ONIG_MISMATCH) {
            ^
12 warnings and 3 errors generated.
Makefile:712: recipe for target 'ii.lo' failed
gmake[6]: *** [ii.lo] Error 1
...
```
As you can see above, `-I/usr/local/include` appears earlier than `-I../vendor/onigmo-source` and I have `/usr/local/include/oniguruma.h` installed by Oniguruma, which is thus picked by the compiler only to fail.

`#include <oniguruma.h>` should be changed to `#include "oniguruma.h"` in case you have `/usr/include/oniguruma.h`.

knu:
Sorry, I don't currently have the time to investigate on mruby and the conflict with onigmo/oniguruma is my only concern right now, so I've reverted the mruby part.  Will revisit when I have mruby at hand.

kou:
Thanks.

(If you can't understand my English, I'll explain it in Japanese...)

It seems that change of `lib/Makefile.am` is only needed. I think that `#include "oniguruma.h"` isn't needed because `oniguruma.h` doesn't exist at `lib/`. If we want to use `#include "..."` style, we should use `#include "../vendor/onigmo-source/oniguruma.h"`. (I should write a link to difference between `"..."` and `<...>` but I can't find the original text...)

And do you set `INCLUDES=-I/usr/local/include` environment variable? It may cause the problem. (Anyway, I think that your `lib/Makefile.am` change is needed.)

Details:

`lib/Makefile` has the following:

```makefile
LTCOMPILE = $(LIBTOOL) $(AM_V_lt) --tag=CC $(AM_LIBTOOLFLAGS) \
        $(LIBTOOLFLAGS) --mode=compile $(CC) $(DEFS) \
        $(DEFAULT_INCLUDES) $(INCLUDES) $(AM_CPPFLAGS) $(CPPFLAGS) \
        $(AM_CFLAGS) $(CFLAGS)
# ...
DEFAULT_INCLUDES = \
        -I$(top_builddir)                       \
        -I$(top_srcdir)/include
# ...
AM_CFLAGS = \
        $(NO_STRICT_ALIASING_CFLAGS)            \
        $(COVERAGE_CFLAGS)                      \
        $(GRN_CFLAGS)                           \
        $(MESSAGE_PACK_CFLAGS)                  \
        $(MRUBY_CFLAGS)                         \
        $(ONIGMO_CFLAGS)
```

It seems that `-I/usr/local/include` in `-I../include -I/usr/local/include -fno-strict-aliasing` exists between `$(DEFAULT_INCLUDES)` and `$(AM_CFLAGS)`. Because `-I../include` is `-I$(top_srcdir)/include` in `$(DEFAULT_INCLUDES)` and `-fno-strict-aliasing` is `$(NO_STRICT_ALIASING_CFLAGS)` in `$(AM_CFLAGS)`.

If changes for `#include "..."` aren't needed, could you remove the changes from your pull request? Then I'll merge this pull request.

knu:
Ah, of course, I'll do that when I get back to my console.

kou:
Thanks!

knu:
Done!

kou:
Thanks!
I've merged it!

knu:
@kou Thanks!

I didn't set INCLUDES; `-I/usrl/local/include` is in `CPPFLAGS`, and moving ONIGMO_CFLAGS from AM_CFLAGS to DEFAULT_INCLUDES has made it come earlier than CPPFLAGS.

kou:
Ah, `CPPFLAGS`!

PR #284, opened 2015-02-21
doc release: don't fold a line

myokoym:
Because this command must be executed on a single line.

Or we should put a backslash the end of the first line:

    % sudo apt-get install -V debootstrap createrepo rpm \
    mercurial python-docutils python-jinja2 ...


kou:
Thanks!

The description is old... It should be updated...

PR #287, opened 2015-02-23
Drop VS 2010 support

cosmo0920:


cosmo0920:
Thank you for merging!

PR #294, opened 2015-02-27
Change default protocol from GQTP to HTTP

kenhys:
This change allows to omit "--protocol http" option


kou:
What do you think about backward compatibility?

kou:
Any response?
How should I handle this pull request?

kenhys:
When sending this PR, I thought that it is reasonable change for most of Groonga user because there is a few case which use GQTP. Actually it breaks backward compatibility, I know, so, I agree that benefit of keeping compatibility is important than changing it.

PR #301, opened 2015-03-11
Support position skip in offline index construction

kou:
Offline index construction assumes that position is increased lineally. But `GRN_TOKEN_SKIP_WITH_POSITION` breaks the assumption.

This patch stores position into `grn_ii_buffer.block_buf` like weight.


kou:
`0x20000000`は`GRN_ID_MAX`とかぶるからダメだなぁ。

kou:
Another implementation: #318

PR #317, opened 2015-03-16
Use . (dot) rather than source to be compat with POSIX.2.

kuriyama:
bash's "source" command extension is not in POSIX.2 standard.  Use
. (dot) command to be worked on non-bash environment (like FreeBSD).


kou:
Thanks!
It's reasonable.

PR #327, opened 2015-03-26
Show warning message if --protocol is not specified

kenhys:
The default protocol is still GQTP because it breaks backward compatibility.
After a while, we will change the default protocol to HTTP in the future
release.


kou:
The concept is acceptable but the implementation isn't acceptable. Because "alone" is used by another meaning in the file. "Alone" is used as "stand alone mode" in the file.

PR #345, opened 2015-05-31
debian: Use network-online.target to wait configured network strictly

cosmo0920:
I think that using `Wants=network-online.target` is more suitable for Groonga's server functionality.

Because systemd's manpage says:

    network-online.target is a target that actively waits until the nework is "up", 
    where the definition of "up" is defined by the network management software. 
    Usually it indicates a configured, routable IP address of some kind.

ref: http://www.freedesktop.org/software/systemd/man/systemd.special.html#network-online.target

### DIFF
--- a/packages/debian/groonga-httpd.service
+++ b/packages/debian/groonga-httpd.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Groonga HTTP Server
-After=network.target
+Wants=network-online.target
 Conflicts=groonga.service
 
 [Service]
--- a/packages/debian/groonga-httpd.service
+++ b/packages/debian/groonga-httpd.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Groonga HTTP Server
 Wants=network-online.target
+After=network-online.target
 Conflicts=groonga.service
 
 [Service]
--- a/packages/debian/groonga-server-gqtp.service
+++ b/packages/debian/groonga-server-gqtp.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Groonga GQTP Server
-After=network.target
+Wants=network-online.target
 
 [Service]
 Type=forking
--- a/packages/debian/groonga-server-gqtp.service
+++ b/packages/debian/groonga-server-gqtp.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Groonga GQTP Server
 Wants=network-online.target
+After=network-online.target
 
 [Service]
 Type=forking


cosmo0920:
# Result

Here is test operations using `Wants=network-online.target` in Unit.

## groonga-server-gqtp:

rewrite (or apply patch) Unit as follows:

```ini
[Unit]
Description=Groonga GQTP Server
Wants=network-online.target

[Service]
Type=forking
EnvironmentFile=-/etc/default/groonga-server-gqtp
User=groonga
Group=groonga
ExecStart=/usr/bin/groonga --protocol gqtp -d --pid-path /var/run/groonga/groonga-gqtp.pid --bind-address $ADDRESS --log-path $LOG_PATH --query-log-path $QUERY_LOG_PATH --port $PORT $DATABASE

[Install]
WantedBy=multi-user.target
```

```bash
$ sudo systemctl daemon-reload
$ sudo systemctl status groonga-server-gqtp
● groonga-server-gqtp.service - Groonga GQTP Server
   Loaded: loaded (/lib/systemd/system/groonga-server-gqtp.service; disabled)
   Active: activating (start) since Wed 2015-06-24 14:11:52 UTC; 8s ago
  Control: 14540 (groonga)
   CGroup: /system.slice/groonga-server-gqtp.service
           └─14540 /usr/bin/groonga --protocol gqtp -d --pid-path /var/run/gr...
```

## groonga-httpd

rewrite (or apply patch) Unit as follows:

```ini
[Unit]
Description=Groonga HTTP Server
Wants=network-online.target
Conflicts=groonga.service

[Service]
Type=forking
User=root
Group=root
ExecStartPre=-/bin/mkdir -p /var/log/groonga/httpd
ExecStartPre=-/bin/mkdir -p /var/run/groonga
ExecStart=/usr/sbin/groonga-httpd
ExecStop=/usr/sbin/groonga-httpd -s stop

[Install]
WantedBy=multi-user.target
```

```bash
$ sudo systemctl daemon-reload
$ sudo systemctl status groonga-httpd
● groonga-httpd.service - Groonga HTTP Server
   Loaded: loaded (/lib/systemd/system/groonga-httpd.service; disabled)
   Active: active (running) since Wed 2015-06-24 14:13:19 UTC; 6s ago
  Process: 14577 ExecStart=/usr/sbin/groonga-httpd (code=exited, status=0/SUCCESS)
  Process: 14574 ExecStartPre=/bin/mkdir -p /var/run/groonga (code=exited, status=0/SUCCESS)
  Process: 14573 ExecStartPre=/bin/mkdir -p /var/log/groonga/httpd (code=exited, status=0/SUCCESS)
 Main PID: 14579 (groonga-httpd)
   CGroup: /system.slice/groonga-httpd.service
           ├─14579 nginx: master process /usr/sbin/groonga-httpd
           └─14580 nginx: worker process
vagrant@web-w-2-15-lb:~$ 
```

cosmo0920:
@myokoym Is there anything which bothers you?

myokoym:
Thanks for your confirming!

I think we should also specify `After=network-online.target`. I referenced the following article (Japanese):

>
Unitの依存関係と起動順序は別の概念なので注意が必要です。「Requires/Wants」は、このUnitと同時に起動されるべきUnitを指定しますが、どちらが先に起動するかは限定していません。
http://enakai00.hatenablog.com/entry/20130917/1379374797

What do you think about it?

cosmo0920:
Thanks for your suggestion.
I've confused about Unit itself dependency and service launch order dependency.

> I think we should also specify `After=network-online.target`. I referenced the following article (Japanese):

Oh, I see....
I never imagined that systemd's dependency concept is separated two concepts which are itself dependency and launch order dependency.


> What do you think about it?

I think that we **SHOULD** specify `After=network-online.target`for Unit service launch order dependency. Furthermore, we **SHOULD** specify `Wants=network-online.target` for Unit itself dependency.
It is same as your opinion.

myokoym:
Let's go with that. I've merged it!

cosmo0920:
Thanks!!

PR #353, opened 2015-06-27
Add build and install information in README.md

hyoshiok:
There is no information regarding build and install in README.md
It should have such information.

For examples, the documentation does not have information  cloning from github.
http://groonga.org/docs/install/mac_os_x.html
A developer does not have information how to generate configure file. If README.md has such information, he/she can build it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,13 @@ See doc/source/ directory or http://groonga.org/docs/.
   * Path: vendor/nginx-${VERSION}
   * License: BSD license. See vendor/nginx-${VERSION}/LICENSE for details.
 
+## Build and Install
+
+ $ ./autogen.sh
+ $ ./configure
+ $ make
+ $ sudo make install
+
 ## Authors
 
 ### Primary authors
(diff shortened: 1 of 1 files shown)


kou:
Thanks for the patch.

We have information for building and installing from source at http://groonga.org/docs/install/others.html#build-from-source

We don't want to have duplicated information in README.md and document. If we need to have such information, we'll add just a reference to the install document to README.md. For example, we did it for "Reference manual": https://github.com/groonga/groonga/blob/master/README.md#reference-manual

By the way, I think that it's better that we recommend package install rather than source build install because many users will user package rather than source.

hyoshiok:
Suppose the following work flow
  1) $ git clone https://github.com/groonga/groonga/
  2) $ cd groonga
  3) But there is no information how to build it from github repository

The point is there is no information building from github.

kou:
I think that it should be written under http://groonga.org/docs/contribution.html instead of README.md.

kou:
I've written documents: http://groonga.org/docs/contribution/development/build.html

Issue #241, opened 2014-11-09
Define grn_rc_to_string() like grn_operator_to_string()


kou:
Rroonga has similar function.

s-yata:
746a9ac5f8d4d4ca849b4777591f42fc18f9a9ff added `grn_rc_to_string()`.

PR #365, opened 2015-07-23
Fixed issue #241. Defined grn_rc_to_string()

furandon-pig:
I added grn_rc_to_string(). This function convert the grn_rc enum value to string.
Conduct a test of grn_rc_to_string() is following procedure.

 * test script
```
$ cat test_grn_rc_to_string.gdb 
set height 0

break main
run

echo val=0 is pass a test.\n

# Test SUCCESS code (positive value range).
call strcmp(grn_rc_to_string(0), "success")
call strcmp(grn_rc_to_string(1), "end of data")
call strcmp(grn_rc_to_string(2), "unknown")

# Test ERROR code (negative value range).
call strcmp(grn_rc_to_string(-1),  "unknown error")
call strcmp(grn_rc_to_string(-2),  "operation not permitted")
call strcmp(grn_rc_to_string(-75), "plugin error")
call strcmp(grn_rc_to_string(-76), "scorer error")
call strcmp(grn_rc_to_string(-77), "unknown")

quit
```

 * test result
```
$ gdb --silent -x ./test_grn_rc_to_string.gdb ./groonga/bin/groonga
Reading symbols from ./groonga/bin/groonga...done.
Breakpoint 1 at 0x403a30: file groonga.c, line 2658.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Breakpoint 1, main (argc=1, argv=0x7fffffffe678) at groonga.c:2658
2658	{
val=0 is pass a test.
$1 = 0
$2 = 0
$3 = 0
$4 = 0
$5 = 0
$6 = 0
$7 = 0
$8 = 0
A debugging session is active.

	Inferior 1 [process 26746] will be killed.

Quit anyway? (y or n) [answered Y; input not from terminal]
```

### DIFF
--- a/lib/operator.c
+++ b/lib/operator.c
@@ -124,6 +124,105 @@ grn_operator_to_string(grn_operator op)
   }
 }
 
+static const char *rc_names[] = {
+  "success",
+  "end of data",
+  "unknown error",
+  "operation not permitted",
+  "no such file or directory",
+  "no such process",
+  "interrupted function call",
+  "input output error",
+  "no such device or address",
+  "arg list too long",
+  "exec format error",
+  "bad file descriptor",
+  "no child processes",
+  "resource temporarily unavailable",
+  "not enough space",
+  "permission denied",
+  "bad address",
+  "resource busy",
+  "file exists",
+  "improper link",
+  "no such device",
+  "not a directory",
+  "is a directory",
+  "invalid argument",
+  "too many open files in system",
+  "too many open files",
+  "inappropriate i o control operation",
+  "file too large",
+  "no space left on device",
+  "invalid seek",
+  "read only file system",
+  "too many links",
+  "broken pipe",
+  "domain error",
+  "result too large",
+  "resource deadlock avoided",
+  "no memory available",
+  "filename too long",
+  "no locks available",
+  "function not implemented",
+  "directory not empty",
+  "illegal byte sequence",
+  "socket not initialized",
+  "operation would block",
+  "address is not available",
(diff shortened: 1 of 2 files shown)


myokoym:
Thanks for the patch!
But I'm sorry that I can't confirm it immediately. Could you work on the other issues?

lib/operator.c
```diff
@@ -124,6 +124,105 @@ grn_operator_to_string(grn_operator op)
...
+  "command error",
+  "plugin error",
+  "scorer error",
+};
+
+#define GRN_RC_SUCCESS_CODE_LAST GRN_END_OF_DATA
+#define GRN_RC_ERROR_CODE_LAST   GRN_SCORER_ERROR
+
+const char *
+grn_rc_to_string(grn_rc rc)
```
kou: Writing not `grn_operator` related code in `operator.c` isn't good.
Please create `rc.c`.

lib/operator.c
```diff
@@ -124,6 +124,105 @@ grn_operator_to_string(grn_operator op)
...
+{
+  if (rc >= GRN_RC_ERROR_CODE_LAST && rc < 0) {
+    // negative value range is ERROR code.
```
kou: Please don't use C++ style comment.

kou:
We're using Cutter as a testing framework for this purpose.
GDB isn't good for us as a test tool...

kou:
Close because there is no response.

PR #406, opened 2015-09-19
Fix autogen.sh extream slow.

hiroyuki-sato:
This configure.ac is extreme slow on libtool-2.4.6.

From libtool mailing list.

You can workaround/fix this by:
-m4_define([nginx_version], m4_include(nginx_version))
-NGINX_VERSION=nginx_version
+NGINX_VERSION=m4_include([nginx_version])

See also.
http://lists.gnu.org/archive/html/libtool/2015-09/msg00004.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1459,8 +1459,7 @@ AC_ARG_WITH(groonga-org-path,
 AC_SUBST(GROONGA_ORG_PATH)
 
 # groonga-httpd
-m4_define([nginx_version], m4_include(nginx_version))
-NGINX_VERSION=nginx_version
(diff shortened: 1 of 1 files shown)


cosmo0920:
It's OK for my machine (MacBook 12' @ 1.1 GHz Intel Core M)! :yum: 
This PR makes Very Fast autogen!!!

kenhys:
Thanks.

LGTM too then merged this PR.

```
% brew info libtool
libtool: stable 2.4.6 (bottled)
```

and it makes faster than ever with this PR.

```
./autogen.sh  18.23s user 2.04s system 93% cpu 21.632 total
```

PR #407, opened 2015-09-19
Fix binary operator expected.

hiroyuki-sato:
Please review this patch.

If multiple gettext package installed,
Following statement does not work properly.

if [ -d $gettext_aclocal ]; then
  ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
fi

./autogen.sh: line 12: [: /usr/local/Cellar/gettext/0.19.4/share/aclocal: binary operator expected

echo /usr/local/Cellar/gettext/*/share/aclocal
/usr/local/Cellar/gettext/0.19.4/share/aclocal /usr/local/Cellar/gettext/0.19.5.1/share/aclocal

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -8,9 +8,9 @@ Darwin)
         if [ -d $homebrew_aclocal ]; then
           ACLOCAL_ARGS="$ACLOCAL_ARGS -I $homebrew_aclocal"
         fi
-        gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal)"
-        if [ -d $gettext_aclocal ]; then
-          ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
+        gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal | awk '{ print $NF }')"
+        if [ -d "$gettext_aclocal" ]; then
+          $ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
         fi
 	;;
 FreeBSD)
--- a/autogen.sh
+++ b/autogen.sh
@@ -10,7 +10,7 @@ Darwin)
         fi
         gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal | awk '{ print $NF }')"
         if [ -d "$gettext_aclocal" ]; then
-          $ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
+          ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
         fi
 	;;
 FreeBSD)


autogen.sh
```diff
@@ -8,9 +8,9 @@ Darwin)
         if [ -d $homebrew_aclocal ]; then
           ACLOCAL_ARGS="$ACLOCAL_ARGS -I $homebrew_aclocal"
         fi
-        gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal)"
-        if [ -d $gettext_aclocal ]; then
-          ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
+        gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal | awk '{ print $NF }')"
+        if [ -d "$gettext_aclocal" ]; then
+          $ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
         fi
```
kenhys: It seems that `ACLOCAL_ARGS` must not be expanded in this context.

It should be:
```
ACLOCAL_ARGS="$ACLOCAL_ARGS -I $gettext_aclocal"
```

hiroyuki-sato:
@kenhys 

Thank you for your comment. I fixed it.

autogen.sh
```diff
@@ -8,8 +8,8 @@ Darwin)
         if [ -d $homebrew_aclocal ]; then
           ACLOCAL_ARGS="$ACLOCAL_ARGS -I $homebrew_aclocal"
         fi
-        gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal)"
-        if [ -d $gettext_aclocal ]; then
+        gettext_aclocal="$(echo /usr/local/Cellar/gettext/*/share/aclocal | awk '{ print $NF }')"
```
kou: `ls ... | tail -n 1` will be better to ensure using the latest gettext. (`ls` sorts output by default.)

hiroyuki-sato:
@kou 
Thank you for your comment.

``ls .. | tail -n 1`` is better. 
But  ls command can't get proper version if  source version are like the following.

```
gettext-0.10
gettext-0.9
```

latest version is 0.19.6 so It is not important.

kou:
We can use `sort --version-sort` for the case. :-)
Please confirm https://github.com/groonga/groonga/commit/559fdc7d0be2df982513486e6517fad5a891afdb

hiroyuki-sato:
Thanks!!.

cosmo0920:
FYI: OS X's user land is not GNU but BSD.
So, OS X's sort command does not have `--version-sort` option.

ref: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/sort.1.html

If you use homebrew, `g-` prefixed GNU coreutils command can install like this:

```bash
$ brew install coreutils
$ gsort --help
...
  -V, --version-sort          natural sort of (version) numbers within text
```

kou:
> FYI: OS X's user land is not GNU but BSD.

> ref: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/sort.1.html

The man page says "Copyright (C) 2005 Free Software Foundation, Inc.". Is `sort` on OS X really based on BSD `sort`?

FYI: I confirmed whether BSD `sort` supports `--version-sort` or not on FreeBSD: https://www.freebsd.org/cgi/man.cgi?query=sort&sektion=1

cosmo0920:
> The man page says "Copyright (C) 2005 Free Software Foundation, Inc.". Is sort on OS X really based on BSD sort?

Hmm..., newer OS X replaces several BSD origin commands with GNU's one.

At least, in OS X 10.11, sort command is GNU origin, but this command is too old.

OSX's sort commands says unrecognized option against `--version-sort` like this:

```log
% sort --version
sort (GNU coreutils) 5.93
Copyright (C) 2005 Free Software Foundation, Inc.
This is free software.  You may redistribute copies of it under the terms of
the GNU General Public License <http://www.gnu.org/licenses/gpl.html>.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
% sort --version-sort
sort: unrecognized option `--version-sort'
Try `sort --help' for more information.
```

PR #425, opened 2015-11-02
travis: use latest pkg-config on Travis CI's OS X instances

cosmo0920:
pkg-config 0.29 was merged in Homebrew 4 days ago.
This causes test failure on Travis CI's OSX instances.

When pkg-config 0.28 is installed on OS X with Homebrew
and then installing pkg-config 0.29:

NG:

``` bash
$ brew install pkg-config
```

OK:

```bash
$ brew unlink pkg-config
$ brew install pkg-config
```

or users can install pkg-config with `brew upgrade pkg-config`.

Above commands are tested on my MacBook '12 El Capitan.

### DIFF
--- a/tools/travis-install.sh
+++ b/tools/travis-install.sh
@@ -16,13 +16,12 @@ case "${TRAVIS_OS_NAME}" in
     ;;
   osx)
     brew update > /dev/null
-    brew unlink pkg-config
+    brew outdated pkg-config || brew upgrade pkg-config
     brew install \
          msgpack \
          libevent \
          mecab \
          mecab-ipadic \
-         pkg-config \
          pcre \
          cutter
     ;;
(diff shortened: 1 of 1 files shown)


kou:
I like `brew upgrade` style.
Could you use `upgrade`?

cosmo0920:
I like `brew upgrade` style too.

But above `brew unlink pkg-config && brew install pkg-config` is a bit of safer for Travis.
Because `brew install` return 0 even if already installation target installed.
`brew upgrade` return 1 if already installation target installed.

How about following `brew upgrade` style command?
ref: http://docs.travis-ci.com/user/osx-ci-environment/#A-note-on-upgrading-packages

```bash
$ brew outdated pkg-config || brew upgrade pkg-config
```

These command make clear what to do.

kou:
OK. We use `outdated || upgrade` style.
Thanks.

cosmo0920:
Thank you for merging!

PR #438, opened 2015-11-29
Avoiding conditional directives that split up parts of statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.


s-yata:
Thank you!
It looks nice to me, but I think this patch should be first applied to the master of nginx ( https://github.com/nginx/nginx ).

kou:
We'll merge the change when nginx merges them.
So I close this issue.

FYI: nginx accepts patches on mailing list: http://mailman.nginx.org/mailman/listinfo/nginx-devel

PR #450, opened 2016-01-02
Output score as float when commend_version > 1

naoa:
I want to output score as float to calculate similarity score such as jaccard similarity and cosine distance.


kou:
Thanks!

naoa:
Thanks for merging!

Issue #453, opened 2016-01-15
doc: groonga create table for memcached automatically.


hiroyuki-sato:
[Reference](http://blog.shigepon.info/blog/2016/01/15/use-groonga-for-django-cache/) (Japanese)

http://groonga.org/ja/docs/server/memcached.html

It is better to write following description.

Example.

You don't need to create table.
When groonga receive request, it create table automatically.
The table name is ``Memcache``.

Could you tell me how to update English doucment?
Should I change xxx.po file directory?

Best regards.

kou:
> Could you tell me how to update English doucment?

@kenhys Could you describe about it?

kenhys:
> Could you tell me how to update English doucment?
> Should I change xxx.po file directory?

Here is the document how to contribute 
http://groonga.org/ja/docs/contribution/documentation/introduction.html

To update English document, modify existent doc/source/*.rst is good enough
because po files for English document is automatically re-generated. (no need to edit doc/locale/en/LC_MESSAGES/*.po)

If you want to do more (translate document), see next i18n document here:
http://groonga.org/ja/docs/contribution/documentation/i18n.html

I hope it may help.

kenhys:
Sorry, I've closed by mistake.

hiroyuki-sato:
Thanks!!

PR #454, opened 2016-01-18
doc: add description about memcached table

hiroyuki-sato:
Follow-up https://github.com/groonga/groonga/issues/453

Due to my machine disk size problem, I can't update *.po files on my machine.
I'll appreciate if you create *.po files.


hiroyuki-sato:
@kenhys @kou I made po file.
But I don't know how to write highlight text.

in rst file
```rst
`Memcache`
```

How to write po file?
```
\"memcache\"
```

Thanks!.

kenhys:
Basically, describe as same as msgid for highlight.

For example, describe like

```
テーブルの名前は `Memcache` です。
```
in msgstr. Note that space is needed for keyword -  `Memcache` in this case.

hiroyuki-sato:
@kenhys.

I fixed. Please check.

I confused how to highlight text.
Because this page has two highlight text.
http://groonga.org/ja/docs/server/memcached.html

One is 
```
`--protocol`
```

The other is 
```
"memcached"
```

hiroyuki-sato:
I branched old tree. so I re-create pr with latest tree.

kou:
@kenhys Could you check this?

@cosmo0920 Could you help @kenhys as an English professional?

doc/locale/ja/LC_MESSAGES/server.po
```diff
@@ -449,6 +449,13 @@ msgstr ""
 "`--protocol` オプションとその引数により、サーバのプロトコルを指定することがで"
 "きます。\"memcached\"はmemcachedバイナリプロトコルを示しています。"
 
+msgid ""
+"You don't need to create table.  When groonga receive request, it create "
```
cosmo0920: `table` is a countable noun. It need to add "a" here.
And `request` is also a countable. Ditto.

cosmo0920: `it` points to `groonga`, right?
Then, it should add "s" like: "it create**s**"

doc/locale/ja/LC_MESSAGES/server.po
```diff
@@ -449,6 +449,13 @@ msgstr ""
...
+"table automatically.  The table name is `Memcache` ."
```
cosmo0920: This table doesn't exist before accepting request?
If so, we should use "future tense" like this: This table will be `Memcache`.

doc/locale/ja/LC_MESSAGES/server.po
cosmo0920: Be careful this `when clause` does not use "future tense" together. XD

cosmo0920:
I've commented several nitpicks about tense, thanks.

kenhys:
I've missed updated comment. thanks @hiroyuki-sato, @cosmo0920.

kenhys:
> I confused how to highlight text.
> Because this page has two highlight text.
> http://groonga.org/ja/docs/server/memcached.html

Good catch, it may not be clear how to markup text for highlight. 
OT: I think that it is better to write a note for this topic in future. 
http://groonga.org/ja/docs/contribution/documentation.html

hiroyuki-sato:
@cosmo0920 
Thanks.

I modified description. Could you check this?

You don't need to create a table.  Groonga creates a table automatically. The table will be `Memcache` .

cosmo0920:
> You don't need to create a table.

:ok: 

> Groonga creates a table automatically.

"When groonga receives a request, it creates table automatically. " is more good sentence.

> The table will be Memcache .

"The table name will be `Memcache`."  is prefer. see: https://github.com/groonga/groonga/pull/454#discussion_r50351291

hiroyuki-sato:
@cosmo0920 Thanks! fixed.

> You don't need to create a table. When groonga receives a request, it creates a table automatically. The table name will be `Memcache` .

I'll update *.po file later.

kou:
(Please use "Groonga" instead of "groonga"...)

hiroyuki-sato:
@kou 
Thanks, Changed.

> You don't need to create a table. When Groonga receives a request, it creates a table automatically. The table name will be `Memcache` .

kenhys:
Merged. 
@hiroyuki-sato, thanks!

PR #458, opened 2016-01-21
doc ja: fix typo

tSU-RooT:
I found other typos.


kenhys:
Merged.
Good catch! @tSU-RooT

PR #459, opened 2016-01-30
plugin functions/char: add char_length()

naoa:
I want to filter with charctor length as following.

```
select Users --filter 'char_length(name) == 2' --output_columns 'name' --command_version 2
```

### DIFF
--- a/plugins/functions/CMakeLists.txt
+++ b/plugins/functions/CMakeLists.txt
@@ -38,3 +38,23 @@ else()
   install(TARGETS vector_functions DESTINATION "${GRN_FUNCTIONS_PLUGIN_DIR}")
 endif()
 target_link_libraries(vector_functions libgroonga)
+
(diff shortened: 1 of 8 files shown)


kou:
Thanks for your pull request.
How about `string_length()` instead of `char_length()`. Because `char_length()` should return length of ONE character.

naoa:
I agree with your thought.

I'll change to ``string~`` from ``char~`` later.

cosmo0920:
Nit: I think that if we use `string_length()` instead of `char_length()`, we should use `string~` name in file name and its test cases name.

naoa:
I've renamed.
Also I've fixed CMakeLists.txt
Then I've squashed these commits.

cosmo0920:
:+1:

kou:
Thanks!

PR #475, opened 2016-02-11
column_remove: show more detail in error message

myokoym:
I improved error messages in `column_remove` command. I referred to [`column_rename`](https://github.com/groonga/groonga/blob/master/lib/proc.c#L1480) command.

### DIFF
--- a/lib/proc.c
+++ b/lib/proc.c
@@ -1437,10 +1437,13 @@ proc_column_remove(grn_ctx *ctx, int nargs, grn_obj **args, grn_user_data *user_
     if (col) {
       grn_obj_remove(ctx, col);
     } else {
-      ERR(GRN_INVALID_ARGUMENT, "column not found.");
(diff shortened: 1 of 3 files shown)


lib/proc.c
```diff
@@ -1437,10 +1437,13 @@ proc_column_remove(grn_ctx *ctx, int nargs, grn_obj **args, grn_user_data *user_
     if (col) {
       grn_obj_remove(ctx, col);
     } else {
-      ERR(GRN_INVALID_ARGUMENT, "column not found.");
+      ERR(GRN_INVALID_ARGUMENT, "[column][remove] column isn't found: <%.*s.%.*s>",
+          (int)GRN_TEXT_LEN(VAR(0)), GRN_TEXT_VALUE(VAR(0)),
+          (int)GRN_TEXT_LEN(VAR(1)), GRN_TEXT_VALUE(VAR(1)));
```
kou: Use `colname` and `colname_len`.

myokoym: Thanks. I've fixed it.

kou:
Thanks.

PR #492, opened 2016-02-26
Fix needless -latomic check on x86 and x64 host

kenhys:


configure.ac
```diff
@@ -1243,8 +1243,15 @@ AM_CONDITIONAL([ENABLE_SUGGEST_LEARNER],
                      "$message_pack_available" = "yes"])
 
 # Check built-in atomic
-AC_CHECK_LIB(atomic, __atomic_store_8, [ATOMIC_LIBS="-latomic"])
-AC_SUBST(ATOMIC_LIBS)
+case "$host" in
```
kou: Use `$host_cpu` instead.

configure.ac
```diff
@@ -1243,8 +1243,15 @@ AM_CONDITIONAL([ENABLE_SUGGEST_LEARNER],
...
+  i*86*|x86_64*)
```
kou: It seems that we don't use atomic functions on PPC: https://github.com/groonga/groonga/blob/master/lib/grn.h#L506

configure.ac
```diff
@@ -1243,8 +1243,15 @@ AM_CONDITIONAL([ENABLE_SUGGEST_LEARNER],
...
+    ;;
+  *)
+    AC_MSG_CHECKING([for platform which requires libatomic])
+    AC_CHECK_LIB(atomic, __atomic_store_8, [ATOMIC_LIBS="-latomic"])
```
kou: Does `atomic_add_32_nv()` require `-latomic`?: https://github.com/groonga/groonga/blob/master/lib/grn.h#L515

configure.ac
kou: http://docs.oracle.com/cd/E19253-01/816-5180/6mbbf01oj/index.html doesn't say about `-latomic`.

kou:
I don't want to merge this pull request with this status. But I couldn't wait for response... Because I need to release a new version... :<

PR #554, opened 2016-06-15
column_remove: add "dependent" parameter

kou:
If "dependent=no" (not default), column_remove reports error for a
column that has one or more indexes.

This change also changes grn_obj_remove(ctx, column) behavior. It
reports error for a column that has one ore more indexes. It's backward
incompatible change.

We need to use grn_obj_remove_dependent(ctx, column) for the previous
behavior.

This is not acceptable. So we need to consider about it carefully.


kou:
Don't merge this pull request. I created it just for memo.

PR #559, opened 2016-06-19
doc: Fix a typo

cafedomancer:
Maybe "problem" is countable?


myokoym:
Good catch! I'll update http://groonga.org.

PR #612, opened 2016-12-09
Add missing NULL check about return value of grn_ctx_at

kenhys:
There is a case that grn_ctx_at returns NULL. In such a case,
missing NULL check causes crash unexpectedly.


kou:
Good catch!

PR #614, opened 2016-12-16
Return immediately if grn_ctx_at returns NULL

kenhys:
GitHub#613

This commit avoids crash bug.


src/groonga.c
```diff
@@ -1796,6 +1796,9 @@ memcached_init(grn_ctx *ctx)
     }
 
     cache_table = grn_ctx_at(ctx, cache_value->header.domain);
+    if (!cache_table) {
+      return GRN_FALSE;
```
kou: Put error details like other codes do.

kenhys: Sure, I'll fix it later.

PR #626, opened 2017-01-14
Change to install groonga-httpd-restart conditionally

obache:
It should be installed with groonga-httpd, and should not be in
distribution file in any case.


kou:
Thanks! It's reasonable.

PR #628, opened 2017-01-16
Allow to select Onigmo support

obache:
Introduce --with-onigmo GNU configure option.

 "bundled" (default):  use in-tree Onigmo
 "system": try to use system Onigomo or rise error
 "yes": try to use system Onigmo or fallback to in-tree one
 "no" (--without-onigmo): disalbe Onigmo support

Default "bundled" is the same behavior so far.


kou:
Thanks!

PR #641, opened 2017-03-14
grn_pat_scan: ignore blanks in target string.

kou:
With this change, highlight family functions support highlighting text
that has some blanks. See the test in this change for example.


kou:
#641 のリベンジ。

PR #651, opened 2017-04-14
test: add test case when match_columns parameter is changed

kenhys:
When the value of --match_columms parameter is changed, Groonga


kenhys:
I fixed test case to `logical_select`.

kou:
Cache tests for `select` are also useful. So it was better that we add both `select` version and `logical_select` version.

PR #652, opened 2017-04-14
test: add no hit dynamic columns with sort_keys without error

kenhys:


kou:
`test/command/suite/select/columns/stage/initial/no_hit/sort_keys.test` is better because this test focuses on `columns` not `sort_keys`.

kenhys:
I've fixed test case.

kou:
Merged!

PR #653, opened 2017-04-14
test: add query test case without cache

kenhys:
Confirm whether query parameter is changed, Groonga returns result
without using cache.


kenhys:
I've fixed test case to use logical_select

kenhys:
I've fixed test case to use logical_select.

PR #656, opened 2017-04-14
test: add columns[NAME].value test case without cache

kenhys:


kenhys:
I've fixed test case to use logical_select

kou:
We should put `.expected` file instead of `.actual` file.

kenhys:
Oops! 
I've put `.expected` file.

kou:
Merged!

PR #657, opened 2017-04-14
test: add columns[NAME].window.sort_keys test case without cache

kenhys:


kou:
`UInt32` type `comment` column is a bit strange... I thought that `comment` column is`ShortText` type but I was wrong.

kenhys:
I've fixed test case to use `logical_select`.

kenhys:
There is one concern about test result. It seems that the value of `window_sum(price)` is calculated each `Logs_XXX`.

kou:
Merged.

> There is one concern about test result. It seems that the value of `window_sum(price)` is calculated each `Logs_XXX`.

It should be documented as a limitation or implemented.

PR #658, opened 2017-04-14
test: add columns[NAME].window.group_keys test case without cache

kenhys:


kou:
All `item` and `user` values are same. They aren't good test cases.

kou:
It's better that `window/group_keys.test` instead of `window_group_keys.test`.

kenhys:
I've fixed test case.

kou:
Merged.

PR #659, opened 2017-04-14
test: add drilldowns[LABEL].filter test case without cache

kenhys:


kenhys:
I've fixed test case to use `logical_select`.

kou:
It's better that query has `--limit 0` option. Because the original records aren't important for drilldown tests.

PR #660, opened 2017-04-14
test: add drilldowns[LABEL].columns[NAME].type test case without cache

kenhys:


kenhys:
I've fixed the test case to use `logical_select`.

kou:
`output_columns _id` for reducing needless output is a good practice.

PR #661, opened 2017-04-14
test: add drilldowns[LABEL].columns[NAME].value test case without cache

kenhys:


kenhys:
I've fixed test case to use `logical_select`

kou:
Using both `--output_columns _id` and `--limit 0` options is a good practice to reduce needless output.

PR #662, opened 2017-04-14
test: add drilldowns[LABEL].columns[NAME].window.sort_keys without cache

kenhys:
WIP because of error. Wrong usage of window.sort_keys?


kou:
`'price * 1.08'` isn't window function call expression.

kenhys:
I've fixed test case to use window function call

kou:
Merged.
But `number` name is meaningless...

PR #664, opened 2017-04-19
test: add no hit dynamic columns with filter without error

kenhys:


kou:
It seems that this is meaningless no hit test because there are all records in "initial" stage. "initial" stage is executed before `filter`.

kenhys:
It means that filtering should be done by `--min` option, doesn't it?

kenhys:
I've fixed test case again.

kou:
Merged!

PR #665, opened 2017-04-19
test: add no hit dynamic columns with match_columns without error

kenhys:
This test case checks whether match_columns with empty records of
dynamic columns (price_with_tax) in initial stage doesn't cause an
error.


kou:
`match_columns` isn't used when `query` isn't specified.

kou:
Normally, `UInt32` value isn't used as `match_columns`. So test data isn't suitable.

kenhys:
I've fixed test case with feedback.

kou:
Merged!

PR #666, opened 2017-04-19
test: add no hit dynamic columns with output_columns without error

kenhys:
This test case checks whether output_columns with empty records of
dynamic columns (price_with_tax) in initial stage doesn't cause an
error.


kou:
Merged.
`filtered`stage case is also needed for `output_columns` case.

PR #667, opened 2017-04-19
test: add columns[NAME].flags test case without cache

kenhys:
When the value of --columns[NAME].flags parameter is changed, Groonga

### DIFF
--- a/test/command/suite/sharding/logical_select/cache/columns/flags.expected
+++ b/test/command/suite/sharding/logical_select/cache/columns/flags.expected
@@ -1,0 +1,35 @@
+plugin_register sharding
+[[0,0.0,0.0],true]
+plugin_register functions/vector
+[[0,0.0,0.0],true]
+table_create Logs_20170415 TABLE_NO_KEY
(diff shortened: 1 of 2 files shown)


kenhys:
WIP because there is no different between output of VECTOR_COLUMN and VECTOR_COLUMN|WITH_WEIGHT.

kou:
How about adding any elements?

kenhys:
It means that `.value 'vector_new(1, 2, 3)'` ? I've tried it but no difference.

kou:
Ah. You need to use reference type or `ShortText` type for outputting weight vector. Or you can implement weight vector output for `uvector` in `grn_output_uvector()` at `output.c.

kenhys:
I've updated the test case which uses `ShortText`.

kou:
Merged.
Supporting outputting non reference column weight vector should be issued.

kenhys:
I've created issue as #669

kou:
Good!

kenhys:
I've also created found issue as #670

PR #668, opened 2017-04-19
test: add no hit dynamic columns with output_columns without error

kenhys:
This test case checks whether output_columns with empty records of
dynamic columns (highlighted_content) in filtered stage doesn't cause
an error.


kou:
It's better that we remove all records by `query` not `min` because `filtered` stage is executed after `query`.

kenhys:
I think that  execution order is: min -> query -> filtered. so it is good enough by `min`. Is `query` better in this case because it reduces scope of processing pipeline?

kou:
No records by `min` and no records by `query` use the different execution pass in implementation.

kenhys:
I've fixed to use `--query 'content:@slow'` in this case.

kou:
Merged.

PR #673, opened 2017-04-20
test: add no hit dynamic columns with drilldowns[LABEL].filter without error

kenhys:
This test case checks whether drilldowns[LABEL].filter with empty
records of dynamic columns (price_with_tax) in initial stage doesn't
cause an error.


kou:
Ah, we should refer a column created by `drilldowns[LABEL].columns[NAME].stage initial` in `drilldowns[LABEL].filter`.

kenhys:
Thanks.
I'll fix it.

kenhys:
I've just fixed test case.

kou:
There is a needless commit: https://github.com/groonga/groonga/pull/673/commits/2ada77138fb9dfc731e4bb3a3165992c70021fa5

kenhys:
Just removed https://github.com/groonga/groonga/pull/673/commits/2ada77138fb9dfc731e4bb3a3165992c70021fa5 .

kou:
Merged.

PR #674, opened 2017-04-20
test: add no hit dynamic columns with drilldown_filter without error

kenhys:
This test case checks whether drilldown_filter with empty drilldown
records (price_with_tax) in initial stage doesn't cause an error.


kou:
Ah, this pattern is meaningless. It's not needed. Sorry.

PR #675, opened 2017-04-20
test: add drilldowns[LABEL].columns[NAME].window.group_keys test case without cache

kenhys:
When the value of --drilldowns[LABEL].columns[NAME].window.group_keys
parameter is changed, Groonga returns result without using cache.

### DIFF
--- a/test/command/suite/sharding/logical_select/cache/drilldowns/columns/window/group_keys.test
+++ b/test/command/suite/sharding/logical_select/cache/drilldowns/columns/window/group_keys.test
@@ -1,0 +1,61 @@
+#@on-error omit
+plugin_register sharding
+#@on-error default
+
+table_create Items TABLE_HASH_KEY ShortText
+column_create Items price COLUMN_SCALAR UInt32
+
+table_create Logs_20170415 TABLE_NO_KEY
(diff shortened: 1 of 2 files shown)


test/command/suite/sharding/logical_select/cache/drilldowns/columns/window/group_keys.test
```diff
@@ -0,0 +1,59 @@
...
+logical_select Logs \
+   --shard_key timestamp \
+   --limit 0 \
+   --output_columns _id \
+   --drilldowns[group].keys item \
+   --drilldowns[group].columns[number].stage initial \
+   --drilldowns[group].columns[number].type UInt32 \
+   --drilldowns[group].columns[number].flags COLUMN_SCALAR \
+   --drilldowns[group].columns[number].value 'window_record_number()' \
+   --drilldowns[group].columns[number].window.group_keys number \
```
kou: Does this refer the self column? The behavior is undefined...

kenhys:
I've changed to use multiple keys in this case.
In rewriting test case process, I've found crash bug so I filed it as a issue https://github.com/groonga/groonga/issues/681

kou:
Merged.
If `output_columns` includes `_value.category`, I could confirm more easily.

PR #676, opened 2017-04-20
test: add drilldown_filter test case without cache

kenhys:
When the value of --drilldown_filter parameter is changed, Groonga


kou:
Merged.

PR #677, opened 2017-04-20
test: add drilldowns[LABEL].columns[NAME].flags test case without cache

kenhys:
When the value of --drilldowns[LABEL].columns[NAME].flags parameter is
changed, Groonga returns result without using cache.


kenhys:
I've expected that the result of `price_with_tax` is different, but not.

kenhys:
Now, I've changed to use `ShortText`. It now works as expected.

kou:
Merged.

PR #678, opened 2017-04-21
WIP test: add columns[NAME].stage test case without cache

kenhys:
When the value of --columns[NAME].stage parameter is changed, Groonga


kenhys:
I used `snippet_html()` but it causes an `constant is null` error.

kou:
`match_columns` is missing.

kenhys:
Thanks.
I've fixed test case. @kou

kenhys:
I've fixed test case to use `highlight_html()`. Now this test case is executed as expected.

kou:
It's strange that `initial` stage case doesn't output any records.

kenhys:
`initial` stage returns non empty results because it should be matched.

PR #679, opened 2017-04-24
test: add drilldowns[LABEL].table test case without cache

kenhys:
When the value of --drilldowns[LABEL].table parameter is changed,
Groonga returns result without using cache.


kou:
Merged.

PR #680, opened 2017-04-24
WIP test: add drilldowns[LABEL].columns[NAME].stage test case without cache

kenhys:
When the value of --drilldowns[LABEL].columns[NAME].stage parameter is
changed, Groonga returns result without using cache.

### DIFF
--- a/test/command/suite/sharding/logical_select/cache/drilldowns/columns/stage.expected
+++ b/test/command/suite/sharding/logical_select/cache/drilldowns/columns/stage.expected
@@ -1,0 +1,162 @@
+plugin_register sharding
+[[0,0.0,0.0],true]
+table_create Items TABLE_HASH_KEY ShortText
+[[0,0.0,0.0],true]
+column_create Items price COLUMN_SCALAR UInt32
+[[0,0.0,0.0],true]
+table_create Logs_20170315 TABLE_NO_KEY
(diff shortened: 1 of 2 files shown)


kenhys:
I've expected that output of this case is different, but not. any idea?

kou:
Could you try `--drilldowns[LABEL].filter` instead of `--query`?

kenhys:
> Could you try --drilldowns[LABEL].filter instead of --query?

Could you explain what you mean precisely?
In other test case, we discussed that `--filter` is not worked expectedly for `stage`, so switched to use `--query`. 
In this case, I think that the situation is same as before. But is `--drilldowns[LABEL].filter` effective?

kou:
Issue number please.

kenhys:
> In other test case

I means #678 as other test case.

kou:
It's a `--columns[NAME]` test. This is a `--drilldowns[LABEL].columns[NAME]` test.

kenhys:
Note that `filtered` stage is not supported yet. so this test always fails.
And the following part is meaningless because query is already processed before drilldown.

```
--match_columns items \
+  --output_columns content \
+  --query "Book"
```

See http://groonga.org/ja/docs/reference/commands/select.html#select-columns-name-stage about processing order of dynamic columns and drilldowns.

Issue #694, opened 2017-05-21
Can't sequetial match by specifying index of vector.


naoa:
``GRN_UVECTOR`` may not be 4 bytes fixed length. For example, ``Int8``, ``Int64``, ``Float`` are also ``GRN_UVECTOR``.
Sequential search can't correctly matched by specifying index of non 4bytes fixed length vector.

```json
table_create numbers TABLE_NO_KEY
[[0,0.0,0.0],true]
column_create numbers int8 COLUMN_VECTOR Int8
[[0,0.0,0.0],true]
column_create numbers int32 COLUMN_VECTOR Int32
[[0,0.0,0.0],true]
column_create numbers int64 COLUMN_VECTOR Int64
[[0,0.0,0.0],true]
column_create numbers float64 COLUMN_VECTOR Float
[[0,0.0,0.0],true]
load --table numbers
[
{"int8": [1, 2], "int32": [11, 12], "int64": [101, 102], "float64": [0.1, 0.2]},
{"int8": [3, 4], "int32": [13, 14], "int64": [103, 104], "float64": [0.3, 0.4]}
]
[[0,0.0,0.0],2]
select numbers   --filter 'int8[1] == 4'   --output_columns '_id, int8'
[[0,0.0,0.0],[[[0],[["_id","UInt32"],["int8","Int8"]]]]]
select numbers   --filter 'int32[1] == 14'   --output_columns '_id, int32'
[[0,0.0,0.0],[[[1],[["_id","UInt32"],["int32","Int32"]],[2,[13,14]]]]]
select numbers   --filter 'int64[1] == 104'   --output_columns '_id, int64'
[[0,0.0,0.0],[[[0],[["_id","UInt32"],["int64","Int64"]]]]]
select numbers   --filter 'float64[1] == 0.4'   --output_columns '_id, float64'
[[0,0.0,0.0],[[[0],[["_id","UInt32"],["float64","Float"]]]]]
```

``vector_slice`` also can't get correctly slice of non 4 bytes fixed vector column.

```json
plugin_register functions/vector
select numbers   --filter true   --output_columns '_id, vector_slice(int32, 1, 1)'
[
  [
    0,
    0.0,
    0.0
  ],
  [
    [
      [
        2
      ],
      [
        [
          "_id",
          "UInt32"
        ],
        [
          "vector_slice",
          null
        ]
      ],
      [
        1,
        [
          12
        ]
      ],
      [
        2,
        [
          14
        ]
      ]
    ]
  ]
[...]

PR #696, opened 2017-05-21
expr: fix a bug that non 4 bytes fixed vector column can't sequential…

naoa:
… match by specifying index of vector

GitHub: #694


kou:
Thanks!!

PR #699, opened 2017-06-02
db: add missing null check for grn_table_setoperation()

kenhys:
This commit fixes a crash bug when grn_table_setoperation() is called
with unexpected NULL arguments.
Such a situation may occur when indexes are broken.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -4464,6 +4464,10 @@ grn_table_setoperation(grn_ctx *ctx, grn_obj *table1, grn_obj *table2, grn_obj *
   uint32_t value_size = 0;
   uint32_t key_size = 0;
   grn_bool have_subrec;
+  if (!table1 || !table2) {
+    ERR(GRN_INVALID_ARGUMENT, "[db] can't set operation against null table");
+    return ctx->rc;
+  }
   if (table1 != res) {
     if (table2 == res) {
       grn_obj *t = table1;


lib/db.c
```diff
@@ -4464,6 +4464,10 @@ grn_table_setoperation(grn_ctx *ctx, grn_obj *table1, grn_obj *table2, grn_obj *
   uint32_t value_size = 0;
   uint32_t key_size = 0;
   grn_bool have_subrec;
+  if (!table1 || !table2) {
+    WARN(GRN_INVALID_ARGUMENT, "table is null");
```
kou: We should provide more meaningful error message like: https://github.com/groonga/groonga/blob/master/lib/db.c#L313

kou: It should be error not warning.

lib/db.c
```diff
@@ -4464,6 +4464,10 @@ grn_table_setoperation(grn_ctx *ctx, grn_obj *table1, grn_obj *table2, grn_obj *
...
+    return GRN_INVALID_ARGUMENT;
```
kou: We can use `ctx->rc`.

kenhys:
I've fixed

* use ERR
* return ctx->rc
* more meaningful error message

lib/db.c
```diff
@@ -4464,6 +4464,10 @@ grn_table_setoperation(grn_ctx *ctx, grn_obj *table1, grn_obj *table2, grn_obj *
   uint32_t value_size = 0;
   uint32_t key_size = 0;
   grn_bool have_subrec;
+  if (!table1 || !table2) {
```
kou: It's better that we can identify which `table1` or `table2` is `NULL` on error.

lib/db.c
```diff
@@ -4464,6 +4464,10 @@ grn_table_setoperation(grn_ctx *ctx, grn_obj *table1, grn_obj *table2, grn_obj *
...
+    ERR(GRN_INVALID_ARGUMENT, "[db] can't set operation against null table");
```
kou: `[db]` tag isn't suitable in `grn_table_XXX()` function. We should use `[table]` tag.

kenhys:
Thanks! I've checked https://github.com/groonga/groonga/commit/f662925c6a0ac3a56efa5facfda51ad8165255e3

PR #700, opened 2017-06-05
Add more delete operation logs

kenhys:
It is needed to identify problem about vanished database files.


kou:
Merged with some modifications.

kenhys:
Thanks! I've checked it.

PR #701, opened 2017-06-05
Add more create operation logs

kenhys:
It may be useful to identify lifetime of created files.


kou:
Merged with some modifications

PR #702, opened 2017-06-05
WIP grndb: support --log-path option

kenhys:
It is useful feature to verify grndb log when something weird happens

### DIFF
--- a/lib/mrb/mrb_logger.c
+++ b/lib/mrb/mrb_logger.c
@@ -58,6 +58,17 @@ logger_log(mrb_state *mrb, mrb_value self)
   return self;
 }
 
+static mrb_value
+logger_path(mrb_state *mrb, mrb_value self)
+{
+  char *path;
+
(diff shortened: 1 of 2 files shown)


lib/mrb/mrb_logger.c
```diff
@@ -70,6 +81,7 @@ grn_mrb_logger_init(grn_ctx *ctx)
 
   mrb_define_method(mrb, klass, "need_log?", logger_need_log_p, MRB_ARGS_REQ(1));
   mrb_define_method(mrb, klass, "log", logger_log, MRB_ARGS_REQ(5));
+  mrb_define_method(mrb, klass, "path=", logger_path, MRB_ARGS_REQ(1));
```
kou: It should be `Groonga::Logger.default_path=` instead of `Groonga::Logger#path=`.

kou:
`--log-path` should be processed before `grn_init()` in `src/grndb.c`.

kou:
I've implemented.

PR #705, opened 2017-06-07
grndb: change log level from notice to info

kenhys:
When physical file operation (add/remove) is now logged in info level,
so log level for grndb should be GRN_LOG_INFO because above logs are
not logged in notice level.

### DIFF
--- a/src/grndb.c
+++ b/src/grndb.c
@@ -152,6 +152,7 @@ main(int argc, char **argv)
   }
 
   grn_default_logger_set_path(log_path);
+  grn_default_logger_set_max_level(GRN_LOG_INFO);
 
   if (grn_init() != GRN_SUCCESS) {
     return EXIT_FAILURE;


kou:
We should provide `--log-level` option.

kenhys:
OK, but default log level should be info for grndb, isn't it ?

* grndb log level to info (This PR) because grndb is used for troubleshooting
* add log-level option for grndb (not yet, I'll support it anther PR)

kou:
> OK, but default log level should be info for grndb, isn't it ?

I don't think so.



> grndb log level to info (This PR) because grndb is used for troubleshooting

With this logic, we should use more verbose log level such as debug log level.

kenhys:
>> grndb log level to info (This PR) because grndb is used for troubleshooting

>With this logic, we should use more verbose log level such as debug log level.

If default log level for grndb is changed to debug, it it still acceptable idea or not?

kou:
Not acceptable. Production tool shouldn't use debug log level by default.

kenhys:
> Production tool shouldn't use debug log level by default.

There is no reason to insist 'changing default log level to debug' anymore, so I close this PR.

PR #706, opened 2017-06-07
grndb: support --log-level option

kenhys:


kou:
Merged with some modifications.

PR #709, opened 2017-06-08
Support persistent expression

naoa:
I would like to use index for computed value of some expression to optimize filter.
(For example, min(A, B) > C)

So, I've implemented persistent expression and ``GRN_OP_EXPR_EXEC`` at first.

Is this implementation acceptable?


naoa:
Oh sorry, I noticed that the expression should not be independent object but meta data of index_column object.

I will PR again when I change.

PR #714, opened 2017-06-20
Add --force-truncate option for grndb

kenhys:
This option enable to recover corrupted object in database.

### DIFF
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -402,6 +402,14 @@ module Groonga
           failed(message)
         end
       end
+
+      class Recoverer
+        attr_writer :database
+
+        def recover
+          @database.recover
+        end
+      end
     end
   end
 end
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -92,8 +92,10 @@ module Groonga
       end
 
       def recover(database, options, arguments)
+        recoverer = Recoverer.new
+        recoverer.database = database
         begin
-          database.recover
+          recoverer.recover
         rescue Error => error
           failed("Failed to recover database: <#{@database_path}>",
                  error.message)
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -45,6 +45,7 @@ module Groonga
 
           options = command.options
           options.banner += " DB_PATH"
+          options.boolean("--force-truncate", "Force to truncate corrupted objects.")
 
           command.add_action do |options|
             open_database(command, options) do |database, rest_arguments|
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -95,6 +95,7 @@ module Groonga
       def recover(database, options, arguments)
         recoverer = Recoverer.new
         recoverer.database = database
+        recoverer.force_truncate = options[:force_truncate]
         begin
           recoverer.recover
         rescue Error => error
@@ -408,9 +409,42 @@ module Groonga
 
       class Recoverer
         attr_writer :database
+        attr_writer :force_truncate
+
+        def initialize
+          @context = Context.instance
+        end
 
         def recover
           @database.recover
+          if @force_truncate
+            truncate_corrupt_objects
+          end
+        end
+
+        def truncate_corrupt_objects
+          @database.each do |object|
+            next unless object.corrupt?
+            logger = @context.logger
+            object_path = object.path
+            object_dirname = File.dirname(object_path)
+            object_basename = File.basename(object_path)
+            object.truncate
+            Dir.foreach(object_dirname) do |path|
+              if path.start_with?("#{object_basename}.")
+                begin
+                  File.unlink("#{object_dirname}/#{path}")
+                  message = "Corrupted <#{object_path}> related file is removed: <#{path}>"
+                  $stdout.puts(message)
+                  logger.log(Logger::Level::INFO.to_i, __FILE__, __LINE__, "truncate_corrupt_objects", message)
+                rescue Error => error
+                  message = "Failed to remove file which is related to corrupted <#{object_path}>: <#{path}>"
+                  $stderr.puts(message)
+                  logger.log_error(message)
+                end
+              end
+            end
+          end
         end
       end
     end


lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +405,37 @@ def failed_to_open(name)
...
+        attr_writer :database
+
+        def initialize
+          @context = Context.instance
+        end
+
+        def recover(force_truncate)
+          @database.recover
+          if force_truncate
+            @database.each do |object|
```
kou: This block should be extracted as a method.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +405,37 @@ def failed_to_open(name)
...
+              if object.corrupt?
```
kou: We should use 'next unless object.corrupt?` instead of deep nested code.
See also: http://www.clear-code.com/blog/2012/3/28.html (in Japanese)

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +405,37 @@ def failed_to_open(name)
...
+                logger = @context.logger
+                Dir.foreach(File.dirname(object.path)) do |path|
```
kou: We should avoid many `object.path` call.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +405,37 @@ def failed_to_open(name)
...
+                  if path =~ /^#{File.basename(object.path)}.+/
```
kou: We should use `Regexp#=~` than `String#=~` because regular expression match should be done by regular expression object.

kou: We should use path.start_with?("#{File.basename(object.path)}.")`.
See also: http://www.clear-code.com/blog/2012/11/1.html (in Japanese)

lib/mrb/scripts/command_line/grndb.rb
kou: We should run `object.truncate` before this block.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +405,37 @@ def failed_to_open(name)
...
+                    begin
+                      File.unlink("#{File.dirname(object.path)}/#{path}")
+                      message = "Removed file which is related to corrupted #{object.path}: #{path}"
+                      logger.log(Logger::Level::INFO.to_i, __FILE__, __LINE__, "recover", message)
```
kou: We should use standard output instead of logger like other codes in this file do.

lib/mrb/scripts/command_line/grndb.rb
kenhys: I tried object.truncate but it's bad for VariableSizeColumn. How should i truncate targeted object in this context?

lib/mrb/scripts/command_line/grndb.rb
kenhys: I think that logging actually removed file to log file is important in this context because standard output may be lost by user without redirecting, is it good enough to use standard output?

kenhys:
I've fixed with above feedback except the following two feedback
* logger or standard output
* object.truncate

lib/mrb/scripts/command_line/grndb.rb
kou: Why is it bad idea?

lib/mrb/scripts/command_line/grndb.rb
kou: Users will not notice the information with this code. Because the default `grndb` log path is invalid for normal users and the `info` level log isn't logged by default.

So I don't agree with

> standard output may be lost by user without redirecting

The code does nothing on many cases. I think that it's danger than standard output without redirect.

If you want to log, you should do both printing to standard output and logging.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +406,42 @@ def failed_to_open(name)
...
+            truncate_corrupt_objects
+          end
+        end
+
+        def truncate_corrupt_objects
+          @database.each do |object|
+            next unless object.corrupt?
+            logger = @context.logger
+            object_path = object.path
+            Dir.foreach(File.dirname(object_path)) do |path|
```
kou: `dirname` is also defined only once.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +406,42 @@ def failed_to_open(name)
...
+              if path.start_with?("#{File.basename(object_path)}.")
+                begin
+                  File.unlink("#{File.dirname(object_path)}/#{path}")
+                  message = "Removed file which is related to corrupted #{object_path}: #{path}"
+                  logger.log(Logger::Level::INFO.to_i, __FILE__, __LINE__, "recover", message)
```
kou: Method name isn't `"recover"`.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -402,6 +406,42 @@ def failed_to_open(name)
...
+        def truncate_corrupt_objects
+          @database.each do |object|
+            next unless object.corrupt?
+            logger = @context.logger
+            object_path = object.path
+            Dir.foreach(File.dirname(object_path)) do |path|
+              if path.start_with?("#{File.basename(object_path)}.")
+                begin
+                  File.unlink("#{File.dirname(object_path)}/#{path}")
+                  message = "Removed file which is related to corrupted #{object_path}: #{path}"
```
kou: Variable in message should be surrounded with `<...>`. You can find useful information such as empty value and garbage spaces at the last by this technique.

lib/mrb/scripts/command_line/grndb.rb
kenhys: Oh, I've forgot the case (the default grndb log path).
I agree with the idea not to omit implementation about standard output.

lib/mrb/scripts/command_line/grndb.rb
kenhys: In this context, "bad" means that "It is not enough".
object.truncate is not applicable against `VariableSizeColumn` (NoMethodError)
So, we need to implement grn_table_trucate, grn_column_truncate, grn_ii_truncate in mrb, does it correct?

lib/mrb/scripts/command_line/grndb.rb
kenhys: Ah, I'll fix it.

lib/mrb/scripts/command_line/grndb.rb
kenhys: Oh, sure. I'll fix it.

lib/mrb/scripts/command_line/grndb.rb
kou: Yes. I've implemented.

kenhys:
Here is the corrupted database.

```
PS C:\cygwin64\home\Administrator\kenhys> dir .\testdb.corrupted


    ディレクトリ: C:\cygwin64\home\Administrator\kenhys\testdb.corrupted


Mode                LastWriteTime     Length Name
----                -------------     ------ ----
-a---        2017/06/23     11:47      65536 db
-a---        2017/06/23     11:47   12910592 db.0000000
-a---        2017/06/23     11:47    8454144 db.0000100
-a---        2017/06/23     11:47 1069875200 db.0000101
-a---        2017/06/23     11:48 1073741824 db.0000101.001
-a---        2017/06/23     11:49  562036736 db.0000101.003
-a---        2017/06/23     11:47    1048576 db.001
-a---        2017/06/23     11:47   12648448 db.conf
```

Then execute grndb check:

```
PS C:\cygwin64\home\Administrator\kenhys> .\groonga-7.0.4-x64\bin\grndb.exe check --log-path grndb.check.win.log --log-level info .\testdb.corrupted\db
[x.a] Data column is corrupt. (1) Truncate the column (truncate x.a or 'C:\cygwin64\home\Administrator\kenhys\groonga-7.
0.4-x64\bin\grndb.exe recover --force-truncate .\testdb.corrupted\db') and (2) load data again.
```

After `grndb check`, execute grndb recover with --force-truncate option.
 
```
PS C:\cygwin64\home\Administrator\kenhys> .\groonga-7.0.4-x64\bin\grndb.exe recover --force-truncate --log-path grndb.re
conver.force.win.log --log-level info .\testdb.corrupted\db
Corrupted <.\testdb.corrupted\db.0000101> related file is removed: <db.0000101.003>
```

.003 is also removed.

```

PS C:\cygwin64\home\Administrator\kenhys> dir .\testdb.corrupted


    ディレクトリ: C:\cygwin64\home\Administrator\kenhys\testdb.corrupted


Mode                LastWriteTime     Length Name
----                -------------     ------ ----
-a---        2017/06/23     18:31      65536 db
-a---        2017/06/23     11:47   12910592 db.0000000
-a---        2017/06/23     11:47    8454144 db.0000100
-a---        2017/06/23     18:31     327680 db.0000101
-a---        2017/06/23     11:47    1048576 db.001
-a---        2017/06/23     11:47   12648448 db.conf
```

PR #716, opened 2017-06-21
Add null check of grn_ctx_at()'s return value in geo.c

komainu8:
GitHub #613
This commit avoids crash bug.


kou:
I've merged with some modifications.

PR #719, opened 2017-06-26
grndb: recover locked objects in --force-truncate mode

kenhys:
In the previous versions, grndb recover --force-truncate doesn't work if there are locked objects.
This PR supports it.

### DIFF
--- a/lib/mrb/mrb_column.c
+++ b/lib/mrb/mrb_column.c
@@ -111,6 +111,17 @@ mrb_grn_column_is_locked(mrb_state *mrb, mrb_value self)
 }
 
 static mrb_value
+mrb_grn_column_clear_lock(mrb_state *mrb, mrb_value self)
+{
+  grn_ctx *ctx = (grn_ctx *)mrb->ud;
+
+  grn_obj_clear_lock(ctx, DATA_PTR(self));
+  grn_mrb_ctx_check(mrb);
+
+  return mrb_nil_value();
+}
+
+static mrb_value
 mrb_grn_column_get_table(mrb_state *mrb, mrb_value self)
 {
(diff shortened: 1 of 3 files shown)


lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -423,12 +427,20 @@ def recover
...
+              when Table, Column
+                object.clear_lock if object.locked?
+              end
+              next
+            end
             logger = @context.logger
             object_path = object.path
             object_dirname = File.dirname(object_path)
             object_basename = File.basename(object_path)
+            object.clear_lock if object.locked?
```
kou: Does we need `clear_lock` for truncated objects?

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -99,8 +99,12 @@ def recover(database, options, arguments)
         begin
           recoverer.recover
         rescue Error => error
-          failed("Failed to recover database: <#{@database_path}>",
-                 error.message)
+          if options[:force_truncate]
```
kou: It should be done `recoverer.recover`.

kenhys: Database#recover raise an error so Recoverer#truncate_corrupt_objects will be skipped.
But as you pointed out, it seems that it is better to do in `recoverer.recover`.

lib/mrb/scripts/command_line/grndb.rb
kenhys: At first, I've misunderstood behavior of truncated object. No need to clear_lock.

kenhys:
This PR is not suitable to solve issue. I'll close it.

PR #720, opened 2017-06-26
grndb: truncate object in advance for force truncate mode

kenhys:
Database#recover fails when corrupted object is locked.
so truncate such a object in advance.

### DIFF
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -416,10 +416,10 @@ module Groonga
         end
 
         def recover
-          @database.recover
           if @force_truncate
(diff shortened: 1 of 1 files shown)


lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -416,9 +416,13 @@ def initialize
         end
 
         def recover
-          @database.recover
-          if @force_truncate
-            truncate_corrupt_objects
+          begin
```
kou: This `begin` is meaningless.

kenhys: Oh, thanks feedback. I've fixed.

PR #726, opened 2017-07-01
Add limit and ofset for window function

naoa:
I would like to define limit and offset for window function for huge size dataset.
(e.g. If a resultset has 1M records, it needs to set 1M ids)

### DIFF
--- a/include/groonga/window_function.h
+++ b/include/groonga/window_function.h
@@ -46,6 +46,8 @@ typedef struct _grn_window_definition {
   size_t n_sort_keys;
   grn_table_sort_key *group_keys;
   size_t n_group_keys;
+  int offset;
+  int limit;
 } grn_window_definition;
 
 typedef grn_rc grn_window_function_func(grn_ctx *ctx,
--- a/lib/mrb/mrb_window_definition.c
+++ b/lib/mrb/mrb_window_definition.c
@@ -138,6 +138,32 @@ mrb_grn_window_definition_set_group_keys(mrb_state *mrb, mrb_value self)
   return mrb_nil_value();
 }
 
+static mrb_value
+mrb_grn_window_definition_set_offset(mrb_state *mrb, mrb_value self)
+{
+  grn_window_definition *definition;
+  mrb_int offset;
+
+  mrb_get_args(mrb, "i", &offset);
+  definition = DATA_PTR(self);
+  definition->offset = offset;
+
+  return mrb_nil_value();
+}
+
+static mrb_value
+mrb_grn_window_definition_set_limit(mrb_state *mrb, mrb_value self)
+{
+  grn_window_definition *definition;
+  mrb_int limit;
+
+  mrb_get_args(mrb, "i", &limit);
+  definition = DATA_PTR(self);
+  definition->limit = limit;
+
+  return mrb_nil_value();
+}
+
 void
 grn_mrb_window_definition_init(grn_ctx *ctx)
 {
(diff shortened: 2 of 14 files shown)


kou:
Umm... Could you show use case?

naoa:
At first, I was thinking to get some slices have different # of records with one query as following.
For example, # of output records is 10-100 including some detail info to show search results, # of output records on slice is 1000-100000 including just short info to visualize document featured vector.

```
select Documents --filter 'index1 @ "データ"' ¥
  --columns[map].stage filtered ¥
  --columns[map].type Float ¥
  --columns[map].flags COLUMN_VECTOR ¥
  --columns[map].value 'window_pca(featured_vector,{"n_clusters": 10})' ¥
  --columns[map].window.sort_keys '-_score' ¥
  --columns[map].window.limit 5 ¥
  --slices[featured_map].filter 'vector_size(map) > 0' ¥
  --slices[featured_map].limit -1 ¥
  --slices[featured_map].output_columns _id,_score,app_id,title,abstract,map,map_cluster ¥
  --output_columns _id,app_id¥
  --sort_keys -_score --limit 1
```
(``window_pca`` function processes Principal component analysis(PCA). PCA is reducing dimension of matrix to make easy to visualize.)

But, After I started to implement WebApp, I am thinking now it is okay with 2 queries using just ``output`` stage. 

```
select Documents --filter 'index1 @ "データ"' ¥
--columns[map].stage output ¥
--columns[map].type Float ¥
--columns[map].flags COLUMN_VECTOR ¥
--columns[map].value 'window_pca(featured_vector,{"n_clusters": 10})' ¥
--columns[map].window.sort_keys '-_score' ¥
--output_columns _id,_score,app_id,title,map,map_cluster ¥
--sort_keys -_score --limit 5
```
So, If you feel it is non natural way, I am okay that this PR will be closed.

kou:
Thanks.

I think that we need dynamic columns support for slices instead of `window.offset` and `window.limit`:

```text
select Documents \
  --filter 'index1 @ "データ"' \
  --slices[featured_map1].columns[map].stage output \
  --slices[featured_map1].columns[map].type Float \
  --slices[featured_map1].columns[map].flags COLUMN_VECTOR \
  --slices[featured_map1].columns[map].value 'window_pca(featured_vector1,{"n_clusters": 10})' \
  --slices[featured_map1].columns[map].window.sort_keys '-_score' \
  --slices[featured_map1].output_columns id,_score,app_id,title,map,map_cluster \
  --slices[featured_map1].sort_keys -_score \
  --slices[featured_map1].limit 5 \
  --slices[featured_map1].columns[map].window.sort_keys '-_score' \
  --slices[featured_map2].columns[map].stage output \
  --slices[featured_map2].columns[map].type Float \
  --slices[featured_map2].columns[map].flags COLUMN_VECTOR \
  --slices[featured_map2].columns[map].value 'window_pca(featured_vector2,{"n_clusters": 10})' \
  --slices[featured_map2].columns[map].window.sort_keys '-_score' \
  --slices[featured_map2].sort_keys -_score \
  --slices[featured_map2].limit 5 \
  --output_columns _id \
  --limit 0
```

naoa:
Thanks.
I think that it is more natural way too.
But, currently ``slices`` require ``filter`` as mandatory.

```
  --slices[featured_map1].filter 'all_records() \
```
So I feel it is a little waste to copy all ids of resultset.
(The resultset may have around 1M records.)

``slices`` allow to pass a case having ``sort_keys`` and  no ``filter`` in special.
The case uses just the original resultset for sorting.

What do you think it? Is this way acceptable?

kou:
> What do you think it? Is this way acceptable?

Ah, no filtering isn't OLAP slicing...
Umm... OK. We accept the no filter case for `slices` as a special case.

naoa:
Thanks for discussion.

I'll implement dynamic column creation on ``slices`` and no filter case for ``slices`` later.
(It may be a while later, but it's useful for me, so I will implement it.)

This PR is closed.

kou:
Thanks!

PR #733, opened 2017-07-11
functions/time: avoid divided by zero

naoa:
It is caused crash process.

### DIFF
--- a/plugins/functions/time.c
+++ b/plugins/functions/time.c
@@ -127,6 +127,22 @@ func_time_classify_raw(grn_ctx *ctx,
     grn_obj_cast(ctx, interval, &casted_interval, GRN_FALSE);
     interval_raw = GRN_UINT32_VALUE(&casted_interval);
     GRN_OBJ_FIN(ctx, &casted_interval);
+
+    if (interval_raw == 0) {
+      grn_obj inspected;
+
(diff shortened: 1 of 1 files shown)


plugins/functions/time.c
```diff
@@ -127,6 +127,16 @@ func_time_classify_raw(grn_ctx *ctx,
...
     interval_raw = GRN_UINT32_VALUE(&casted_interval);
     GRN_OBJ_FIN(ctx, &casted_interval);
+
+    if (interval_raw == 0) {
+      GRN_PLUGIN_ERROR(ctx, GRN_INVALID_ARGUMENT,
+                       "%s(): "
+                       "the second argument must not be zero: "
+                       "<%u>",
+                       function_name,
+                       interval_raw);
```
kou: `"<%u>"` and `interval_raw` are meaningless because message says `must not be zero`.
Please remove them or show `interval`. `interval` will be helpful. Non-`0` `interval` value may be zero such as `-0`. (Maybe overflowed case too?)

naoa:
I've fixed to show the ``interval``.

overflow case sample:
```
"time_classify_month(): the second argument must not be zero: <4294967296>
```

kou:
Thanks!
Adding a test is better. :-)

PR #734, opened 2017-07-11
functions/time: add test for overflow interval case

naoa:
I added test case base on the comment of #733.
I thought it doesn't need because it is invalid case.


kou:
Thanks!
We need invalid cases to confirm Groonga doesn't crash for invalid cases.

naoa:
I see.
Thanks for merging.

PR #735, opened 2017-07-12
min,max: support number in vector

naoa:
I would like to retrieve number in`` grn_uvector`` with ``min`` and ``max``.


kou:
Thanks!!

PR #736, opened 2017-07-12
Add `--sort_hash_table` option in dump command

komainu8:


kou:
I've merged with some modifications.

PR #746, opened 2017-07-24
Remove needless `of` in file name.

komainu8:


kou:
Thanks!

kou:
`doc/files.am` should also be updated.

PR #752, opened 2017-08-16
Add selector to check lexicon record which desn't contain any posting list

kenhys:


kou:
It should be implemented as a C API that the target token ID doesn't match anything.

kou:
All or nothing match tests are insufficient. Some tokens are matched but others aren't matched case is needed.

kou:
Test for invalid index name argument case is missing.

plugins/functions/index_column.c
```diff
@@ -234,6 +234,63 @@ func_index_column_df_ratio(grn_ctx *ctx,
...
+    rc = ctx->rc;
+    goto exit;
+  }
+
+  column_name = GRN_TEXT_VALUE(args[1]);
+  
+  index_column = grn_obj_column(ctx, table, column_name, strlen(column_name));
+  ii = (grn_ii *)index_column;
+
+  source_table = grn_ctx_at(ctx, grn_obj_get_range(ctx, index_column));
```
kou: It's not used.

plugins/functions/index_column.c
```diff
@@ -234,6 +234,63 @@ func_index_column_df_ratio(grn_ctx *ctx,
...
+        pos.pos = 0;
+        pos.weight = 0;
+        grn_ii_posting_add(ctx, &pos, (grn_hash *)res, op);
+      }
+      grn_ii_cursor_close(ctx, ii_cursor);
+    }
+  } GRN_TABLE_EACH_END(ctx, cursor);
+
+exit :
+  grn_ii_resolve_sel_and(ctx, (grn_hash *)res, op);
```
kou: Should it be called on `exit` case?

plugins/functions/index_column.c
kenhys: Thanks, I've removed it.

kenhys:
I've `push -f` fixed version, 
* Changed to use experimental C API
* Added invalid index name testcase
* Added some tokens are matched testcase

include/groonga/ii.h
```diff
@@ -63,6 +63,7 @@ GRN_API grn_ii_cursor *grn_ii_cursor_open(grn_ctx *ctx, grn_ii *ii, grn_id tid,
 GRN_API grn_posting *grn_ii_cursor_next(grn_ctx *ctx, grn_ii_cursor *c);
 GRN_API grn_posting *grn_ii_cursor_next_pos(grn_ctx *ctx, grn_ii_cursor *c);
 GRN_API grn_rc grn_ii_cursor_close(grn_ctx *ctx, grn_ii_cursor *c);
+GRN_API grn_bool grn_ii_target_token_id_not_match(grn_ctx *ctx, grn_ii *ii, grn_id tid);
```
kou: How about `grn_ii_match_any()`?

lib/ii.c
```diff
@@ -5533,6 +5533,27 @@ grn_ii_cursor_close(grn_ctx *ctx, grn_ii_cursor *c)
...
 }
 
+grn_bool
+grn_ii_target_token_id_not_match(grn_ctx *ctx, grn_ii *ii, grn_id tid)
+{
+  grn_bool not_match = GRN_FALSE;
+
+  if (!ii) { return GRN_FALSE; }
+
+  uint32_t n_elements = grn_ii_get_n_elements(ctx, ii);
```
kou: Don't define variable after any expressions.

plugins/functions/index_column.c
```diff
@@ -234,6 +234,55 @@ func_index_column_df_ratio(grn_ctx *ctx,
   }
 }
 
+static grn_rc
+selector_index_column_not_match(grn_ctx *ctx,
```
kou: How about `selector_index_column_match_none()`?

plugins/functions/index_column.c
```diff
@@ -234,6 +234,55 @@ func_index_column_df_ratio(grn_ctx *ctx,
...
+  
+  index_column = grn_obj_column(ctx, table, column_name, strlen(column_name));
+  if (!index_column) {
+    GRN_PLUGIN_ERROR(ctx,
+                     GRN_INVALID_ARGUMENT,
+                     "index_column_not_match(): "
+                     "no such index column: <%s>", column_name);
+    return ctx->rc;
+  }
+
```
kou: Need `grn_obj_is_index_column()` check and test for it.

PR #758, opened 2017-08-27
Fix crash when ruby_eval returned syntax error

ryo-pinus:
Fixed crash when ruby_eval returned syntax error. (e.g. #751)


kou:
Thanks!

PR #759, opened 2017-08-31
Don't keep created 0byte file on error

kenhys:


lib/io.c
```diff
@@ -114,6 +114,7 @@ inline static grn_rc grn_pread(grn_ctx *ctx, fileinfo *fi, void *buf,
                                size_t count, off_t offset);
 inline static grn_rc grn_pwrite(grn_ctx *ctx, fileinfo *fi, void *buf,
                                 size_t count, off_t offset);
+inline static grn_bool grn_path_has_existed(const char *path);
```
kou: Why `has_existed`?
Is `grn_path_exist` enough?

lib/io.c
```diff
@@ -1224,8 +1225,20 @@ grn_io_seg_map_(grn_ctx *ctx, grn_io *io, uint32_t segno, grn_io_mapinfo *info)
     if (!grn_fileinfo_opened(fi)) {
       char path[PATH_MAX];
       gen_pathname(io->path, path, fno);
+      grn_bool path_exists = GRN_TRUE;
```
kou: We don't use "third person singular" form for variable name.
You can use `path_exist`.

kou: We define all variables at the top of block as I mentioned before.

lib/io.c
```diff
@@ -1224,8 +1225,20 @@ grn_io_seg_map_(grn_ctx *ctx, grn_io *io, uint32_t segno, grn_io_mapinfo *info)
...
+      if (!grn_path_has_existed(path)) {
+        path_exists = GRN_FALSE;
+      }
```
kou: It's redundant.

```c
grn_bool path_exist;

path_exist = grn_path_exist(path);
```

lib/io.c
```diff
@@ -1224,8 +1225,20 @@ grn_io_seg_map_(grn_ctx *ctx, grn_io *io, uint32_t segno, grn_io_mapinfo *info)
...
       if (!grn_fileinfo_open(ctx, fi, path, O_RDWR|O_CREAT)) {
         DO_MAP(io, &info->fmo, fi, pos, segment_size, segno, info->map);
+        if (!info->map && path_exists == GRN_FALSE) {
```
kou: `!path_exist`

lib/io.c
```diff
@@ -1224,8 +1225,20 @@ grn_io_seg_map_(grn_ctx *ctx, grn_io *io, uint32_t segno, grn_io_mapinfo *info)
...
+          if (grn_unlink(path) == 0) {
+            GRN_LOG(ctx, GRN_LOG_INFO,
+                    "[io][map][error] removed 0byte file: <%s>", path);
```
kou: Why does it use `0byte` not `empty`?

lib/io.c
```diff
@@ -1548,6 +1561,20 @@ grn_io_get_disk_usage(grn_ctx *ctx, grn_io *io)
...
 }
 
+grn_bool
+grn_path_has_existed(const char *path)
+{
+  if (!path) {
+    return GRN_FALSE;
+  }
+
+  struct stat statbuf;
```
kou: We define variables at the top of block.

lib/io.c
```diff
@@ -1548,6 +1561,20 @@ grn_io_get_disk_usage(grn_ctx *ctx, grn_io *io)
...
+  if (!stat(path, &statbuf)) {
```
kou: It's redundant.

```c
return stat(path, &statbuf) == 0;
```

lib/io.c
kou: We use `_` as word separator for variable name.

kenhys:
I've fixed some feedback points.

* Use `grn_path_exist` utility function.
* Change variable name from `path_exists` to `path_exist`
* Use `!path_exist` conditional expression instead of `path_exist == GRN_FALSE`
* Use `empty` in message instead of `0byte`.

PR #761, opened 2017-09-14
Support --default-command-version 3

kenhys:
In the previous versions, --default-command-version 3 is not supported.

Before:

  % groonga --default-command-version 3 -n db/db
  invalid command version: <3>

After:

  % groonga --default-command-version 3 -n db/db
  >

### DIFF
--- a/src/groonga.c
+++ b/src/groonga.c
@@ -3601,18 +3601,7 @@ main(int argc, char **argv)
               default_command_version_arg);
       return EXIT_FAILURE;
     }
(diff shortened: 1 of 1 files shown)


kou:
We don't need the whole `switch`. We can just assign `value` to `default_command_version`.

kenhys:
I see.
It is good enough to assign `value` to `default_comand_version`. The range of `value` is checked in beforehand.

kou:
Merged.

PR #762, opened 2017-09-14
Support to log location uri as is in query log

kenhys:
In the previous versions, even though "location /cmd/xxx"
 configuration is set in groonga-httpd.conf, "/cmd/xxx" query is
 always converted into "/d/xxx" in query log.
It may be confusing situation by such a implicit conversion.

Before:
  /cmd/status is logged as /d/status in query log.

After:
  If "groonga_log_location_uri on;" is set, /cmd/status is
  logged as /cmd/status in query log.

### DIFF
--- a/lib/ctx.c
+++ b/lib/ctx.c
@@ -1054,6 +1054,7 @@ grn_ctx_qe_exec_uri(grn_ctx *ctx, const char *path, uint32_t path_len)
   grn_obj request_id;
   double request_timeout;
   const char *p = path, *e = path + path_len, *v, *key_end, *filename_end;
+  const char *slash_delimiter;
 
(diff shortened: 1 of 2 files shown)


kou:
Is `examples/dictionary/` still workable with this change and `groonga --protocol http`?

kenhys:
Yes, it's workable with `groonga --protocol http` in above examples.

kou:
Really? Did you really get `/js/dictionary.js`? I think that Groonga processes it as a `dictionary` command instead of static file. It breaks backward compatibility.

kou:
Do you still really want this feature?

kenhys:
Ah, I did not take into account that case.
I don't want to break backward compatibility unexpectedly, so It should be close this feature.

PR #763, opened 2017-09-28
dat: limit the number of nodes per key

s-yata:
#### Description

This change limits the number of nodes per key of a dat trie.
If it is unlimited, `grn_dat_rebuild_trie` creates a huge file in the worst case.


kou:
Merged.

PR #764, opened 2017-10-04
io: set proper log level to DEBUG

kenhys:
It should be DEBUG because it doesn't cause problem by itself.
This message is logged when mapped area for indexes is unmapped.


kou:
Merged.

PR #768, opened 2017-10-10
plugin suggest: add missing NULL check for grn_ctx_at()

kenhys:


plugins/suggest/suggest.c
```diff
@@ -657,10 +657,22 @@ learner_init_columns(grn_ctx *ctx, grn_suggest_learner *learner)
   grn_obj *seqs, *events, *post_item, *items, *pairs;
 
   learner->seqs = seqs = grn_ctx_at(ctx, GRN_OBJ_GET_DOMAIN(learner->seq));
+  if (!learner->seqs) {
+    ERR(GRN_DOMAIN_ERROR,
```
kou: Why `GRN_DOMAIN_ERROR`?

It's based on `EDOM`:

>        EDOM            Mathematics argument out of domain of function
>                       (POSIX.1, C99).

http://man7.org/linux/man-pages/man3/errno.3.html

plugins/suggest/suggest.c
```diff
@@ -657,10 +657,22 @@ learner_init_columns(grn_ctx *ctx, grn_suggest_learner *learner)
...
+        GRN_OBJ_GET_DOMAIN(learner->seq));
+    return;
+  }
   learner->seqs_events = grn_obj_column(ctx, seqs, CONST_STR_LEN("events"));
 
   events_id = grn_obj_get_range(ctx, learner->seqs_events);
   learner->events = events = grn_ctx_at(ctx, events_id);
+  if (!learner->events) {
+    ERR(GRN_UNKNOWN_ERROR,
+        "failed to initialize columns because event id: <%d> is not found.",
```
kou: The last "." is needless.

kou: `: <...>` should be placed at the last. We should not put it the middle of sentence.

`"failed to refer sequence's event table: <%.*s>: <%u>(<%.*s>)", seqs_events_name_size, seqs_events_name, events_id, events_name_size, events_name)"`

`events_name` will be able to get by `grn_table_get_key()` against `grn_ctx_db()`.

PR #773, opened 2017-10-25
Added feature that sort the results of `logical_select` in descending…

komainu8:
… order.

### DIFF
--- a/lib/mrb/scripts/table.rb
+++ b/lib/mrb/scripts/table.rb
@@ -102,7 +102,7 @@ module Groonga
         order    = key[:order] || order
         offset   = key[:offset] || offset
       else
-        key_name = key
+        key_name = key.dup
       end
 
       case key_name
--- a/lib/mrb/scripts/table.rb
+++ b/lib/mrb/scripts/table.rb
@@ -102,7 +102,7 @@ module Groonga
         order    = key[:order] || order
         offset   = key[:offset] || offset
       else
-        key_name = key.dup
+        key_name = key
       end
 
       case key_name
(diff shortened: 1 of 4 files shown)


lib/mrb/scripts/table.rb
```diff
@@ -102,7 +102,7 @@ def ensure_sort_key(key, converted_keys)
         order    = key[:order] || order
         offset   = key[:offset] || offset
       else
-        key_name = key
+        key_name = key.dup
```
kou: We should not use `#dup`. We should stop to update `key_name` directory by `key_name[0] = ""` instead. `key_name = key_name[1..-1]` is alternative way.

plugins/sharding/logical_select.rb
```diff
@@ -152,6 +152,11 @@ def write_records(writer, context)
             :offset => current_offset,
             :limit => current_limit,
           }
+          context.sort_keys.each do |sort_key|
+            if sort_key.match("-timestamp")
```
kou: We shouldn't assume `timestamp` as sort key.
It should be `sort_key.start_with?("-")`.

kou: If `sort_keys` have two `-XXX`, `result_sets` isn't reversed.
It should be:

```ruby
if context.sort_keys.any? {|sort_key| sort_key.start_with?("-")}
  result_sets = result_sets.reverse
end
```

or something.

komainu8:
I've fied them below commit.
https://github.com/groonga/groonga/pull/773/commits/abcb4f41a6ae47c30b110c62710204bf9f2446ab

kou:
Merged.

PR #777, opened 2017-11-09
Add parse groonga log

komainu8:
Todo:

- [x] Modified test

### DIFF
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -1,3 +1,5 @@
+require "groonga-log"
+
 module Groonga
   module CommandLine
     class Grndb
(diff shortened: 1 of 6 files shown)


lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -26,16 +28,30 @@ def create_command_line_parser
         program_name = File.basename(@program_path)
         parser = CommandLineParser.new(program_name)
 
-        parser.add_command("check") do |command|
-          command.description = "Check database"
-
-          options = command.options
-          options.banner += " DB_PATH"
-          options.string("--target", "Check only the target object.")
+        if @arguments[1] == "--groonga-log" then
```
kou: We don't use `then`.

kou:
I've implemented with different approach. Please check  8560e44.

PR #785, opened 2017-12-05
delete: log deletion count and error count into query-log

komainu8:


kou:
We should count one by one instead of using `grn_table_size()` because `grn_table_size()` may return wrong value when `load` or `delete` are executed in the same time.

And this pull request needless commit for `logical_select`.

PR #788, opened 2017-12-06
Fix a memory leak in func_highlight_full

komainu8:


lib/proc/proc_highlight.c
```diff
@@ -374,6 +374,7 @@ func_highlight_full(grn_ctx *ctx, int nargs, grn_obj **args,
                                            string, keywords,
                                            use_html_escape);
     }
+    grn_obj_unlink(ctx, keywords);
```
kou: It should be in `if (keywords) {`. If `keywords` is `NULL`, this causes a crash.

komainu8: Thank you for your comment.
I've fixed it  in ac723de53ac2bd56e927c3568526e90bccf67b7a

PR #790, opened 2017-12-21
Add index_column_source_records() function

naoa:
I would like to add function to retrieve source records of index column in order to efficiently get value of referenced column such as parent-child relationship.

Syntax:
```c
  index_column_source_records(index_column_name[, limit])
```

*  index_column_name: name of index columns as string
* limit(option): number of records to get (default: -1, -1 means all records)


kou:
Thanks.

`index_column_source_records(index_column_name[, {"limit": -1}])` may be better because we may want `offset` parameter. Can you try it?

naoa:
Thanks for merging. Maybe I can do it tonight.

PR #802, opened 2017-12-28
doc: Fix a tag

komainu8:


kou:
Why? The original tag is correct.

komainu8:
I'm sorry.
I had a misunderstanding.

PR #807, opened 2018-01-10
grndb: add --force-lock-clear option

komainu8:
This option force to clear lock of object(Database, Column, Table).
You can use if you want to database recover even if remain locked of database.

But this option very risky.
Because can not safty recover database if remain locked of database.

Usage:
  grndb recover --force-lock-clear DB_PATH


kou:
いくつか直しておいたので確認しておいてください。

PR #810, opened 2018-01-23
between: accept between() without borders

### LINKED ISSUE #685: between() is redundant
[between()](http://groonga.org/docs/reference/functions/between.html) API is redundant.
I want to write simple like [groonga-clinet's between()](https://github.com/ranguba/groonga-client/blob/master/lib/groonga/client/request/select.rb#L335).

```
--filter 'between(column, min, max)'
```

is short hand for:

```
--filter 'between(column, min, "include", max, "include")
```

s-yata:
If the number of arguments passed to between() is 3, the 2nd and 3rd arguments are handled as the inclusive edges.

GitHub: fix #685 

----

I'm not sure this change is acceptable.


s-yata:
Thank you!

PR #814, opened 2018-01-30
doc: Add a note about dropped Ubuntu 17.04 to 7.1.1 news

myokoym:
ref: 58779cd70c9b2e473052b42a3cb976c450b37b53


kou:
Good catch!

PR #823, opened 2018-02-16
logical_count: add post_filter

komainu8:


kou:
This implementation has some problems:

  * `filtered` stage dynamic columns are evaluated before `filter`
  * Any dynamic columns don't exist in `post_filter` when `filter` isn't specified
  * `post_filter` is ignored when `min` or `max` is specified and `filter` isn't specified

I've implemented.

PR #828, opened 2018-02-20
doc logical_range_filter: add execution examples of logical_table

komainu8:


kou:
We need to update file list by `make -C doc update-files` when we add/remove files.

PR #830, opened 2018-03-03
file_lock: Copy path to buffer to avoid storing a pointer to a local buffer.

genki:
`grn_file_lock` should have path buffer in its struct to prevent storing a pointer to a stack local buffer.


kou:
Thanks!

PR #831, opened 2018-03-03
Generational lock

genki:
This PR wants to introduce a more robust way to recovery from crash in comparision to `lock_clear`.

Unfortunately, all software systems complicated enough can crash. Groonga is too.
Crashed Groonga processes could remain some locks in its database. Previously, so we have to unlock them by calling `lock_clear`.
In production system, it is not acceptable there to be down time even while clearing locks manually.
By this PR, we can automatically recover the dead lock by introducting generantional lock to the io lock.

Every `grn_io_lock`/`grn_io_unlock` works with generation ID which is uniquely related with a process.
generation ID is only effective while a process is alive. It is achieved by using `flock`/`LockFileEx`.

At the starting up, check running processes and its generation IDs.
If lock failed, extract the generation ID from the lock and test if the ID is in the available ID table. If not, the lock was made by crashed process, so we clear it.

### DIFF
--- a/lib/c_sources.am
+++ b/lib/c_sources.am
@@ -110,4 +110,6 @@ libgroonga_c_sources =				\
 	window_function.c			\
 	grn_window_function.h			\
 	window_functions.c			\
-	grn_window_functions.h
+	grn_window_functions.h \
+  gen.c \
+  grn_gen.h
--- a/lib/ctx.c
+++ b/lib/ctx.c
@@ -35,6 +35,7 @@
 #include "grn_logger.h"
 #include "grn_cache.h"
 #include "grn_expr.h"
+#include "grn_gen.h"
 #include <stdio.h>
 #include <stdarg.h>
 #include <time.h>
@@ -1545,6 +1546,7 @@ grn_ctx_use(grn_ctx *ctx, grn_obj *db)
         grn_obj_get_info(ctx, db, GRN_INFO_ENCODING, &buf);
         ctx->encoding = *(grn_encoding *)GRN_BULK_HEAD(&buf);
         grn_obj_close(ctx, &buf);
+        ctx->rc = grn_gen_init(ctx, db);
       }
     }
   }
--- a/lib/db.c
+++ b/lib/db.c
@@ -217,6 +217,7 @@ grn_db_create(grn_ctx *ctx, const char *path, grn_db_create_optarg *optarg)
   s->specs = NULL;
   s->config = NULL;
   s->cache = NULL;
+  s->gen = NULL;
 
   {
     grn_bool use_default_db_key = GRN_TRUE;
@@ -335,6 +336,7 @@ grn_db_open(grn_ctx *ctx, const char *path)
   s->specs = NULL;
   s->config = NULL;
   s->cache = NULL;
+  s->gen = NULL;
 
   {
     uint32_t type = grn_io_detect_type(ctx, path);
@@ -448,6 +450,9 @@ grn_db_close(grn_ctx *ctx, grn_obj *db)
   GRN_API_ENTER;
 
   ctx_used_db = ctx->impl && ctx->impl->db == db;
+  if (ctx_used_db) {
+    grn_gen_fin(ctx, s);
+  }
   if (ctx_used_db) {
     grn_ctx_loader_clear(ctx);
     if (ctx->impl->parser) {
--- a/lib/file_lock.c
+++ b/lib/file_lock.c
@@ -42,7 +42,7 @@ grn_file_lock_init(grn_ctx *ctx,
                    grn_file_lock *file_lock,
                    const char *path)
 {
-  file_lock->path = path;
+  grn_strcpy(file_lock->path, PATH_MAX, path);
 #ifdef WIN32
   file_lock->handle = INVALID_HANDLE_VALUE;
 #else /* WIN32 */
@@ -109,7 +109,7 @@ grn_file_lock_release(grn_ctx *ctx, grn_file_lock *file_lock)
 
   file_lock->fd = -1;
 #endif /* WIN32 */
-  file_lock->path = NULL;
+  grn_strcpy(file_lock->path, PATH_MAX, "");
 }
 
 void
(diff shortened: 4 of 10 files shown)


s-yata:
### Abstract

The idea seems to be good.
The implicit `lock_clear` is helpful for not only you but also other users.

However, this pull request is not acceptable due to several problems.

### Major problems

#### Multi-thread safety

The provided implementation is dangerous if the same DB is opened by multiple processes or threads.
For example, `grn_io_clear_lock()` assumes that other threads and processes don't attempt to lock the `grn_io` at the same time, because `*io->lock = 0` might be executed between `GRN_ATOMIC_ADD_EX(io->lock, 1, lock)` and `GRN_ATOMIC_ADD_EX(io->lock, -1, lock)`.

If there is only one process with one thread, `lock_clear` works well.

#### Initilization

`grn_obj_clear_lock()` should not be called before `grn_db_init_builtin_types()`.

### Minor problems

#### Compilation errors

Compilation for Windows fails as follows:

```
C:\projects\groonga\lib\file_lock.c(141): error C2065: 'TESTSTRLEN' : undeclared identifier [C:\projects\groonga\lib\libgroonga.vcxproj]
4972
C:\projects\groonga\lib\file_lock.c(172): error C2065: 'TESTSTRLEN' : undeclared identifier [C:\projects\groonga\lib\libgroonga.vcxproj]
```

#### Compilation warnings

`#include`s are missing.

```
file_lock.c: In function ‘grn_file_lock_takeover’:
file_lock.c:143:7: warning: implicit declaration of function ‘flock’ [-Wimplicit-function-declaration]
   if (flock(file_lock->fd, LOCK_EX|LOCK_NB) != 0) {
       ^
  CC       io.lo
```

```
io.c: In function ‘grn_io_lock’:
io.c:1379:10: warning: implicit declaration of function ‘grn_gen_lock’ [-Wimplicit-function-declaration]
     if (!grn_gen_lock(ctx, io, count)) {
          ^
io.c: In function ‘grn_io_unlock’:
io.c:1417:5: warning: implicit declaration of function ‘grn_gen_unlock’ [-Wimplicit-function-declaration]
     grn_gen_unlock(io);
     ^
  CC       gen.lo
```

A typecast is missing.

```
gen.c: In function ‘grn_gen_init’:
gen.c:122:29: warning: passing argument 2 of ‘grn_obj_clear_lock’ from incompatible pointer type [-Wincompatible-pointer-types]
     grn_obj_clear_lock(ctx, db);
                             ^
In file included from ../include/groonga.h:22:0,
                 from grn.h:743,
                 from gen.c:19:
../include/groonga/groonga.h:767:16: note: expected ‘grn_obj * {aka struct _grn_obj *}’ but argument is of type ‘grn_db * {aka struct _grn_db *}’
 GRN_API grn_rc grn_obj_clear_lock(grn_ctx *ctx, grn_obj *obj);
                ^
```

#### Interfaces

Local functions should be `static`.

```
int
gen_id_is_newer(grn_gen *gen, int id)
{
  // ...
}
```

If a function has no argument, `void` should be specified in its declaration.

```c
void grn_gen_enable();
```

Functions `grn_gen_*()` should be defined for `grn_gen`.

```c
grn_rc grn_gen_init(grn_ctx *ctx, grn_db *db);
grn_rc grn_gen_fin(grn_ctx *ctx, grn_db *db);
int grn_gen_lock(grn_ctx *ctx, grn_io *io, uint32_t count);
int grn_gen_unlock(grn_io *io);
```

The difference between `grn_file_lock_fin()` and `grn_file_lock_close()` is not clear.
`*_close()` is generally used in combination with `*_open()` but `grn_file_lock_open()` does not exist.

```c
void grn_file_lock_fin(grn_ctx *ctx, grn_file_lock *file_lock);
void grn_file_lock_close(grn_ctx *ctx, grn_file_lock *file_lock);
```

#### Performance

`grn_gen_init()` checks `GRN_GEN_SIZE` (2,048) file locks in initialization but most checks are not required in practice.

```c
for(int i = 1; i < GRN_GEN_SIZE; i++){
  gen_path_fill(ctx, db, i, gen_path);
  grn_file_lock file_lock;
  grn_file_lock_init(ctx, &file_lock, gen_path);
  if (grn_file_lock_exist(ctx, &file_lock)) {
    // ...
  }
  grn_file_lock_fin(ctx, &file_lock);
}
```

`grn_gen_unlock()` requires an additional atomic operation even if the generational lock is disabled.

```c
GRN_ATOMIC_ADD_EX(io->lock, -1, lock);
if ((lock & grn_gen_lock_count_mask) == 1) {
  GRN_ATOMIC_ADD_EX(io->lock, -(lock & GEN_ID_MASK), lock);
  return 1;
}
```

genki:
Thank you for reviewing.

Sorry for not tested yet on Windows environment because currently I have no Windows machines. This PR is still WIP and I have understood there must be more improvements. My motivation was offering a concept in advance of further works.

So, let me think about individual issues.
At first, `grn_io_lock_clear` is not thread safe. Yes, you right.
The purpose is ignoring the locks made by crashed process.
If the crashed locker is one process, there is no doubt to be only single overlocking for each dead locks. If there are more crashed lockers (at the same time!), even althrough, the overlocking count still must be one.
So we can use `GRN_ATOMIC_ADD_EX(io->lock, -1, lock)` for clearing the lock instead of `grn_io_lock_clear`.
It is going to be thread safe.
If I won't use `grn_io_lock_clear`, the next issue relating with the initialization is solved as well.

Succeeding minor issues. Yeah, I will fix them. It's ok.
I am sorry in advance for my blindness in Windows environment.

The last, performance issue. I knew it can be solved if we use directory reading API or such like system calls. But I had no self-confidence to use it in WIN32 block according previously I mentioned.
Not only for that, I had tested it and found there is virtually no time to check existense of the files. In addition, I have experience which the listing up directory entries is relatively high cost compared to testing file existance, in particular, if there are many items in it.
I don't know which way is low cost.

One more option to get rid of that cost is using `grn_config_get/set` to remember the paths of lock files.
Let say what messy is, it uses `grn_io_lock` internally. It draws unnecessary complication. Preparing a special file for recording the paths can achive same thing without complication.

genki:
Fixed issues about thread-safety and minor issues.

genki:
Uses the id list file for remembering which generation id is acquired and safely released.
So now file existence test is only applied for files listed.

PR #832, opened 2018-03-15
logical_range_filter: Support decimal point

komainu8:


kou:
Merged but it's better that we should use test data that care msec.

For example:

  * Data: `2018/03/15 11:11:02.001` and `2018/03/15 11:11:02.002`
  * Min/Max: `2018/03/15 11:11:02.001` with `excluded` border

PR #836, opened 2018-04-04
Add Groonga docker installation document

cosmo0920:
Groonga Alpine Linux minimal image is distributed on DockerHub.
Should we describe this minimal image on document?


kou:
Thanks!

PR #840, opened 2018-04-23
ii: add log enhancements that output contents of a token

komainu8:
I added ii's log enhancements that not only token id but also output contents of a token.

### DIFF
--- a/include/groonga/ii.h
+++ b/include/groonga/ii.h
@@ -55,8 +55,6 @@ GRN_API grn_rc grn_ii_posting_add(grn_ctx *ctx, grn_posting *pos,
                                   grn_hash *s, grn_operator op);
 GRN_API void grn_ii_resolve_sel_and(grn_ctx *ctx, grn_hash *s, grn_operator op);
 
-GRN_API void grn_ii_get_token_from_token_id(grn_ctx *ctx, grn_ii *ii, grn_id tid, grn_obj *token);
-
 /* Experimental */
 typedef struct _grn_ii_cursor grn_ii_cursor;
 GRN_API grn_ii_cursor *grn_ii_cursor_open(grn_ctx *ctx, grn_ii *ii, grn_id tid,
--- a/include/groonga/ii.h
+++ b/include/groonga/ii.h
@@ -55,6 +55,7 @@ GRN_API grn_rc grn_ii_posting_add(grn_ctx *ctx, grn_posting *pos,
                                   grn_hash *s, grn_operator op);
 GRN_API void grn_ii_resolve_sel_and(grn_ctx *ctx, grn_hash *s, grn_operator op);
 
+GRN_API void grn_ii_get_token_from_token_id(grn_ctx *ctx, grn_ii *ii, grn_id tid, grn_obj *token);
 
 /* Experimental */
 typedef struct _grn_ii_cursor grn_ii_cursor;
--- a/include/groonga/ii.h
+++ b/include/groonga/ii.h
@@ -55,6 +55,7 @@ GRN_API grn_rc grn_ii_posting_add(grn_ctx *ctx, grn_posting *pos,
                                   grn_hash *s, grn_operator op);
 GRN_API void grn_ii_resolve_sel_and(grn_ctx *ctx, grn_hash *s, grn_operator op);
 
+
 /* Experimental */
 typedef struct _grn_ii_cursor grn_ii_cursor;
 GRN_API grn_ii_cursor *grn_ii_cursor_open(grn_ctx *ctx, grn_ii *ii, grn_id tid,
--- a/lib/ii.c
+++ b/lib/ii.c
@@ -110,7 +110,6 @@ grn_ii_get_token_from_token_id(grn_ctx *ctx, grn_ii *ii,
 {
    char key[GRN_TABLE_MAX_KEY_SIZE];
    int key_size;
-   GRN_TEXT_INIT(token, 0);
    key_size = grn_table_get_key(ctx, ii->lexicon, tid,
                                 key, GRN_TABLE_MAX_KEY_SIZE);
    if (key_size != 0) {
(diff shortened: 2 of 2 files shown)


lib/ii.c
```diff
@@ -13064,3 +13173,21 @@ grn_ii_build2(grn_ctx *ctx, grn_ii *ii, const grn_ii_builder_options *options)
   }
   return rc;
 }
+
+void
+grn_ii_get_token_from_token_id(grn_ctx *ctx, grn_ii *ii,
```
kou: It should not be exported. It should be a static function.

lib/ii.c
```diff
@@ -13064,3 +13173,21 @@ grn_ii_build2(grn_ctx *ctx, grn_ii *ii, const grn_ii_builder_options *options)
...
+                               grn_id tid, grn_obj *token)
+{
+   grn_obj key_buf;
```
kou: We can reduce scope of this local variable.

lib/ii.c
```diff
@@ -13064,3 +13173,21 @@ grn_ii_build2(grn_ctx *ctx, grn_ii *ii, const grn_ii_builder_options *options)
...
+   char key[GRN_TABLE_MAX_KEY_SIZE];
+   int key_size;
+   GRN_TEXT_INIT(token, 0);
```
kou: It should be done by caller.

lib/ii.c
```diff
@@ -13064,3 +13173,21 @@ grn_ii_build2(grn_ctx *ctx, grn_ii *ii, const grn_ii_builder_options *options)
...
+   key_size = grn_table_get_key(ctx, ii->lexicon, tid,
+                                key, GRN_TABLE_MAX_KEY_SIZE);
+   if (key_size != 0) {
+     GRN_OBJ_INIT(&key_buf, GRN_BULK, 0, ii->lexicon->header.domain);
```
kou: We can use `GRN_OBJ_DO_SHALLOW_COPY` instead of `0` to reduce memory copy.

lib/ii.c
```diff
@@ -12962,14 +13066,19 @@ grn_ii_builder_register_chunks(grn_ctx *ctx, grn_ii_builder *builder)
...
          "failed to allocate an array in segment: "
          "<%.*s>: "
-         "tid=<%u>: max_n_segments=<%u>",
+         "tid=<%u>: max_n_segments=<%u>: "
+         "token:<%.*s>",
          name_size, name,
          builder->chunk.tid,
-         builder->ii->seg->header->max_segment);
+         builder->ii->seg->header->max_segment,
+         GRN_TEXT_LEN(&token), GRN_TEXT_VALUE(&token));
```
kou: We should cast `GRN_TEXT_LEN(&token)` to `int` to suppress a warning on some environment.

kou: https://travis-ci.org/groonga/groonga/jobs/369929306#L1549

> ```text
> ii.c:2912:7: warning: field precision specifier ‘.*’ expects argument of type ‘int’, but argument 7 has type ‘long int’ [-Wformat=]
>       MERGE_BC(bid.rid <= rid || cid.rid);
```

Did you see this warning on your environment?
       ^

File: lib/ii.c [hunk 6]
```diff
@@ -12962,14 +13066,19 @@ grn_ii_builder_register_chunks(grn_ctx *ctx, grn_ii_builder *builder)
 
   a = array_get(ctx, builder->ii, builder->chunk.tid);
   if (!a) {
+    grn_obj token;
+    grn_ii_get_token_from_token_id(ctx, builder->ii, builder->chunk.tid, &token);
```
kou: It'll cause "variable definitions after expression".
We should call it after ` DEFINE_NAME`.

komainu8:
I've fixed them.

include/groonga/ii.h
```diff
@@ -55,7 +55,6 @@ GRN_API grn_rc grn_ii_posting_add(grn_ctx *ctx, grn_posting *pos,
                                   grn_hash *s, grn_operator op);
 GRN_API void grn_ii_resolve_sel_and(grn_ctx *ctx, grn_hash *s, grn_operator op);
 
-
 /* Experimental */
```
kou: This is a needless change.

lib/ii.c
```diff
@@ -4566,20 +4591,32 @@ grn_ii_update_one(grn_ctx *ctx, grn_ii *ii, grn_id tid, grn_ii_updspec *u, grn_h
...
   if (!(a = array_get(ctx, ii, tid))) {
     DEFINE_NAME(ii);
+    grn_obj token;
+    GRN_TEXT_INIT(&token, 0);
+    grn_ii_get_token_from_token_id(ctx, ii, tid, &token);
     MERR("[ii][update][one] failed to allocate an array: "
          "<%.*s>: "
-         "<%u>:<%u>:<%u>",
+         "<%u>:<%u>:<%u>"
+         "<%.*s>: ",
```
kou: `: ` should be places after `"...:<%u>"`

lib/ii.c
```diff
@@ -4566,20 +4591,32 @@ grn_ii_update_one(grn_ctx *ctx, grn_ii *ii, grn_id tid, grn_ii_updspec *u, grn_h
...
   if (!(bs = encode_rec(ctx, ii, u, &size, 0))) {
     DEFINE_NAME(ii);
+    grn_obj token;
+    GRN_TEXT_INIT(&token, 0);
+    grn_ii_get_token_from_token_id(ctx, ii, tid, &token);
     MERR("[ii][update][one] failed to encode a record: "
          "<%.*s>: "
-         "<%u>:<%u>:<%u>",
+         "<%u>:<%u>:<%u>"
+         "<%.*s>: ",
```
kou: Ditto.

lib/ii.c
```diff
@@ -4588,13 +4625,19 @@ grn_ii_update_one(grn_ctx *ctx, grn_ii *ii, grn_id tid, grn_ii_updspec *u, grn_h
         pos = a[0];
         if ((pseg = buffer_open(ctx, ii, pos, &bt, &b)) == GRN_II_PSEG_NOT_ASSIGNED) {
           DEFINE_NAME(ii);
+          grn_obj token;
```
kou: We should not put it after `DEFINE_NAME` because it has code not variable definitions.

kou: It should be:

```c
grn_obj token;
DEFINE_NAME(ii);
```

komainu8:
I've fixed them again.

PR #841, opened 2018-04-24
Translate 8.0.2 news

piroor:
I've translated news for Groonga 8.0.2.


piroor:
Thanks, I've updated the po file.

PR #843, opened 2018-05-19
doc: correct command lines of groonga executable in the client mode

noritada:
"--host" option is not available in the client mode.


kou:
Good catch!

PR #844, opened 2018-05-20
doc: correct a description on protocols available in the client mode.

noritada:
As far as I understand from a description as follows (and from source code and behavior),
the client mode only supports GQTP and HTTP is not supported.
Users cannot specify a protocol to access the server.

> In client mode, ``groonga`` executable file runs as a client for GQTP
> protocol Groonga server.

Sorry if my understanding is not correct.
Thanks!


kou:
Good catch!
You're right.

PR #850, opened 2018-06-19
Add check exist of "WITH SECTION" flag when a full-text search for vector column

komainu8:
Because we can't a full-text search for vector column if we don't set "WITH_SECTION" flag.


test/command/suite/load/index/offline/vector/int32_position.expected
```diff
@@ -11,7 +11,7 @@ load --table Entries
 [[0,0.0,0.0],2]
 table_create Numbers TABLE_PAT_KEY Int32
 [[0,0.0,0.0],true]
-column_create Numbers entries_numbers COLUMN_INDEX|WITH_POSITION Entries numbers
+column_create Numbers entries_numbers COLUMN_INDEX|WITH_POSITION|WITH_SECTION Entries numbers
```
kou: `WITH_POSITION` is needless because it's not a full text search index.

lib/db.c
```diff
@@ -8818,9 +8818,47 @@ grn_obj_set_info_source_validate(grn_ctx *ctx, grn_obj *obj, grn_obj *value)
...
+  }
 
   source_ids = (grn_id *)GRN_BULK_HEAD(value);
   n_source_ids = GRN_BULK_VSIZE(value) / sizeof(grn_id);
+  if (grn_obj_is_index_column(ctx, obj)) {
+    if (n_source_ids == 1) {
+      grn_obj *source;
+
+      source = grn_ctx_at(ctx, source_ids[0]);
+      switch (source->header.type) {
```
kou: Can we use `grn_obj_is_vector_column()`?

I've fixed it in 197d92f.

kou:
Merged.

PR #852, opened 2018-06-20
test: remove needless flag

komainu8:
Because it does not have to be a full-text search index.

### DIFF
--- a/test/command/suite/load/index/offline/vector/text_without_section.expected
+++ b/test/command/suite/load/index/offline/vector/text_without_section.expected
@@ -11,7 +11,7 @@ load --table Docs
 [[0,0.0,0.0],1]
 table_create Words TABLE_PAT_KEY ShortText --default_tokenizer TokenBigram
 [[0,0.0,0.0],true]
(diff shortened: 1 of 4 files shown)


test/command/suite/load/index/offline/vector/text_without_section.expected
```diff
@@ -11,7 +11,7 @@ load --table Docs
 [[0,0.0,0.0],1]
 table_create Words TABLE_PAT_KEY ShortText --default_tokenizer TokenBigram
 [[0,0.0,0.0],true]
-column_create Words docs_sentences COLUMN_INDEX|WITH_POSITION Docs sentences
+column_create Words docs_sentences COLUMN_INDEX Docs sentences
```
kou: It uses `TokenBigram`. So it's an index for full text search.

komainu8: Oh...
In that case, these tests should be an error.

I'll fix them and commit to #850.
Because these fixes are related #850's modify.
So, I close this pull request.

PR #854, opened 2018-06-20
test: remove unused variables

komainu8:
When building Groonga,  I found an below warning.
So, I removed these variables.

```
test-string.c:72:20: warning: ‘normalized_text_ja_utf8’ defined but not used [-Wunused-const-variable=]
 static const gchar normalized_text_ja_utf8[] =
                    ^~~~~~~~~~~~~~~~~~~~~~~
test-string.c:64:20: warning: ‘text_ja_utf8’ defined but not used [-Wunused-const-variable=]
 static const gchar text_ja_utf8[] =
                    ^~~~~~~~~~~~
```


kou:
Good catch!

PR #859, opened 2018-07-20
Use POSIX compatible futimens method

siepkes:
`futimes` (ie. without the 'n') is the non-POSIX compatible brother of the POSIX compatible method `futimens`. `futimes` Is only implemented on Linux and some of the BSD's. Other POSIX compliant platforms do not provide this method but do provide `futimens`.

This came up when I was trying to build groonga on SmartOS (ie. Illumos / Solaris).

See also: http://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html

### DIFF
--- a/build/ac_macros/check_functions.m4
+++ b/build/ac_macros/check_functions.m4
@@ -3,6 +3,8 @@
 AC_CHECK_FUNCS(_gmtime64_s)
 AC_CHECK_FUNCS(_localtime64_s)
 AC_CHECK_FUNCS(_strtoui64)
+AC_CHECK_FUNCS(futimens)
+AC_CHECK_FUNCS(futimes)
 AC_CHECK_FUNCS(gmtime_r)
 AC_CHECK_FUNCS(localtime_r)
 AC_CHECK_FUNCS(mkstemp)
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -139,6 +139,8 @@
 #cmakedefine HAVE_CLOCK
 #cmakedefine HAVE_CLOCK_GETTIME
 #cmakedefine HAVE_FPCLASSIFY
+#cmakedefine HAVE_FUTIMENS
+#cmakedefine HAVE_FUTIMES
 #cmakedefine HAVE_GMTIME_R
 #cmakedefine HAVE_LOCALTIME_R
 #cmakedefine HAVE_MKSTEMP
--- a/lib/io.c
+++ b/lib/io.c
@@ -49,6 +49,18 @@
 # define GRN_IO_FILE_SIZE_V0  GRN_IO_FILE_SIZE_V1
 #endif /* WIN32 */
 
+#ifndef WIN32
+# ifdef HAVE_FUTIMENS
+#  define grn_futimens(fd, times) futimens((fd), (times))
+#  define grn_futimens_name "futimens"
+# elif defined(HAVE_FUTIMES) /* HAVE_FUTIMENS */
+#  define grn_futimens(fd, times) futimes((fd), (times))
+#  define grn_futimens_name "futimes"
+# else /* HAVE_FUTIMENS */
+#  error Support for either the futimens() or futimens() function is required
+# endif /* HAVE_FUTIMENS */
+#endif /* WIN32 */
+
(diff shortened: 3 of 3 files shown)


siepkes:
Interestingly enough it seems the build for OS X breaks. From what I can find OS X only supports `futimens` starting from OS X 10.13.

In that case, would you guys rather do this with an `ifdef` ?

kou:
It seems that it's better that we check `futimes` and `futimens` availability in `configure` (and `cmake`). We will be able to use `HAVE_FUNTIMENS` and `HAVE_FUTIMES` by the following change:

```diff
diff --git a/build/ac_macros/check_functions.m4 b/build/ac_macros/check_functions.m4
index a0b424b31..8c2081dbb 100644
--- a/build/ac_macros/check_functions.m4
+++ b/build/ac_macros/check_functions.m4
@@ -3,6 +3,8 @@
 AC_CHECK_FUNCS(_gmtime64_s)
 AC_CHECK_FUNCS(_localtime64_s)
 AC_CHECK_FUNCS(_strtoui64)
+AC_CHECK_FUNCS(futimens)
+AC_CHECK_FUNCS(futimes)
 AC_CHECK_FUNCS(gmtime_r)
 AC_CHECK_FUNCS(localtime_r)
 AC_CHECK_FUNCS(mkstemp)
diff --git a/config.h.cmake b/config.h.cmake
index 5ee382357..e899f3444 100644
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -139,6 +139,8 @@
 #cmakedefine HAVE_CLOCK
 #cmakedefine HAVE_CLOCK_GETTIME
 #cmakedefine HAVE_FPCLASSIFY
+#cmakedefine HAVE_FUTIMENS
+#cmakedefine HAVE_FUTIMES
 #cmakedefine HAVE_GMTIME_R
 #cmakedefine HAVE_LOCALTIME_R
 #cmakedefine HAVE_MKSTEMP
```

I like that we use a macro for portability instead of embedding `ifdef` in code. For example:

```diff
diff --git a/lib/io.c b/lib/io.c
index b3171c2e9..5d5dbf270 100644
--- a/lib/io.c
+++ b/lib/io.c
@@ -49,6 +49,16 @@
 # define GRN_IO_FILE_SIZE_V0  GRN_IO_FILE_SIZE_V1
 #endif /* WIN32 */
 
+#ifdef HAVE_FUTIMENS
+# define grn_futimens(fd, times) futimens((fd), (times))
+# define grn_futimens_name "futimens"
+#elif defined(HAVE_FUTIMES) /* HAVE_FUTIMENS */
+# define grn_futimens(fd, times) futimes((fd), (times))
+# define grn_futimens_name "futimes"
+#else /* HAVE_FUTIMENS */
+# error ERROR_MESSAGE
+#endif /* HAVE_FUTIMENS */
+
 typedef struct _grn_io_fileinfo {
 #ifdef WIN32
   HANDLE fh;
@@ -2141,9 +2151,9 @@ grn_msync(grn_ctx *ctx, fileinfo *fi, void *start, size_t length)
     struct timeval times[2];
     gettimeofday(&(times[0]), NULL);
     times[1] = times[0];
-    r = futimes(fi->fd, times);
+    r = grn_futimens(fi->fd, times);
     if (r == -1) {
-      SERR("futimes failed: <%d>", fi->fd);
+      SERR("%s failed: <%d>", grn_futimens_name, fi->fd);
     }
   }
 
```

siepkes:
@kou Thanks for the thorough feedback! I've incorporated it in the PR. I've verified this also works for SmartOS (and should also work on Solaris). However the automated Windows build breaks. I'm assuming this is because the Windows build doesn't run `./autogen.sh`?

kou:
Ah, we need to add `ifndef WIN32` check because we don't need `futimens()` family on Windows. For example:

```c
#ifndef WIN32
# ifdef HAVE_FUTIMENS
...
# endif /* HAVE_FUTIMENS */
#endif /* WIN32 */
```

We use `SystemTimeToFileTime()` on Windows not `futimens()` family.

kou:
Thanks.
I've merged.

siepkes:
Thanks for all the assistance!

PR #869, opened 2018-09-10
scorer test: add a test to multiply the score

komainu8:


test/command/suite/select/scorer/expression/multiplication.test
```diff
@@ -0,0 +1,15 @@
+table_create Users TABLE_HASH_KEY ShortText
+column_create Users tags COLUMN_VECTOR|WITH_WEIGHT ShortText
+
+table_create Tags TABLE_HASH_KEY ShortText
+column_create Tags users_tags COLUMN_INDEX|WITH_WEIGHT Users tags
```
kou: Do we need index?

test/command/suite/select/scorer/expression/multiplication.test
```diff
@@ -0,0 +1,15 @@
+table_create Users TABLE_HASH_KEY ShortText
+column_create Users tags COLUMN_VECTOR|WITH_WEIGHT ShortText
```
kou: Do we need `WITH_WEIGHT`?

komainu8:
I've fixed them.

kou:
We should fix this ASAP.

PR #870, opened 2018-09-11
scorer test: add a test for a score of reference to a vector column

komainu8:


kou:
I've fixed this case.

PR #885, opened 2018-12-13
logical_select: add a test fot load_table option

komainu8:


test/command/suite/sharding/logical_select/load_table/load_table.test
```diff
@@ -0,0 +1,26 @@
...
+}
+]
+load --table Logs_20150204
+[
+{
+  "timestamp": "2015-02-04 00:00:00",
+  "memo":      "2015-02-04 00:00:00"
+}
+]
+logical_select --logical_table Logs --shard_key timestamp --load_table Temp --load_columns "memo, timestamp" --load_values "timestamp, memo"
```
kou: It should be broken to multiple lines.

kou: It should use different column names.

komainu8:
@kou 
Thank you for your comments.
I've fixed them.

komainu8:
This pull request close. Because It has already been implemented.

PR #893, opened 2019-01-17
[WIP] logger: add log-flags option

komainu8:
TODO: test

### DIFF
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -118,6 +118,40 @@ grn_log_level_parse(const char *string, grn_log_level *level)
   }
 }
 
+int
+grn_log_flags_parse(const char *string)
+{
+  const char *string_end = string + strlen(string);
+  int flags = GRN_LOG_TIME|GRN_LOG_MESSAGE;
+
+  if (!string) {
+    return flags;
+  }
+
+  while (string < string_end) {
+    if (*string == '|' || *string == ' ') {
+      string += 1;
+      continue;
+    }
+
+    if (strcmp(string, "+pid")) {
+      flags |= GRN_LOG_PID;
+      string += strlen(string);
+      continue;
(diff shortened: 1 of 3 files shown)


lib/logger.c
```diff
@@ -118,6 +118,40 @@ grn_log_level_parse(const char *string, grn_log_level *level)
   }
 }
 
+int
+grn_log_flags_parse(const char *string)
+{
+  const char *string_end = string + strlen(string);
```
kou: We should not touch `NULL`-able value before `NULL` check.

lib/logger.c
```diff
@@ -118,6 +118,40 @@ grn_log_level_parse(const char *string, grn_log_level *level)
...
+    return flags;
+  }
+
+  while (string < string_end) {
+    if (*string == '|' || *string == ' ') {
+      string += 1;
+      continue;
+    }
+
+    if (strcmp(string, "+pid")) {
```
kou: `strcmp()` returns `0` when the arguments have the same value.

lib/logger.c
```diff
@@ -118,6 +118,40 @@ grn_log_level_parse(const char *string, grn_log_level *level)
...
+      flags |= GRN_LOG_PID;
+      string += strlen(string);
```
kou: It always goes to the end of `string`.

src/groonga.c
```diff
@@ -3522,6 +3525,13 @@ main(int argc, char **argv)
     grn_default_logger_set_rotate_threshold_size(value);
   }
 
+  if (log_flags_arg) {
+    int log_flags = grn_log_flags_parse(log_flags_arg);
+    if (!log_flags) {
```
kou: `0` is a valid flags value.

PR #897, opened 2019-02-04
appveyor: execute command line test

kenhys:
TODO: It should work on appveyor worker image even though non-ascii directory name is used

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -109,6 +109,8 @@ test_script:
      --timeout 60
      --n-retries 2
      test\command\suite
+  - ruby test\command_line\run-test.rb
+     --groonga-install-prefix=%FULL_GROONGA_INSTALL_FOLDER%
 
 on_success:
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -52,7 +52,8 @@ install:
   # - choco install -y imdisk-toolkit
   # - mkdir tmp
   # - imdisk -a -t file -m tmp -o awe -s 1G -p "/fs:ntfs /q /y"
-  - set GROONGA_INSTALL_SUB_FOLDER=インストール
+  # - set GROONGA_INSTALL_SUB_FOLDER=インストール
```
komainu8: Is this comment need?
I think that no problem to remove this comment.

kenhys: It is better to test japanese directory name, but not succeeded yet.(because of AppVeyor specific reason, maybe)
So, for the record, I intentionally keep it left as a comment.

PR #904, opened 2019-02-10
Modify for suppressing a warning that C4018

komainu8:
A type of variable "i" is int and a type of variable "n_floating_threads" is uint32_t.
So, if the size of the value of "n_floating_threads" over the size of "int", "i" occurs overflow.


kou:
Thanks.

PR #913, opened 2019-03-20
normalizer: use GRN_CHAR_TYPE

komainu8:
Because char_type is sometimes OR operated with GRN_CHAR_BLANK.

Therefore, when load data that there is whitespace after KATAKANA,
load data are not normalize.
Because KATAKANA before whitespace is operated OR with GRN_CHAR_BLANK,
it can not be decided as KATAKANA.

Fix issues #912

### DIFF
--- a/lib/normalizer.c
+++ b/lib/normalizer.c
@@ -1141,7 +1141,7 @@ grn_nfkc_normalize_unify_stateless(grn_ctx *ctx,
     }
 
     if (data->options->unify_kana &&
-        char_type == GRN_CHAR_KATAKANA &&
+        GRN_CHAR_TYPE(char_type) == GRN_CHAR_KATAKANA &&
         unified_char_length == 3) {
       unifying = grn_nfkc_normalize_unify_kana(unifying, unified_kana);
       if (unifying == unified_kana) {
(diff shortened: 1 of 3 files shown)


komainu8:

lib/normalizer.c
```diff
@@ -1141,7 +1141,7 @@ grn_nfkc_normalize_unify_stateless(grn_ctx *ctx,
     }
 
     if (data->options->unify_kana &&
-        char_type == GRN_CHAR_KATAKANA &&
+        GRN_CHAR_TYPE(char_type) == GRN_CHAR_KATAKANA &&
         unified_char_length == 3) {
       unifying = grn_nfkc_normalize_unify_kana(unifying, unified_kana);
       if (unifying == unified_kana) {
```
kou: Should we need to change the following line?

```c
char_type = GRN_CHAR_HIRAGANA;
```

```c
char_type = GRN_CHAR_HIRAGANA | (char_type & GRN_CHAR_BLANK));
````

kou: There are more `char_type = ...` in this function.

komainu8:

kou:
I've merged but we should also have a test for `test/command/suite/normalizers/nfkc100/`.

PR #914, opened 2019-03-21
[WIP] ii: add check for destination chunk area

komainu8:
Because there is a problem that crash when a destination chunk area isn't enough.
If the destination chunk area isn't enough, we expand its area and output log.

TODO: Test


komainu8:
This feature already implemented by 3d1071188b32fe0bebb3f484da5c5482e04b4b2b.
Therefore, this PR close.

PR #917, opened 2019-04-02
windows: install missing pdb files

kenhys:
string and vector pdb files are exists, but time, number, index_column, math pdb files are missing


kenhys:

PR #918, opened 2019-04-02
appveyor: install groonga-admin

kenhys:
Because MinGW bundles groonga-admin.


kenhys:
Thanks!

PR #920, opened 2019-04-02
appveyor: install license files under share\license

kenhys:


kenhys:
@komainu8  thanks!

kenhys:
I've got a feedback that `share/license` is not common way. so withdraw this PR.

PR #921, opened 2019-04-03
appveyor: fix wrong argument to move

kenhys:
In the previous version, it causes the following error.

   The filename, directory name, or volume label syntax is incorrect.


kenhys:

Now It's ready for creating an artifact again.
https://ci.appveyor.com/project/kenhys/groonga/builds/23556931/job/slrgb3lpca006v6b/artifacts

PR #922, opened 2019-04-03
cmake message pack: install missing message pack header files

kenhys:


kenhys:
It works as expected.
https://ci.appveyor.com/project/kenhys/groonga/builds/23560885/job/dbj99s1pa2uctl7s

PR #923, opened 2019-04-03
cmake: install missing *.hpp

kenhys:
groonga.hpp and arrow.hpp are missing.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -14,9 +14,13 @@
 # License along with this library; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
-install(FILES groonga.h DESTINATION "${GRN_INCLUDE_DIR}")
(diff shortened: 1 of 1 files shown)


include/CMakeLists.txt
```diff
@@ -16,7 +16,11 @@
 
 install(FILES groonga.h DESTINATION "${GRN_INCLUDE_DIR}")
```
komainu8: Do we needless install "groonga.hpp"?

kenhys: Oh, sure.

kenhys:
@komainu8 Thanks!

I've fixed it.

Also, I've found the problem that groonga.h is installed to a wrong path, so I've also fixed it.

kenhys:
Hmm, not completed yet... it need more fixes.

kenhys:
https://github.com/groonga/groonga/commit/61de0f9ded701eff88a31d1a25d29423fc603d2a is not needed.
@komainu8 now, ready for review again.

kenhys:
Thanks!

PR #924, opened 2019-04-04
cmake windows: enable USE_EXACT_ALLOC_COUNT

kenhys:
WIP: Without USE_EXACT_ALLOC_COUNT, grn_fin(N) reports unexpected leaks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,6 +334,7 @@ if(WIN32)
 
   SET(PSAPI_LIBS "psapi.lib")
 
+  set(USE_EXACT_ALLOC_COUNT)
   set(USE_SELECT TRUE)
 else()
   ac_check_headers(sys/epoll.h)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,7 +334,7 @@ if(WIN32)
 
   SET(PSAPI_LIBS "psapi.lib")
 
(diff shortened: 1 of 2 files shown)


kenhys:
with intrinsic bultin function, command test will succeeds

kenhys:

CMakeLists.txt
```diff
@@ -334,6 +334,7 @@ if(WIN32)
 
   SET(PSAPI_LIBS "psapi.lib")
 
+  set(USE_EXACT_ALLOC_COUNT)
```
kou: Could you use `set(USE_EXACT_ALLOC_COUNT TRUE)` like `USE_SELECT`?

lib/grn.h
```diff
@@ -135,6 +135,7 @@ typedef SOCKET grn_sock;
 
 # ifndef __GNUC__
 #  include <intrin.h>
+#  pragma intrinsic (_InterlockedExchangeAdd)
```
kou: Could you remove a space between `intrinsic` and `(`?

CMakeLists.txt
kenhys: I've fixed.

lib/grn.h
kenhys: I've fixed.

kenhys:
@kou I've fixed based on your feedback!

PR #925, opened 2019-04-04
appveyor: enable http command test

kenhys:


appveyor.yml
```diff
@@ -120,6 +120,15 @@ test_script:
      --timeout 60
      --n-retries 2
      test\command\suite
+  - ruby test\command\bin\grntest
+     --groonga %FULL_GROONGA_INSTALL_FOLDER%\bin\groonga.exe
+     --base-directory test\command
+     --reporter mark
+     --n-workers 1
```
komainu8: In my opinion, `--n-workers` are multiple better.
Because we are easy to find problems.

kenhys: It is reasonable to change.

kenhys:
@komainu8  I've rebased and fixed to use multiple worker.

komainu8:
@kenhys 
The following test fails.
https://ci.appveyor.com/project/groonga/groonga/builds/23730567/job/ey22198bc9qwydew#L8235

kenhys:
@komainu8 
Hmm, https://ci.appveyor.com/project/groonga/groonga/builds/23730567
some works, but not all.
So, It seems that multiple worker is not appropriate...

kenhys:
I've rebased and reverted multiple worker commit.

kenhys:
https://ci.appveyor.com/project/groonga/groonga/builds/23876216 indicates that it can't be enabled http test because of execution time limit. (1h over) :cry:

PR #931, opened 2019-04-09
snippet, snippet_html: add an option for changing return value for no match

komainu8:
I added this feature for backward compatibility.

We can change the return value for no match by the second argument of snippet_html.
The default return value is "null".

For example, if we specify "[]" in the second argument,  return values is "[]" for no match.
If we specify none in the second argument, return value is "null" for no match.


komainu8:

kenhys:
@komainu8 I'll review it, just a moment, please.

komainu8:

kou:
This is too ad-hoc. For example, we can't use `"[]"` string (not empty vector) as the default value.
We should implement real solution.

komainu8:
I've fixed to use an "option object".

lib/grn_ecmascript.lemon
```diff
@@ -417,9 +417,18 @@ primary_expression ::= STRING.
...
 
+
 array_literal ::= BRACKETL elision BRACKETR.
 array_literal ::= BRACKETL element_list elision BRACKETR.
 array_literal ::= BRACKETL element_list BRACKETR.
+array_literal ::= BRACKETL BRACKETR. {
+  grn_ctx *ctx = efsi->ctx;
+
+  GRN_PTR_INIT(&efsi->array_literal, GRN_OBJ_VECTOR, GRN_ID_NIL);
+  grn_expr_take_obj(ctx, efsi->e, &efsi->array_literal);
```
kou: We should not pass static object to `grn_expr_take_obj()`. We should pass heap allocated object created by `grn_obj_open()`.

lib/grn_ecmascript.lemon
```diff
@@ -417,9 +417,18 @@ primary_expression ::= STRING.
 primary_expression ::= BOOLEAN.
 primary_expression ::= NULL.
 
+
```
kou: Could you revert needless change?

lib/grn_ecmascript.lemon
```diff
@@ -417,9 +417,18 @@ primary_expression ::= STRING.
...
 array_literal ::= BRACKETL elision BRACKETR.
 array_literal ::= BRACKETL element_list elision BRACKETR.
 array_literal ::= BRACKETL element_list BRACKETR.
+array_literal ::= BRACKETL BRACKETR. {
```
kou: Can we use existing `BRACKETL elision BRACKETR` for this instead of creating a new rule?

Ah, existing `array_literal` definition isn't good.

It should be:

```text
array_literal ::= BRACKETL element_list BRACKETR.

element_list ::= .
element_list ::= assignment_expression.
element_list ::= element_list COMMA assignment_expression.
```

lib/proc/proc_snippet.c
```diff
@@ -47,15 +48,19 @@ snippet_exec(grn_ctx *ctx, grn_obj *snip, grn_obj *text,
     return NULL;
   }
 
+  if (n_results == 0) {
+    if (default_return_value == NULL) {
```
kou: Please don't use `== NULL` or `!= NULL`.

lib/proc/proc_snippet.c
```diff
@@ -251,19 +261,55 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
   grn_obj *snippets = NULL;
 
   /* TODO: support parameters */
```
kou: Can we remove this comment?

lib/proc/proc_snippet.c
```diff
@@ -251,19 +261,55 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+    grn_obj *default_return_value = NULL;
     int flags = GRN_SNIP_SKIP_LEADING_SPACES;
     unsigned int width = 200;
     unsigned int max_n_results = 3;
     const char *open_tag = "<span class=\"keyword\">";
     const char *close_tag = "</span>";
     grn_snip_mapping *mapping = GRN_SNIP_MAPPING_HTML_ESCAPE;
 
+    int i;
+    for (i = 1; i < nargs; i++) {
```
kou: Why should we need to loop here?

lib/proc/proc_snippet.c
```diff
@@ -251,19 +261,55 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+      if (args[i]->header.type == GRN_TABLE_HASH_KEY) {
+        grn_obj *options = args[i];
+        grn_hash_cursor *cursor;
+        void *key;
+        int key_size;
+        grn_obj *value;
+  
+        cursor = grn_hash_cursor_open(ctx, (grn_hash *)options,
```
kou: Could you use `GRN_HASH_EACH_BEGIN` and `GRN_HASH_EACH_END` macros?

lib/proc/proc_snippet.c
```diff
@@ -251,19 +261,55 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+        if (!cursor) {
+          GRN_PLUGIN_ERROR(ctx, GRN_NO_MEMORY_AVAILABLE,
+                           "snippet_html(): couldn't open cursor");
+          goto exit;
+        }
+        while (grn_hash_cursor_next(ctx, cursor) != GRN_ID_NIL) {
+          grn_hash_cursor_get_key_value(ctx, cursor,
+                                        &key, &key_size,
+                                        (void **)&value);
+          if (key_size == strlen("default") && !memcmp(key, "default", strlen("default"))) {
```
kou: Could you use `grn_raw_string`?

lib/proc/proc_snippet.c
```diff
@@ -251,19 +261,55 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+            default_return_value = value;
+          } else {
+            GRN_PLUGIN_ERROR(ctx, GRN_INVALID_ARGUMENT,
+                             "invalid option name: <%.*s>",
```
kou: Could you add a `snippet_html(): ` tag in log message?

lib/proc/proc_snippet.c
komainu8: Ah...This loop is needless. Therefore I've removed it.

komainu8:
I've fixed them.

komainu8:
Ah...Sorry.
This PR is error now in Travis CI.

I request a review again after I'll fix it.

lib/proc/proc_snippet.c
```diff
@@ -250,20 +260,47 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+                                      (void **)&value);
+        if (key_size == option_name.length
+            && !memcmp(key, option_name.value, option_name.length)) {
+          default_return_value = value;
+        } else {
+          GRN_PLUGIN_ERROR(ctx, GRN_INVALID_ARGUMENT,
+                           "snippet_html(): invalid option name: <%.*s>",
+                           key_size, (char *)key);
+          grn_hash_cursor_close(ctx, cursor);
+          goto exit;
```
kou: We should not use `goto` in `GRN_HASH_EACH_BEGIN {}`.
We should `break` and `goto`.

lib/grn_ecmascript.lemon
```diff
@@ -417,16 +417,20 @@ primary_expression ::= STRING.
 primary_expression ::= BOOLEAN.
 primary_expression ::= NULL.
 
-array_literal ::= BRACKETL elision BRACKETR.
-array_literal ::= BRACKETL element_list elision BRACKETR.
+array_literal ::= BRACKETL elision BRACKETR. {
```
kou: Is `elision` needed?

lib/proc/proc_snippet.c
```diff
@@ -250,20 +260,47 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+  
+      GRN_HASH_EACH_BEGIN(ctx, (grn_hash *)options, cursor, id) {
+        grn_raw_string option_name;
+        option_name.value = "default";
+        option_name.length = strlen("default");
+
+        grn_hash_cursor_get_key_value(ctx, cursor,
+                                      &key, &key_size,
+                                      (void **)&value);
+        if (key_size == option_name.length
```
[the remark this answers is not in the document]
komainu8: I tried to execute your suggestion codes. However, this code occurred error.

```
"snippet_html(): invalid option name: <default>"
```

Cause of this error is a value of `option_name.length` is strange as below.

```
(gdb) p option_name
$12 = {
  value = 0x7fbfb48bb9e0 "default",
  length = 12884901895
}
```

It seems to be we can't get the correct value of `option_name.length` if cast to `int` as below.
```
(gdb) p option_name.length
$14 = 12884901895
(gdb) p (int)option_name.length
$15 = 7
```

Therefore, how about as below code?

```c
if (nargs > 1 && args[1]->header.type == GRN_TABLE_HASH_KEY) {
  grn_obj *options = args[1];
  void *key;
  int key_size;
  grn_obj *value;

  GRN_HASH_EACH_BEGIN(ctx, (grn_hash *)options, cursor, id) {
    grn_raw_string option_name;

    grn_hash_cursor_get_key_value(ctx, cursor,
                                  &key, &key_size,
                                  (void **)&value);
    option_name.value = key;
    option_name.length = key_size;
    if (GRN_RAW_STRING_EQUAL_CSTRING(option_name, "default")) {
```

lib/grn_ecmascript.lemon
komainu8: Yes. I think elision is needed.

If we execute as below code, occurs error in tests(`test/command/run-test.sh test/command/suite/select/filter/vector/not_equal.test` ans `equal.test`).

```
array_literal ::= BRACKETL element_list BRACKETR.

element_list ::= .
element_list ::= assignment_expression.
element_list ::= element_list COMMA assignment_expression.
```

* Error of tests

```
=======================================================================================================
[3] .
  not_equal                                                                            0.1688s [failed]
=======================================================================================================
--- (expected)
+++ (actual)
@@ -1,46 +1,13 @@
  table_create Memos TABLE_HASH_KEY ShortText
  [[0,0.0,0.0],true]
  column_create Memos tags COLUMN_VECTOR ShortText
  [[0,0.0,0.0],true]
  load --table Memos
  [
  {"_key": "groonga is very fast", "tags": ["groonga"]}
  ]
  [[0,0.0,0.0],1]
  select Memos --filter 'tags != ["rroonga"]'
- [
-   [
-     0,
-     0.0,
-     0.0
-   ],
-   [
-     [
-       [
-         1
-       ],
-       [
-         [
-           "_id",
-           "UInt32"
-         ],
-         [
-           "_key",
?                 "Sh                or      tT   e   xt"  
? [[[-22,0.0,0.0], can't find class f   objec  typ : 0 6 ]]
-         ],
-         [
-           "tags",
-           "ShortText"
-         ]
-       ],
-       [
-         1,
?         "g       roo    nga  is very fas           t",        
? #|e| Ar   umentEr  r: ca 't f nd cl     s for objec  type: 0x6
-         [
-           "groonga"
-         ]
-       ]
-     ]
-   ]
- ]
  
=======================================================================================================
[0] .
  equal                                                                                0.1766s [failed]
=======================================================================================================
--- (expected)
+++ (actual)
@@ -1,39 +1,13 @@
  table_create Memos TABLE_HASH_KEY ShortText
  [[0,0.0,0.0],true]
  column_create Memos tags COLUMN_VECTOR ShortText
  [[0,0.0,0.0],true]
  load --table Memos
  [
  {"_key": "groonga is very fast", "tags": ["groonga"]}
  ]
  [[0,0.0,0.0],1]
  select Memos --filter 'tags == ["groonga"]'
- [
-   [
-     0,
-     0.0,
-     0.0
-   ],
-   [
-     [
-       [
-         0
-       ],
-       [
-         [
-           "_id",
-           "UInt32"
-         ],
-         [
-           "_key",
?                 "Sh                or      tT   e   xt"  
? [[[-22,0.0,0.0], can't find class f   objec  typ : 0 6 ]]
-         ],
-         [
-           "tags",
?           "Sh   or      tT                  e          xt"
? #|e| ArgumentErr  : can'  find class for obj ct type: 0 6 
-         ]
-       ]
-     ]
-   ]
- ]
```

Therefore, I make the definition of `elision`.

lib/proc/proc_snippet.c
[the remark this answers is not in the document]
kou: I see.

FYI: We should use `uint32_t key_size`.

lib/grn_ecmascript.lemon
kou: The error is caused by generating invalid expression.
It generates the following expression:

  * `get_value: tags`
  * `push: "groonga"`
  * `push: []`
  * `equal`

It should be the following:

  * `get_value: tags`
  * `push: ["groonga"]`
  * `equal`

I pushed array literal support to master. Could you rebase on master?

lib/proc/proc_snippet.c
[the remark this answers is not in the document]
komainu8: Thank you for your information.
I'll use `uint32_t key_size`.

lib/grn_ecmascript.lemon
komainu8: Thank you.
I'll rebase on master.

komainu8:
I've fixed them.

kou:
Could you rebase on master?

komainu8:
I've finished rebasing on master.

lib/proc/proc_snippet.c
```diff
@@ -250,20 +260,49 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+                                      (void **)&value);
+        option_name.value = key;
+        option_name.length = key_size;
+        if (GRN_RAW_STRING_EQUAL_CSTRING(option_name, "default")) {
+          default_return_value = value;
+        } else {
+          GRN_PLUGIN_ERROR(ctx, GRN_INVALID_ARGUMENT,
+                           "snippet_html(): invalid option name: <%.*s>",
+                           key_size, (char *)key);
+          grn_hash_cursor_close(ctx, cursor);
```
kou: This is needless.

lib/proc/proc_snippet.c
```diff
@@ -250,20 +260,49 @@ func_snippet_html(grn_ctx *ctx, int nargs, grn_obj **args,
...
+          break;
+        }
+      } GRN_HASH_EACH_END(ctx, cursor);
+      if (!default_return_value) {
```
kou: What is expected behavior of `snippet_html("...", {})`?

komainu8: We expect return null in this case.
Therefore, this `if` section(`if (!default_return_value) {` ) is needless.

I'll fix it.

komainu8:
I've fixed them.

lib/proc/proc_snippet.c
kou: Could you add a test for this case?

kou:
Could you remove the change for `lib/grn_ecmascript.c`? Is it needed?

komainu8:
I've added a test case for the error case and remove the needless change.

kou:
Could you also add empty option case? `snippet_html("...", {})`

komainu8:
I've added a test for empty option.
However, Groonga is crash when executing a test for invalid option case.
I'll fix it.

test/command/suite/select/function/snippet_html/none_with_invalid_option.test
```diff
@@ -0,0 +1,15 @@
...
+
+load --table Documents
+[
+["_key", "content"],
+["Groonga", "Groonga can be used with MySQL."]
+]
+
+select Documents \
+  --match_columns content --query 'MySQL' \
+  --output_columns '_key, snippet_html(_key, {"...": {}})'
```
kou: "nonexistent" or something is better than "...".

kou:
I've fixed remained problems.

PR #943, opened 2019-04-19
test logical_count: add test for window function over shards

komainu8:


komainu8:

PR #944, opened 2019-04-19
Support grndb --log-flags

kenhys:


kenhys:
If https://github.com/kou/mruby-slop/pull/1 is merged, this PR works as expected.

test/command_line/suite/grndb/test_options.rb
```diff
@@ -0,0 +1,64 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+class TestGrnDBOptions < GroongaTestCase
+  sub_test_case("--log-flags") do
+
```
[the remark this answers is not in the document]
kenhys: Thanks. I've added it too.

kenhys:
@komainu8 I've rebased and added default test case. could you review again please?

kenhys:
now, mruby-slop supports slop 4.6.2. https://github.com/kou/mruby-slop/commit/0aa5b832315ccd1ddc55f5391f2d8a1e9d6145b0

kenhys:
To use latest mruby-slop, need to update vendor/mruby-source, but I've found that there is some incompatible changes are introduced.

```
/usr/bin/ld: ./.libs/libgroonga.so: undefined reference to `MKOP_A'
/usr/bin/ld: ./.libs/libgroonga.so: undefined reference to `GET_OPCODE'
/usr/bin/ld: ./.libs/libgroonga.so: undefined reference to `MKOP_AB'
collect2: error: ld returned 1 exit status
```

This incompatibility is introduced at https://github.com/mruby/mruby/commit/891839b976c75c77f238931123ac472e3284e95d

The feature about keyword argument is supported after above commit, so need to support new bytecode implementation.

https://github.com/mruby/mruby/commits/stable?before=8ee516436b8d174a50764939bee23a442aa00b3f+661

kenhys:
And recent mruby-time split header file, so include path need to be updated build mruby.

```
diff --git a/vendor/mruby/Makefile.am b/vendor/mruby/Makefile.am
index c7c021ac4..e330469a2 100644
--- a/vendor/mruby/Makefile.am
+++ b/vendor/mruby/Makefile.am
@@ -11,7 +11,8 @@ DEFAULT_INCLUDES =                                            \
        -I$(srcdir)/../mruby-source/include                     \
        -I$(srcdir)/../mruby-source/src                         \
        -I$(srcdir)/../mruby-source/mrbgems/mruby-compiler/core \
-       -I$(srcdir)/../mruby-source/mrbgems/mruby-io/include
+       -I$(srcdir)/../mruby-source/mrbgems/mruby-io/include    \
+       -I$(srcdir)/../mruby-source/mrbgems/mruby-time/include
 
```

kenhys:
even though on only travis fails, now test passes on locally.

kenhys:
Could you review this PR, @kou

kou:
Could you rebase on master?
Most of CI failures will be fixed.

test/command_line/suite/grndb/test_options.rb
```diff
@@ -0,0 +1,72 @@
...
```
kou: Could you remove this empty line?

lib/mrb/scripts/logger/flags.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      attr_reader :name, :flag
+      def initialize(name, flag)
+        @@flags[flag] = self
+        @name  = name
+        @flag = flag
+      end
+
+      def to_s
+        @@flags_names
```
kou: Why `@name` isn't used?

lib/mrb/scripts/logger/flags.rb
```diff
@@ -0,0 +1,36 @@
...
+    class Flags
+      @@flags = {}
+      @@flag_names = nil
+      class << self
+        def set_flag_names(flag)
+          flag_names = @@flags.collect do |index, candidate|
+            candidate.name if candidate.flag & flag > 0
+          end
+          @@flag_names = flag_names.compact.join("|")
+        end
```
kou: `format` or something is better.

lib/mrb/scripts/logger/flags.rb
```diff
@@ -0,0 +1,36 @@
+module Groonga
+  class Logger
+    class Flags
+      @@flags = {}
+      @@flag_names = nil
```
kou: Is this needed?

lib/mrb/scripts/logger/flags.rb
```diff
@@ -0,0 +1,36 @@
+module Groonga
+  class Logger
+    class Flags
```
kou: Why `Flags` not `Flag`?

test/command_line/suite/grndb/test_options.rb
kenhys: I've fixed.

lib/mrb/scripts/logger/flags.rb
kenhys: I've fixed.

lib/mrb/scripts/logger/flags.rb
kenhys: No, It can be removed.

lib/mrb/scripts/logger/flags.rb
kenhys: Oh, for consistency, it seems that `Flag` is better.

kenhys: I've fixed.

lib/mrb/scripts/logger/flags.rb
kenhys: `@name` is implicitly used when correcting each flag name.

kenhys: (`@name` is not enough to collect all default flags, so I've used `@flags_names` in the previous version)

kenhys:
@kou Thanks, I've fixed based on your feedback.

src/grndb.c
```diff
@@ -183,6 +195,17 @@ main(int argc, char **argv)
     grn_default_logger_set_max_level(log_level);
   }
 
+  int log_flags = GRN_LOG_DEFAULT;
+  grn_default_logger_set_flags(log_flags);
```
kou: Why is this needed?

test/command_line/suite/grndb/test_options.rb
```diff
@@ -0,0 +1,71 @@
...
+      grndb("check")
+      assert_equal("1970-01-01 00:00:00.000000|n| " +
+                   "grn_init: <VERSION>",
+                   normalize_init_line(File.readlines(@log_path).first))
+    end
+
+    test("add: one") do
+      flags = "+pid"
+      groonga("status",
+              command_line: ["--log-flags", flags])
```
kou: Why is this option is needed for `groonga`?

lib/mrb/scripts/command_line_parser.rb
```diff
@@ -97,6 +97,9 @@ def setup_common_options(options)
       options.string("--log-level",
                      "Change log level (#{default_log_level.name})",
                      default: default_log_level)
+      default_log_flags = Logger.default_flags
+      options.string("--log-flags",
+                     "Change log flags (#{default_log_flags.format})")
```
kou: Why is`default:` not used?

lib/mrb/mrb_logger.c
```diff
@@ -57,6 +57,18 @@ logger_s_get_default_level(mrb_state *mrb, mrb_value self)
...
 
+static mrb_value
+logger_s_get_default_flags(mrb_state *mrb, mrb_value self)
+{
+  mrb_value mrb_flags_class;
+  mrb_value mrb_flags;
+
+  mrb_flags_class = mrb_const_get(mrb, self, mrb_intern_lit(mrb, "Flag"));
+  mrb_flags = mrb_fixnum_value(grn_default_logger_get_flags());
+  return mrb_funcall(mrb, mrb_flags_class, "set_flags", 1, mrb_flags);
```
kou: Calling setter in `default_XXX` isn't good API.

lib/mrb/mrb_logger.c
```diff
@@ -57,6 +57,18 @@ logger_s_get_default_level(mrb_state *mrb, mrb_value self)
...
+}
+
+
```
kou: Could you remove an empty needless line?

lib/mrb/scripts/logger/flag.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      NONE       = new(:none,       0)
+      TIME       = new(:time,       1)
+      TITLE      = new(:title,      2)
+      MESSAGE    = new(:message,    4)
+      LOCATION   = new(:location,   8)
+      PID        = new(:pid,        16)
+      PROCESS_ID = new(:process_id, 16)
+      THREAD_ID  = new(:thread_id,  32)
+    end
```
kou: Umm. This class isn't a good API.

How about the following?

```ruby
module Groonga
  class Logger
    class Flags
      VALUES = {
        time: 1,
        title: 2,
        message: 4,
        location: 8,
        process_id: 16,
        thread_id: 32,
      }

      attr_reader :value
      def initialize(value)
        case value
        when String, Symbol
          @value = VALUES[value.to_sym]
          if @value.nil?
            available_names = VALUES.keys.inspect
            message = "unknown flag name: #{value.inspect}: #{available_names}"
            raise ArgumentError, message
          end
        else
          @value = value
        end
      end

      def to_i
        @value
      end

      def to_s
        names = []
        VALUES.each do |name, value|
          names << name.to_s unless (value & @value) == 0
        end
        names.join("|")
      end

      def |(other)
        other = self.class.new(other) unless other.is_a?(self.class)
        new(@value | other.to_i)
      end

      def &(other)
        other = self.class.new(other) unless other.is_a?(self.class)
        new(@value & other.to_i)
      end

      NONE = new(0)
      TIME = new(:time)
      TITLE = new(:title)
      MESSAGE = new(:message)
      LOCATION = new(:location)
      PROCESS_ID = new(:process_id)
      PID = PROCESS_ID
      THREAD_ID = new(:thread_id)
    end
  end
end
```

```c
static mrb_value
logger_s_get_default_flags(mrb_state *mrb, mrb_value self)
{
  mrb_value mrb_flags_class =
    mrb_const_get(mrb, self, mrb_intern_lit(mrb, "Flag"));
  mrb_value mrb_flags = mrb_fixnum_value(grn_default_logger_get_flags());
  return mrb_funcall(mrb, mrb_flags_class, "new", 1, mrb_flags);
}
```

src/grndb.c
kenhys: Oh, It is needless. fixed.

test/command_line/suite/grndb/test_options.rb
kenhys: Ah, It is needless. And I've noticed that test case is wrong. I've fixed.

lib/mrb/mrb_logger.c
kenhys: fixed.

lib/mrb/scripts/logger/flag.rb
kenhys: It seems that it is reasonable API which I couldn't design.

kenhys:
@kou I've fixed based on your suggestion. could you review again?

lib/mrb/scripts/logger/flag.rb
```diff
@@ -0,0 +1,60 @@
+module Groonga
+  class Logger
+    class Flag
```
kou: In this API, this should be `Flags`. Because this object has multiple flag values. For example, `Flags.new(1 | 2 | 4)` is valid for `time`, `title` and `message` flags.

src/grndb.c
```diff
@@ -184,6 +196,16 @@ main(int argc, char **argv)
     grn_default_logger_set_max_level(log_level);
   }
 
+  int log_flags = GRN_LOG_DEFAULT;
```
kou: Could you reduce scope by putting this into `if (log_flags_name)` block?

kou: We don't need to initialize this by `GRN_LOG_DEFAULT`. Because the default value isn't used.

lib/mrb/scripts/logger/flag.rb
kenhys: I've got it.

src/grndb.c
kenhys: Fixed.

kenhys:
@kou I've fixed them!

PR #945, opened 2019-04-19
doc: add grndb --log-flags explanation

kenhys:


doc/source/reference/executables/grndb.rst
```diff
@@ -118,6 +118,18 @@ Here is an example that specifies ``--log-path`` option::
 
   % grndb check --log-path /var/log/groonga/grndb.log /var/lib/groonga/db/db
 
+``--log-flags``
+"""""""""""""""
+
+.. versionadded:: 9.0.2
+
+It specifies a flag of ``grndb`` log.
```
komainu8: In my opinion, we should write about we can output what.

komainu8: In my opinion, we should write about the default value.

kenhys: Thanks!
I've fixed.

kenhys: Thanks!
I've fixed.

kenhys:
@komainu8  I've fixed based on your review, could you review again, please?

PR #946, opened 2019-04-19
test logical_range_filter: add test for window function over shards

komainu8:


komainu8:

komainu8:
@kou 
I've modified these tests to the case that uses only two shards at most.
Could you review this PR again, please?

PR #965, opened 2019-05-22
doc: add indexes with weight description

kenhys:


doc/source/tutorial/match_columns.rst
```diff
@@ -290,6 +290,18 @@ On the other hand, the second one matches one article only because of ``Replies2
 Indexes with Weight
 -------------------
 
-TODO
+You can search blog entries which contains important keyword in ``Blog1`` table.
+For example, let's try to search blog entries which contains ``Groonga`` as important keyword.
+Generally speaking, an important keyword tend to be included in blog title, so if ``title`` contains ``Groonga``, its ``_score`` must be raised.
```
komainu8: In my opinion, we should clearly describe needing an index for using this feature.

doc/source/tutorial/match_columns.rst
```diff
@@ -290,6 +290,18 @@ On the other hand, the second one matches one article only because of ``Replies2
...
 
-.. TODO: match_columnsの重み指定機能についても触れる。
+Here is the example which search blog entries with ``Groonga`` as important keyword in ``title`` or ``message`` columns.
+
+.. groonga-command
+.. include:: ../example/tutorial/match_columns-indexes-with-weight.log
+.. select --table Blog1 --match_columns 'IndexBlog1.index_title * 10 || IndexBlog1.index_message * 1' --query 'Groonga' --output_columns "_id, _score, *"
```
komainu8: In my opinion, we should add an example of the execution of the setting of the index.

doc/source/tutorial/match_columns.rst
```diff
@@ -290,6 +290,18 @@ On the other hand, the second one matches one article only because of ``Replies2
...
+
+In above query, ``'IndexBlog1.index_title * 10 || IndexBlog1.index_message * 1'`` is specified for ``--match_columns``.
```
komainu8: Why do we set a default value of weight?
If we set it or not, I think the result of this query does not change.

doc/source/tutorial/match_columns.rst
kenhys: I got it.

doc/source/tutorial/match_columns.rst
kenhys: I've added reference.

doc/source/tutorial/match_columns.rst
kenhys: Sure, it may be redundant. I've removed it.

kenhys:
@komainu8 I've rebased and fixed based on your feedback.
Could you review it again, please?

kenhys:
@komainu8  Thanks!

PR #968, opened 2019-05-23
ctx: add support lua vm

komainu8:
However, currently, this feature is just only initialized the lua vm
and remove the lua vm when groonga finish.

### DIFF
--- a/lib/grn_ctx_impl.h
+++ b/lib/grn_ctx_impl.h
@@ -28,6 +28,12 @@
 # include <mruby.h>
 #endif
 
+#ifdef GRN_WITH_LUAJIT
+# include <lauxlib.h>
+# include <lualib.h>
+# include <lua.h>
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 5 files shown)


lib/grn_ctx_impl.h
```diff
@@ -137,6 +143,15 @@ struct _grn_mrb_data {
 #endif
 };
 
+typedef struct _grn_lua_data grn_lua_data;
+struct _grn_lua_data {
+  grn_bool initialized;
+#ifdef GRN_WITH_LUAJIT
+  lua_State *state;
+  char base_directory[PATH_MAX];
```
kou: Is this needed?

lib/grn_ctx_impl.h
```diff
@@ -28,6 +28,12 @@
 # include <mruby.h>
 #endif
 
+#ifdef GRN_WITH_LUAJIT
+# include <lauxlib.h>
+# include <lualib.h>
+# include <lua.h>
```
kou: Is this needed?

komainu8:
I've fixed them.

PR #970, opened 2019-05-27
windows: remove extended path prefix \\?\

kenhys:
There is a case that GetModuleFileNameW returns path which contains
extended path prefix \\?\ as module filename. It causes a bug that
plugin can't be found correctly because of this kind of extended path
prefix.

TODO: Support long path correctly on Windows. (Drop MAX_PATH limitation)

GitHub:#958

Reported by @yagisumi Thanks!!!

### DIFF
--- a/lib/grn_windows.h
+++ b/lib/grn_windows.h
@@ -19,6 +19,7 @@
 #pragma once
 
 #include "grn.h"
+#include "wchar.h"
 
 #ifdef __cplusplus
 extern "C" {
--- a/lib/plugin.c
+++ b/lib/plugin.c
@@ -440,7 +440,7 @@ grn_plugin_open(grn_ctx *ctx, const char *filename)
   } else {
     const char *label;
     label = grn_dl_open_error_label();
-    SERR("%s", label);
+    SERR("%s: <%.*s>", label, filename_size, filename);
   }
 
 exit:
--- a/lib/windows.c
+++ b/lib/windows.c
@@ -43,6 +43,12 @@ grn_windows_base_dir(void)
 
         absolute_dll_filename[absolute_dll_filename_size] = L'\0';
 
+        if (wcsncmp(absolute_dll_filename, L"\\\\?\\", 4) == 0) {
+          wmemmove(absolute_dll_filename, &absolute_dll_filename[4], absolute_dll_filename_size - 4);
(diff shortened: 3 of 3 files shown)


kenhys:
@komainu8 Could you review this PR?

lib/windows.c
```diff
@@ -43,6 +43,12 @@ grn_windows_base_dir(void)
 
         absolute_dll_filename[absolute_dll_filename_size] = L'\0';
 
+        if (!wcsncmp(absolute_dll_filename, L"\\\\?\\", 4)) {
```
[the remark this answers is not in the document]
kenhys: なるほど！と思いました。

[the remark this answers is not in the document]
kenhys: 修正しました。

kenhys:
@komainu8 I've fixed it!

kenhys:
@komainu8 Thanks!

lib/grn_windows.h
```diff
@@ -19,6 +19,7 @@
 #pragma once
 
 #include "grn.h"
+#include "wchar.h"
```
kou: Why is this included in `windows.c`?

kou: Can we move this to `windows.c`?

kenhys: because `wmemmove` requires it.
https://docs.microsoft.com/ja-jp/cpp/c-runtime-library/reference/memmove-wmemmove?view=vs-2019

kenhys: I'll fix it.

lib/plugin.c
```diff
@@ -440,7 +440,7 @@ grn_plugin_open(grn_ctx *ctx, const char *filename)
   } else {
     const char *label;
     label = grn_dl_open_error_label();
-    SERR("%s", label);
+    SERR("%s: <%.*s>", label, filename_size, filename);
```
kou: `(int)filename_size`.

This breaks Mroonga CI: https://travis-ci.org/mroonga/mroonga/jobs/538586519#L1680

kou: BTW, is this safe?
Does `filename`'s encoding always equal to `grn_ctx->encoding`?
For example, if this is local encoding, we need to convert encoding by `grn_encoding_convert_from_locale()`.

kenhys: Oh, I didn't noticed before that commit. :disappointed:

PR #972, opened 2019-05-27
doc: add TokenPattern description

kenhys:


doc/locale/ja/LC_MESSAGES/reference.po
```diff
@@ -28954,6 +28954,68 @@ msgstr ""
...
+msgstr ""
+"``TokenPattern`` の例です。``TokenPattern`` は指定した正規表現にマッチする"
+"トークンのみを抽出します。そのためホワイトリストみたいなことができます。"
+
+msgid ""
+"For example, let's hide a specific keyword from search results even though "
+"search query contains it."
+msgstr ""
+"例えば、検索するクエリーに含まれていても、特定のキーワードは検索結果から隠す"
+"というのをやってみましょう。"
```
komainu8: In this case, "TokenPattern" looks like that used as blacklists.

kenhys: I've fixed description.

kenhys:
@komainu8 could you review again?

doc/locale/ja/LC_MESSAGES/reference.po
komainu8: Ah...I'm sorry.

In before example, the description was as a whitelist but the example was as a blacklist.
Therefore, in my opinion, we should align the example and description.

kenhys:
@komainu8 Did you review based on old version?
I've rewritten that explanation, so it should be removed already.
(Added one more sample and explain hit/not hit test case)

PR #976, opened 2019-05-31
doc: add TokenTable document

kenhys:


doc/locale/ja/LC_MESSAGES/reference.po
```diff
@@ -29008,6 +29008,60 @@ msgstr ""
...
 
+msgid "``TokenTable``"
+msgstr ""
+
+msgid ""
+"``TokenTable`` is a tokenizer which treats only known keywords as a token. "
+"The known keywords must be registered as a key in the table for "
+"``TokenTable``."
+msgstr ""
+"``TokenTable`` は既知のキーワードのみをトークンとして扱うトークナイザーです。"
```
komainu8: In my opinion, to make it clear, an example use case is needed.
e.g) White lists.

kenhys: I've added description.

kenhys:
@komainu8 
I've added more description, does it reasonable?

PR #981, opened 2019-06-11
appveyor: bundle missing licence files

kenhys:


kou:
We should do this in `CMakeLists.txt`.

appveyor.yml
```diff
@@ -102,6 +102,16 @@ build_script:
...
   - cd ..
+  # Install groonga-admin license files
+  - mkdir %FULL_GROONGA_INSTALL_FOLDER%\share\groonga\groonga-admin
+  - for %%i in ( LICENSE README.md ) do ( copy %%i %FULL_GROONGA_INSTALL_FOLDER%\share\groonga\groonga-admin\ )
+  # Install msgpack license files
+  - cd %APPVEYOR_BUILD_FOLDER%\vendor\msgpack-*
+  - copy NOTICE %FULL_GROONGA_INSTALL_FOLDER%\share\groonga\msgpack\
+  # Install mruby license files
+  - cd %APPVEYOR_BUILD_FOLDER%\vendor\mruby-source
+  - for %%i in ( AUTHORS LICENSE README.md ) do ( copy %%i %FULL_GROONGA_INSTALL_FOLDER%\share\groonga\mruby\ )
```
kou: Does https://github.com/groonga/groonga/blob/master/vendor/mruby/CMakeLists.txt#L101-L103 not work?

kenhys: Thanks, it should be fixed in CMakeLists.txt

PR #984, opened 2019-06-11
appveyor: add UCRT runtime

kenhys:
Usually UCRT runtime is automatically installed, but there is a case that it doesn't installed yet.
For such a specific environment, we use local-deployment.

### DIFF
--- a/packages/windows/vcruntime/ucrt-readme.txt
+++ b/packages/windows/vcruntime/ucrt-readme.txt
@@ -1,0 +1,53 @@
+Subject to MICROSOFT WINDOWS SOFTWARE DEVELOPMENT KIT (SDK) FOR WINDOWS 10 [1], This package
+ contains Universal C Runtime files as redistributable code:
+
+[1] https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk
+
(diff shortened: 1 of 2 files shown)


kenhys:
zip package is not built yet. so just waiting.

kenhys:
It seems that ucrt bundle succeeds.

https://ci.appveyor.com/project/kenhys/groonga/builds/25187602/job/9m2mbohw8slh5t3f

kenhys:
I've found wrong license file bundle, so I'll create one more separated PR.

packages/windows/vcruntime/ucrt-readme.txt
```diff
@@ -0,0 +1,53 @@
+Subject to MICROSOFT WINDOWS SOFTWARE DEVELOPMENT KIT (SDK) FOR WINDOWS 10 [1], This package
+ contains Universal C Runtime files as redistributable code:
+
+[1] https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk
+
+See sdk_license.rtf in SDK section 2.a DISTRIBUTABLE CODE, "ii. Distribution Requirements" and
```
komainu8: Is this OK without a link to the license file(sdk_license.rtf)?

kenhys: I've searched license text online, but can't find it.

PR #989, opened 2019-06-14
test: fix typos

komainu8:
Currently, these tests are not checked.
Because these file names are not correct.

Was this done on purpose?


komainu8:

PR #993, opened 2019-06-18
grndb: add --target-timestamp option

kenhys:


lib/db.c
```diff
@@ -128,6 +129,20 @@ gen_pathname(const char *path, char *buffer, int fno)
   }
 }
 
+static uint32_t target_last_modified = 0;
+
+grn_inline static grn_bool
+grn_db_recover_skip_target(grn_ctx *ctx, grn_obj *obj, uint32_t last_modified)
+{
+  struct stat stat_buf;
+  stat(grn_obj_get_io(ctx, obj)->path, &stat_buf);
```
komainu8: Is needless that check return value of `grn_obj_get_io`?

lib/db.c
```diff
@@ -15230,6 +15245,10 @@ grn_db_recover_database_remove_orphan_inspect(grn_ctx *ctx, grn_obj *db)
 static void
 grn_db_recover_database(grn_ctx *ctx, grn_obj *db)
 {
+  if (grn_db_recover_skip_target(ctx, db, target_last_modified)) {
+    GRN_LOG(ctx, GRN_LOG_INFO, "[db][recover] skipped to recover: <%s>", grn_obj_get_io(ctx, db)->path);
```
komainu8: In my opinion, we should write recover target in a log as below.
 
`"[db][recover] skipped to recover database: <%s>"`

lib/db.c
```diff
@@ -128,6 +129,20 @@ gen_pathname(const char *path, char *buffer, int fno)
   }
 }
 
+static uint32_t target_last_modified = 0;
```
komainu8: This scope of a variable is too large.
I think a static variable is not threaded safe.

We should scope of a variable is local in function.

lib/db.c
kenhys: It should be.

lib/db.c
```diff
@@ -128,6 +129,20 @@ gen_pathname(const char *path, char *buffer, int fno)
...
+
+grn_inline static grn_bool
+grn_db_recover_skip_target(grn_ctx *ctx, grn_obj *obj, uint32_t last_modified)
+{
+  struct stat stat_buf;
+  stat(grn_obj_get_io(ctx, obj)->path, &stat_buf);
+  if (stat_buf.st_mtime < target_last_modified) {
```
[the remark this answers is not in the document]
kenhys: st_mtime is equivalent to time_t so it seems ok in practice.

lib/db.c
kenhys: fixed.

lib/db.c
kenhys: fixed.

kenhys:
@komainu8 could you review again?

komainu8:
Some tests are errors in TravisCI.

```
rb-grn-database.c: In function ‘rb_grn_database_recover’:
rb-grn-database.c:590:10: error: too few arguments to function ‘grn_db_recover’
     rc = grn_db_recover(context, database);
```

kenhys:
Already fixed. I'll close this PR.

PR #999, opened 2019-07-03
mrb: add null check for preventing a crash

komainu8:
`mrb_generate_code` returns NULL when fails allocate memory in this function.

Groonga is crash executing `MRB_PROC_SET_TARGET_CLASS` when `mrb_generate_code` is NULL.

Because `proc` is write-accessed into `MRB_PROC_SET_TARGET_CLASS` as below.
(`proc` is return value of `mrb_generate_code`)

```
// `p` is `proc`
(p)->e.target_class = (tc);
```

### DIFF
--- a/lib/mrb.c
+++ b/lib/mrb.c
@@ -230,6 +230,16 @@ grn_mrb_load(grn_ctx *ctx, const char *path)
       struct RProc *proc;
       int arena_index;
       proc = mrb_generate_code(mrb, parser);
+      if (!proc) {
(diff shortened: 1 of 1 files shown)


komainu8:
@kenhys  Could you review this PR, please.

kenhys:
I'll review.

komainu8:
@kenhys Thank you for your comment. I'll fix other parts later.

komainu8:

PR #1002, opened 2019-07-08
Install missing COPYING

kenhys:
Put COPYING under /usr/share/groonga/COPYING

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,7 +18,6 @@ SUBDIRS =					\
 	data					\
 	tools					\
 	doc
-#dist_data_DATA =
 EXTRA_DIST =					\
 	CMakeLists.txt				\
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -32,6 +31,7 @@ EXTRA_DIST =					\
 	gpg_uid					\
 	nginx_version				\
 	version-gen.sh
+dist_pkgdata_DATA = COPYING
```
kou: Is `dist_` needed?

kenhys: It seems that no need to `dist_` prefix.

kenhys: Thanks.

kenhys:
@kou Thanks,  I've removed needless prefix from it.

PR #1003, opened 2019-07-08
windows: fix build error on MinGW

kenhys:
Patch via
https://github.com/msys2/MINGW-packages/commit/9c76ae395dd490b259d270608b47d130c6245490#diff-b266d04f16a509dc5bd3a3d12ca9df53

PR #1005, opened 2019-07-10
appveyor: CI out-of-tree build with MINGW

kenhys:
enable checking GitHub: #1003 build error issue continuously.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,9 @@ clone_depth: 10
 environment:
   matrix:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      MINGW_CHOST: mingw-w64-x86_64
+      MINGW_PREFIX: /mingw64
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       VS_VERSION: 15
       ARCH: amd64
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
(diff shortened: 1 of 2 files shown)


kenhys:
It can't be reproducible on Appveyor. (without patch, build failure is expected but not.)

[groonga-no-patch-makepkg-mingw.log](https://github.com/groonga/groonga/files/3376855/groonga-no-patch-makepkg-mingw.log) expected failure.
[groonga-no-patch-on-appveyor.log](https://github.com/groonga/groonga/files/3376856/groonga-no-patch-on-appveyor.log)
[groonga-no-patch-on-local.log](https://github.com/groonga/groonga/files/3376857/groonga-no-patch-on-local.log)

appveyor.yml
```diff
@@ -179,3 +182,40 @@ on_finish:
...
+  build_script:
+    - bash -c "./autogen.sh"
+    - bash -c "./configure --host=x86_64-w64-mingw32 --target=x86_64-w64-mingw32 --build=x86_64-w64-mingw32 --enable-document --with-ruby=ruby --enable-mruby"
+    - bash -c "make dist"
+    - cd %APPVEYOR_BUILD_FOLDER%\..
+    - mkdir build
+    - cd build
+    - cp %APPVEYOR_BUILD_FOLDER%\appveyor\PKGBUILD .
+    - cp %APPVEYOR_BUILD_FOLDER%\groonga-%GROONGA_VERSION%.tar.gz .
+    - bash -c "GROONGA_VERSION=%GROONGA_VERSION% MINGW_INSTALLS=mingw64 makepkg-mingw -sLf --noconfirm --needed"
```
kou: Why do we need to use `makepkg-mingw`?

kenhys: Because build error issue is found on PKGBUILD environment. If `makepkg-mingw` doesn't affect it, there is no need to use `makepkg-mingw`.

kenhys: I've tried without makepkg-mingw. It needs to be fixed additional linkage error issue.
https://ci.appveyor.com/project/kenhys/groonga/builds/26000711/job/l79vkjnpkrtsmgrm

kenhys: It seems that using `makepkg-mingw` is better instead of directly using ./configure ... because it absorbs build related bothersome things.

kou:
Done by GitHub Actions.

PR #1008, opened 2019-07-19
windows: download appveyor release artifacts automatically

kenhys:
It download artifacts files and rename to distinct filename.

### DIFF
--- a/packages/windows/download_appveyor_build.rb
+++ b/packages/windows/download_appveyor_build.rb
@@ -5,10 +5,10 @@ require "open-uri"
 require "veyor"
 require "octokit"
 
-def download(url, local_file)
-  open(url) do |remote_zip|
-    File.open(local_file, "wb") do |local_zip|
-      local_zip.print(remote_zip.read)
+def download(url, local_path)
+  open(url) do |remote_file|
+    File.open(local_path, "wb") do |local_file|
+      local_file.print(remote_file.read)
     end
   end
 end
--- a/packages/windows/download_appveyor_build.rb
+++ b/packages/windows/download_appveyor_build.rb
@@ -1,0 +1,64 @@
+#!/usr/bin/env ruby
+
+require "pathname"
+require "open-uri"
+require "veyor"
+require "octokit"
+
+def download(url, local_file)
+  ssl_verify_mode = nil
+  open(url, :ssl_verify_mode => ssl_verify_mode) do |remote_zip|
+    File.open(local_file, "wb") do |local_zip|
+      local_zip.print(remote_zip.read)
+    end
+  end
+end
+
+client = Octokit::Client.new
+
+groonga_repository = "groonga/groonga"
+
+base_dir = Pathname.new(__FILE__).expand_path.dirname
+top_dir = base_dir.parent.parent
+base_version = (top_dir + "base_version").read.strip
+tag_name = "v#{base_version}"
+
+appveyor_url = "https://ci.appveyor.com/"
+appveyor_info = nil
+client.statuses(groonga_repository, tag_name).each do |status|
+  next unless status.target_url.start_with?(appveyor_url)
+  case status.state
+  when "pending"
+    # Ignore
+  else
+    match_data = /\/([^\/]+?)\/([^\/]+?)\/builds\/(\d+)\z/.match(status.target_url)
+    appveyor_info = {
+      account: match_data[1],
+      project: match_data[2],
+      build_id: match_data[3],
+    }
+    break
+  end
+end
+if appveyor_info.nil?
+  raise "No AppVeyor build"
+end
+
+start_build = appveyor_info[:build_id].to_i + 1
+build_history = Veyor.project_history(account: appveyor_info[:account],
+                                      project: appveyor_info[:project],
+                                      start_build: start_build,
+                                      limit: 1)
+build_version = build_history["builds"][0]["buildNumber"]
+project = Veyor.project(account: appveyor_info[:account],
+                        project: appveyor_info[:project],
+                        version: build_version)
+project["build"]["jobs"].each do |job|
+  job_id = job["jobId"]
+  artifacts = Veyor.build_artifacts(job_id: job_id)
+  artifacts.each do |artifact|
+    file_name = artifact["fileName"]
+    url = "#{appveyor_url}api/buildjobs/#{job_id}/artifacts/#{file_name}"
+    download(url, "files/#{file_name}")
+  end
+end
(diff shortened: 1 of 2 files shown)


kenhys:
@kou Thanks, I've fixed to use `veyor` gem. (it can fetch job_id correctly)

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,81 @@
...
+    suffix << "-vs2013"
+  end
+  artifacts.each do |artifact|
+    file_name = artifact["fileName"]
+    url = "#{appveyor_url}api/buildjobs/#{job_id}/artifacts/#{file_name}"
+    file_name  = artifact["fileName"].sub(/#{arch}/, suffix)
+    system("curl", "--location", "--output", "files/#{file_name}", url)
+    options = {
+      :content_type => "application/zip",
+    }
```
kou: Where is this used?

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,81 @@
...
+base_version = (top_dir + "base_version").read.strip
+tag_name = "v#{base_version}"
+
+appveyor_url = "https://ci.appveyor.com/"
+appveyor_info = nil
+client.statuses(groonga_repository, tag_name).each do |status|
+  next unless status.target_url.start_with?(appveyor_url)
+  case status.state
+  when "pending"
+  # Ignore
```
kou: Need to indent.

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,81 @@
...
+    suffix << "-vs2017"
+  when /VS_VERSION=14/
+    suffix << "-vs2015"
+  when /VS_VERSION=12/
+    suffix << "-vs2013"
+  end
+  artifacts.each do |artifact|
+    file_name = artifact["fileName"]
+    url = "#{appveyor_url}api/buildjobs/#{job_id}/artifacts/#{file_name}"
+    file_name  = artifact["fileName"].sub(/#{arch}/, suffix)
```
kou: `sub(/#{Regexp.escape(arch)}/, ...)` or `sub(arch, ...)`

kou: `local_file_name = file_name.sub(...)`

kou: How about using `groonga-${VERSION}-${ARCHITECTURE}-${VISUAL_STUDIO_VERSION}.zip` for file name generated by CMake?

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,81 @@
...
+    system("curl", "--location", "--output", "files/#{file_name}", url)
```
kou: Using `open-uri` is better to reduce `curl` dependency.
We don't need workaround for MinGW because we run this only on GNU/Linux.

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,81 @@
...
+require "veyor"
+require "octokit"
+
+def env_value(name)
+  value = ENV[name]
+  raise "Specify #{name} environment variable" if value.nil?
+  value
+end
+
+client = Octokit::Client.new(:access_token => env_value("GITHUB_TOKEN"))
```
kou: GitHub token is needless because `groonga/groonga` is a public repository.

packages/windows/download_appveyor_build.rb
kenhys: It should be removed.

packages/windows/download_appveyor_build.rb
kenhys: It seems that good idea to contain VISUAL_STUDIO_VERSIOn in archive name. I'll change it.

packages/windows/download_appveyor_build.rb
kenhys: Oh, surely no need to use GITHUB_TOKEN. I'll fix.

kenhys: removed.

packages/windows/download_appveyor_build.rb
kenhys: fixed.

packages/windows/download_appveyor_build.rb
kenhys: fixed to use "fileName" (download zip as is)

packages/windows/download_appveyor_build.rb
kenhys: fixed.

packages/windows/download_appveyor_build.rb
kenhys: removed.

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env ruby
+
+require "pathname"
+require "open-uri"
+require "veyor"
+require "octokit"
+
+def download(url, local_file)
+  ssl_verify_mode = nil
```
kou: Why is this needed?

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,64 @@
...
+  open(url, :ssl_verify_mode => ssl_verify_mode) do |remote_zip|
```
kou: `remote_file`.

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env ruby
+
+require "pathname"
+require "open-uri"
+require "veyor"
+require "octokit"
+
+def download(url, local_file)
```
kou: `local_path`

packages/windows/download_appveyor_build.rb
```diff
@@ -0,0 +1,64 @@
...
+  ssl_verify_mode = nil
+  open(url, :ssl_verify_mode => ssl_verify_mode) do |remote_zip|
+    File.open(local_file, "wb") do |local_zip|
```
kou: `local_zip` -> `local_file`

packages/windows/download_appveyor_build.rb
kenhys: Thanks, there is no need to use it. removed.

packages/windows/download_appveyor_build.rb
kenhys: fixed.

packages/windows/download_appveyor_build.rb
kenhys: fixed.

packages/windows/download_appveyor_build.rb
kenhys: fixed.

Issue #43, opened 2012-12-07
Make enums UPPER_CASE


kou:

kou:
Keep compatibility.

komainu8:
We close this issue.
Because we modified to upper case for all the enum value name.

PR #1022, opened 2019-08-02
Make enum UPPSER_CASE for "mode_xxx"

komainu8:
A part of a fix for issue #43.

### DIFF
--- a/src/groonga_benchmark.c
+++ b/src/groonga_benchmark.c
@@ -2757,9 +2757,9 @@ usage(void)
 }
 
 enum {
-  mode_default = 0,
-  mode_list,
-  mode_usage,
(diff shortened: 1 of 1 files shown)


komainu8:

src/groonga_benchmark.c
```diff
@@ -2756,12 +2760,6 @@ usage(void)
   exit(1);
 }
 
-enum {
-  mode_default = 0,
-  mode_list,
-  mode_usage,
-};
```
kou: They are modes for command line options.
So we can't share them with `#define MODE_*`.

komainu8:
Thank you for your comments.
I've fixed it.

PR #1026, opened 2019-08-06
Add missing check about m4 macro availability

kenhys:
When RapidJSON support is enabled, proper compiler option must be
set by AX_CXX_COMPILE_STDCXX_11 macro. This m4 macro
is distributed in autoconf-archive package. We need to
enable it before regenerating configure script correctly.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -30,4 +30,21 @@ git submodule update --init
 
 mkdir -p m4
 
+# ax_cxx_compile_stdcxx macro is required to check compiler option correctly
+if [ -x /usr/bin/dpkg -o -x /bin/dpkg ]; then
+  if ! dpkg -s autoconf-archive > /dev/null; then
+    echo "ERROR: autoconf-archive package is not installed yet."
+    exit 1
+  fi
+elif [ -x /usr/bin/rpm -o -x /bin/rpm ]; then
+  if ! rpm -q autoconf-archive > /dev/null; then
+    echo "ERROR: autoconf-archive package is not installed yet."
+    exit 1
+  fi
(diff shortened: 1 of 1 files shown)


autogen.sh
```diff
@@ -30,4 +30,17 @@ git submodule update --init
...
+  if ! dpkg -s autoconf-archive > /dev/null; then
+    echo "ERROR: autoconf-archive package is not installed yet."
+    exit 1
+  fi
+elif [ -x /usr/bin/rpm -o -x /bin/rpm ]; then
+  if ! rpm -q autoconf-archive > /dev/null; then
+    echo "ERROR: autoconf-archive package is not installed yet."
+    exit 1
+  fi
+fi
```
komainu8: What if `[ -x /usr/bin/dpkg -o -x /bin/dpkg ]` is false and `[ -x /usr/bin/rpm -o -x /bin/rpm ]` is false?

kenhys: Ah, surely such a case is out-of-scope in above commit.
It should be fixed.

kenhys: I've fixed to check about existence of macro file as a fallback.

kenhys:
@komainu8 Thanks, I've fixed based on your review.

kou:
CentOS 6 doesn't provide `autoconf-archive`. So we didn't require `autoconf-archive` by `m4_ifdef([AX_CXX_COMPILE_STDCXX_11], ...)`.

kenhys:
Thanks, https://github.com/groonga/groonga/commit/724151708678eb11238d86db461b977fd244b1ab is a better approach.

PR #1028, opened 2019-08-07
Fix a bug that doesn't use a suitable index

### LINKED ISSUE #1007: Don't use index for full text search when prefix search is requested
See also: https://github.com/pgroonga/pgroonga/issues/96

komainu8:
Fixed #1007

### DIFF
--- a/test/command/suite/select/query/prefix_search/index_multiple.test
+++ b/test/command/suite/select/query/prefix_search/index_multiple.test
@@ -1,0 +1,46 @@
+table_create \
+  --name NamesFullTextSearchIndex \
+  --flags TABLE_PAT_KEY \
+  --key_type ShortText \
+  --default_tokenizer TokenBigramSplitSymbolAlphaDigit \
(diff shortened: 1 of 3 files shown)


komainu8:

test/command/suite/select/query/prefix_search/index_multiple.test
```diff
@@ -0,0 +1,42 @@
...
+  --type Users \
+  --source name
+column_create \
+  --table NamesIndex \
+  --name index \
+  --flags COLUMN_INDEX \
+  --type Users \
+  --source name
+
+select --table Users --query 'name:^"Ali"'
```
kou: Could you show used index by using the following:

```text
log_level --level info
#@add-important-log-levels info
select --table Users --query 'name:^"Ali"'
#@remove-important-log-levels info
log_level --level notice
```

komainu8:
I've fixed it.

PR #1030, opened 2019-08-14
Detect libedit at configure

yu:
The configure phase did not detect libedit despite installing dev files and this fix makes it work properly.
Verified on FreeBSD 11 and Debian 9.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -823,8 +823,8 @@ AM_CONDITIONAL([WITH_LEMON], [test "$lemon_available" = "yes"])
 # libedit
 libedit_available="no"
 AC_ARG_WITH([libedit],
-            AS_HELP_STRING([--with-libedit=PATH],
-                           [libedit path (default: auto)]),
+            AS_HELP_STRING([--with-libedit],
+                           [Support line editing (default: auto)]),
             [LIBEDIT="$withval"],
             [: ${LIBEDIT:=auto}])
 
--- a/configure.ac
+++ b/configure.ac
@@ -821,33 +821,41 @@ AC_SUBST(LEMON)
 AM_CONDITIONAL([WITH_LEMON], [test "$lemon_available" = "yes"])
 
 # libedit
-AC_ARG_ENABLE(libedit,
-  [AS_HELP_STRING([--disable-libedit],
-    [use libedit for console. [default=auto-detect]])],
-  [enable_libedit="$enableval"],
-  [enable_libedit="auto"])
-if test "x$enable_libedit" != "xno"; then
-  m4_ifdef([PKG_CHECK_MODULES], [
-    PKG_CHECK_MODULES([LIBEDIT], [libedit >= 3.0],
-      [LIBS_SAVE="$LIBS"
-       LDFLAGS_SAVE="$LDFLAGS"
-       LDFLAGS="$LIBEDIT_LIBS $LDFLAGS"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -821,33 +821,41 @@ AC_SUBST(LEMON)
...
-       LDFLAGS_SAVE="$LDFLAGS"
-       LDFLAGS="$LIBEDIT_LIBS $LDFLAGS"
-       AC_SEARCH_LIBS(el_wline, edit,
-                      [libedit_available=yes],
-                      [libedit_available=no])
-       LIBS="$LIBS_SAVE"
-       LDFLAGS="$LDFLAGS_SAVE"],
-      [libedit_available=no])
-    ],
-    [libedit_available=no])
```
kou: Could you remain pkg-config support?
Could you use pkg-config first then use `AC_CHECK_LIB()` fallback? Because pkg-config is more generic than `AC_CHECK_LIB()`.

configure.ac
```diff
@@ -821,33 +821,41 @@ AC_SUBST(LEMON)
...
+libedit_available="no"
+AC_ARG_WITH([libedit],
+            AS_HELP_STRING([--with-libedit],
+                           [Support line editing (default: auto)]),
+            [LIBEDIT="$withval"],
+            [: ${LIBEDIT:=auto}])
+
+if test "$LIBEDIT" = "no"; then
+  LIBEDIT=
+  AC_MSG_WARN([libedit is disabled. CLI does not support line editing.])
```
kou: We don't warn this because user disables libedit explicitly.

configure.ac
```diff
@@ -821,33 +821,41 @@ AC_SUBST(LEMON)
...
+else
+  if test "$LIBEDIT" = "auto"; then
+    AC_CHECK_LIB(edit, el_init,
+                 [LIBEDIT="yes"],
```
kou: We can set `LIBEDIT_LIBS="-ledit"; libedit_available="yes"` here

configure.ac
```diff
@@ -821,33 +821,41 @@ AC_SUBST(LEMON)
...
+                 [LIBEDIT="none"])
```
kou: We can use empty expression `[]` here.

configure.ac
```diff
@@ -821,33 +821,41 @@ AC_SUBST(LEMON)
...
-      [libedit_available=no])
-    ],
-    [libedit_available=no])
-  if test "x$libedit_available" = "xyes"; then
-    AC_DEFINE(GRN_WITH_LIBEDIT, [1], [Use libedit with multibyte support.])
+libedit_available="no"
+AC_ARG_WITH([libedit],
+            AS_HELP_STRING([--with-libedit],
+                           [Support line editing (default: auto)]),
+            [LIBEDIT="$withval"],
```
kou: If we use `--with-XXX` style than `--enable-XXX` style, we should respect `$withval` to find libedit.
See https://github.com/groonga/groonga/blob/master/configure.ac#L1139-L1150 for example.

yu:
Appreciate your review!

So basically b39ce25 checks conditions in a better way and deals with `--with-XXX` style, plus vanishes unnecessaries! Now pkg-config primally runs, and as its fallback AC_CHECK will take care of them with PATH info if it's mentioned so gets no pkgconf env (so was mine) covered.

PR #1036, opened 2019-08-26
packages ubuntu: make more robust archive detection

kenhys:
It supports tar.xz and so on.

### DIFF
--- a/packages/ubuntu/upload.rb
+++ b/packages/ubuntu/upload.rb
@@ -112,10 +112,11 @@ allow_unsigned_uploads = 0
 
   def upload(ubuntu_code_name, ubuntu_version)
     in_temporary_directory do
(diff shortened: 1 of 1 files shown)


packages/ubuntu/upload.rb
```diff
@@ -112,10 +112,11 @@ def parse_command_line!
 
   def upload(ubuntu_code_name, ubuntu_version)
     in_temporary_directory do
+      archive_basename, archive_suffix = @source_archive.to_s.split(".tar.")
```
kou: Could you use `split(..., 2)` to ensure splitting to 2 components.

kou: `File.basename(@source_archive.to_s).split(".tar.")` will be better.

PR #1048, opened 2019-09-30
packages apt: enable test during building deb package

kenhys:


packages/Rakefile
```diff
@@ -63,10 +63,7 @@ class GroongaPackageTask < PackageTask
     end
     yum_spec_in_file = File.join(__dir__, "groonga.spec.in")
     original_spec_in_file = File.join(__dir__, "rpm", "centos", "groonga.spec.in")
-    file yum_spec_in_file => original_spec_in_file do |task|
-      cp(original_spec_in_file,
-         task.name)
-    end
+    cp(yum_spec_in_file, yum_spec_in_file)
```
kou: You should not run any actions outside of task.
This runs with `rake -T`.

packages/apt/debian-buster/Dockerfile
```diff
@@ -15,19 +15,42 @@ RUN \
...
+  apt install -y -V ${quiet} \
+    cutter-testing-framework && \
   apt clean && \
   rm -rf /var/lib/apt/lists/*
+
+RUN \
+  git clone --depth 1 https://github.com/sstephenson/rbenv.git ${HOME}/.rbenv && \
+  git clone --depth 1 https://github.com/sstephenson/ruby-build.git ${HOME}/.rbenv/plugins/ruby-build && \
+  CONFIGURE_OPTS="--disable-install-rdoc" ${HOME}/.rbenv/bin/rbenv install 2.6.5 && \
+  ${HOME}/.rbenv/bin/rbenv global 2.6.5
```
kou: Why can not use system Ruby?

packages/debian/rules
```diff
@@ -10,15 +10,18 @@ export DEB_BUILD_HARDENING = 1
...
 %:
 	dh $@
 
 override_dh_auto_configure:
-	dh_auto_configure -- --with-munin-plugins --enable-mruby
+	dh_auto_configure -- --with-munin-plugins --with-ruby=$(shell rbenv which ruby) --enable-mruby
 
-# disable 'make check'.
 override_dh_auto_test:
+	bash -x ./test/unit/run-test.sh
```
kou: Is `bash -x` needed?

packages/yum/centos-6/Dockerfile
```diff
@@ -5,28 +5,40 @@ ARG DEBUG
 
 RUN \
   quiet=$([ "${DEBUG}" = "yes" ] || echo "--quiet") && \
-  yum install -y https://packages.groonga.org/centos/groonga-release-latest.noarch.rpm && \
   yum update -y ${quiet} && \
   yum install -y ${quiet} epel-release centos-release-scl && \
+  yum install -y https://packages.groonga.org/centos/groonga-release-latest.noarch.rpm && \
+  yum install -y http://iij.dl.osdn.jp/storage/g/c/cu/cutter/centos/cutter-release-1.3.0-1.noarch.rpm && \
```
kou: Why is https://osdn.net/projects/cutter/storage/centos/cutter-release-1.3.0-1.noarch.rpm not used?

kou:
I hope that you don't merge your pull request by yourself.
Normally, your pull request has several concerns without review.

packages/apt/debian-buster/Dockerfile
kenhys: grntest requires newer ruby. at least >= 2.2.

packages/yum/centos-6/Dockerfile
kenhys: It should be , but It seems that sometimes fails to fetch from osdn.net.

packages/debian/rules
kenhys: Yes, It may be helpful when test has failed.

PR #1050, opened 2019-10-10
Make a record of sending a message into a query log

komainu8:


lib/ctx.c
```diff
@@ -1474,14 +1474,14 @@ grn_ctx_send(grn_ctx *ctx, const char *str, unsigned int str_len, int flags)
...
-        GRN_QUERY_LOG(ctx, GRN_QUERY_LOG_RESULT_CODE,
-                      "<", "rc=%d", ctx->rc);
       }
     output :
       if (!(ctx->impl->command.flags & GRN_CTX_QUIET) &&
           ctx->impl->output.func) {
         ctx->impl->output.func(ctx, GRN_CTX_TAIL, ctx->impl->output.data.ptr);
       }
+      GRN_QUERY_LOG(ctx, GRN_QUERY_LOG_RESULT_CODE,
+                    "<", "rc=%d", ctx->rc);
```
kou: We should not log "rc=" while loading.

src/groonga.c
```diff
@@ -1209,6 +1209,9 @@ h_output_send(grn_ctx *ctx, grn_sock fd,
             "couldn't send all data (%" GRN_FMT_LLD "/%" GRN_FMT_LLD ")",
             (long long int)ret, (long long int)len);
   }
+  GRN_QUERY_LOG(ctx, GRN_QUERY_LOG_SIZE,
+                ":", "sendmsg(%zd)",
+                len);
```
kou: We have not only HTTP but also GQTP.

src/groonga.c
```diff
@@ -1209,6 +1209,9 @@ h_output_send(grn_ctx *ctx, grn_sock fd,
             "couldn't send all data (%" GRN_FMT_LLD "/%" GRN_FMT_LLD ")",
             (long long int)ret, (long long int)len);
   }
+  GRN_QUERY_LOG(ctx, GRN_QUERY_LOG_SIZE,
+                ":", "sendmsg(%zd)",
```
kou: `GRN_FMT_SIZE` should be used.

PR #1051, opened 2019-10-10
Add statistics for the number of requests in the queue into the output of the status command

komainu8:


lib/com.c
```diff
@@ -66,7 +66,18 @@
 #  endif
 #  define MSG_NOSIGNAL 0
 #endif /* USE_MSG_NOSIGNAL */
+
+static int64_t queue_count = 0;
```
kou: We should count only the size of the target queue. We should not count all size.

PR #1052, opened 2019-10-10
com: fix a bug that conditional expression always false

komainu8:
I think `if (!(q->next = e->next))` is always false.
So, `q->tail = &q->next;` doesn't execute.

Does this code on purpose, or a bug?


kou:
If `e->next` is not `NULL`, it's true.

PR #1056, opened 2019-10-21
packages: add support to build with docker

kenhys:


.github/workflows/debian.yml
```diff
@@ -46,10 +46,18 @@ jobs:
...
         run: |
-          cd packages
-          dch --newversion $(cat ../base_version)-1 "New upstream version"
+          OLD_RELEASE=$(grep -E -o '[0-9.]+' packages/debian/changelog | \
+                          head -n1)
+          OLD_RELEASE_DATE_FULL="$(grep '^ -- ' packages/debian/changelog | \
+                                     head -n1 | ^
+                                     sed -E -e 's/ -- .+<[^>]+>  //')"
+          OLD_RELEASE_DATE=$(date --date="${OLD_RELEASE_DATE}" +%Y-%m-%d)
+          make update-latest-release \
```
kenhys:
```
OLD_RELEASE_DATE=$(date --date="${OLD_RELEASE_DATE_FULL}" +%Y-%m-%d)
```

Should it be OLD_RELEASE_DATE_FULL?

kou: Good catch.

.github/workflows/debian.yml
```diff
@@ -0,0 +1,69 @@
...
+          git clone --depth 1 https://github.com/apache/arrow.git
+          git clone --depth 1 https://github.com/groonga/groonga.org.git
+          git clone --depth 1 https://github.com/clear-code/cutter.git
+      - name: Generate configure
+        run: |
+          ./autogen.sh
+      - name: Configure for archive
+        run: |
+          ./configure \
+            --cutter-source-path=../cutter \
```
kenhys: It seems that --with-cutter-source-path is valid

kenhys: I've fixed this commit.

PR #1058, opened 2019-10-23
packages: support to build with docker (YUM_TARGETS)

kenhys:


kou (commented):
Why did you create a new workflow file?
Can we reuse `debian.yml`?

kenhys:
Hmm, at first, debian specific task may be added in the future work, so it may be better to split into new workflow.
But it seems that enaming debian.yml -> package.yml and adding centos into matrix may be better 
at the moment. 

I'll fix them.

PR #1063, opened 2019-10-29
packages yum: support CentOS 8

kenhys:
Because of some *-devel packages are missing, just disable them for a
while.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1416,6 +1416,7 @@ AM_CONDITIONAL([CENTOS_INIT_PLATFORM],
                 test "${package_platform}" = "centos6"])
 AM_CONDITIONAL([CENTOS_SYSTEMD_PLATFORM],
                [test "${package_platform}" = "centos7" ||
+                test "${package_platform}" = "centos8" ||
                 test "${package_platform}" = "fedora"])
 
 # plugins check
--- a/packages/yum/centos-8/Dockerfile
+++ b/packages/yum/centos-8/Dockerfile
@@ -1,0 +1,29 @@
+ARG FROM=centos:8
+FROM ${FROM}
+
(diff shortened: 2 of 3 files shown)


kou:
We should use `dnf` instead of `yum`.

configure.ac
```diff
@@ -1416,6 +1416,7 @@ AM_CONDITIONAL([CENTOS_INIT_PLATFORM],
                 test "${package_platform}" = "centos6"])
 AM_CONDITIONAL([CENTOS_SYSTEMD_PLATFORM],
                [test "${package_platform}" = "centos7" ||
+                test "${package_platform}" = "centos8" ||
                 test "${package_platform}" = "fedora"])
```
kou: `test "${package_platform}" != "centos6"`

We need to add this list when CentOS 9 is release if we still use this logic.

packages/yum/centos-8/Dockerfile
```diff
@@ -0,0 +1,29 @@
+ARG FROM=centos:8
+FROM ${FROM}
+
+ARG DEBUG
+
+RUN \
+  quiet=$([ "${DEBUG}" = "yes" ] || echo "--quiet") && \
+  yum install -y https://packages.groonga.org/centos/groonga-release-latest.noarch.rpm && \
```
kou: Is this needed?

configure.ac
kenhys: Thanks, I'll fix this.

packages/yum/centos-8/Dockerfile
kenhys: No, I'll remove it.

kenhys:
I've fixed them. Thanks!

PR #1065, opened 2019-10-31
configure: fix test(1) portability

obache:
`==` operator for test(1) is not in the standard, then not portable.


kenhys:
@obache Good catch!

PR #1074, opened 2019-12-06
Return error response when fail to wait

komainu8:
I'm not sure if this imprements is correct.
Therefore, Could you review this.


komainu8:
This modify is needless.

PR #1078, opened 2020-01-22
github-actions package: remove redundant definition

komainu8:


kou:
How about removing `on.pull_request.branch` too?

komainu8:
Ah... That is true for sure.
I will remove `on.pull_request.branch` too.

komainu8:
@kou 
Thank you for your comment.
I removed `on.pull_request.branch` too.

PR #1102, opened 2020-05-10
select: add cache for dynamic columns value

naoa:


lib/proc/proc_select.c
```diff
@@ -3690,6 +3690,25 @@ grn_select(grn_ctx *ctx, grn_select_data *data)
         sizeof(int) * 2;
     } GRN_HASH_EACH_END(ctx, cursor);
   }
+#define ADD_COLUMNS_CACHE_SIZE(columns) do {                     \
+    GRN_HASH_EACH_BEGIN(ctx, columns, cursor, id) {              \
+      grn_column_data *column;                                   \
+      grn_hash_cursor_get_value(ctx, cursor, (void **)&column);  \
+      cache_key_size +=                                          \
+        column->value.length + 1;                                \
```
kou, line 3698: We need to add not only the value parameter but also all parameters to cache key.
If we only add the value parameter, we may use wrong cache.

naoa: Thanks for reviewing.

I've added the other parameters of column data.

kou:
Thanks!

PR #1107, opened 2020-05-20
io: fix a wrong error code

komainu8:
I think ``GRN_INVALID_ARGUMENT`` is not suitable in this case.
Because this error cause is Groonga can't allocate memory.


kou:
Good catch!

PR #1120, opened 2020-07-09
aggregator_*: support score accessor for target

naoa:


kou:
Thanks!
I've add tests for them.

PR #1126, opened 2020-08-24
Optimize offline index construction using token column

naoa:
It is 3x faster with source data size of about 13GiB.

source_column size: 12.840GiB

Before: 1499 sec
Patched: 412 sec

### DIFF
--- a/lib/ii.c
+++ b/lib/ii.c
@@ -14606,17 +14606,35 @@ grn_ii_builder_append_tokens(grn_ctx *ctx,
   size_t i;
   for (i = 0; i < n_tokens; i++) {
     float weight;
+    grn_id tid;
     grn_id src_tid = grn_uvector_get_element_record(ctx, tokens, i, &weight);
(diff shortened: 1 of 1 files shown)


lib/ii.c
```diff
@@ -14606,17 +14606,35 @@ grn_ii_builder_append_tokens(grn_ctx *ctx,
...
+                        token_value, token_value_size, NULL, NULL);
+      break;
+    case GRN_TABLE_HASH_KEY :
+      tid = grn_hash_add(ctx, (grn_hash *)builder->lexicon,
+                         token_value, token_value_size, NULL, NULL);
+      break;
+    case GRN_TABLE_NO_KEY :
+      tid = *(grn_id *)token_value;
+      break;
+    default :
```
kou, line 14633:
```suggestion
    default :
      /* This case must not be happen. */
```

kou:
Thanks!
It's a big improvement!

PR #1129, opened 2020-08-25
package groonga-tokenizer-mecab: add support for CentOS8

komainu8:
groonga-tokenizer-mecab requires mecab-devel package.
However, currently, mecab-devel package doesn't provide in the repository of CentOS8.

Therefore, we provide mecab-devel package in the repository of Groonga(packages.groonga.org) for a while.
Thereby we can provide groonga-tonkenizer-mecab on CentOS8.
We can build groonga-tokenizer-mecab package by this PR.

### DIFF
--- a/packages/yum/centos-8/Dockerfile
+++ b/packages/yum/centos-8/Dockerfile
@@ -27,6 +27,5 @@ RUN \
     tar \
     zlib-devel && \
   dnf install -y ${quiet} https://packages.groonga.org/centos/8/groonga-release-latest.noarch.rpm && \
-  dnf install -y --enablerepo=epel groonga && \
   dnf install -y ${quiet} mecab-devel && \
   dnf clean ${quiet} all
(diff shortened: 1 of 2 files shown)


packages/yum/centos-8/Dockerfile
```diff
@@ -26,4 +26,7 @@ RUN \
     ruby \
     tar \
     zlib-devel && \
+  dnf install -y ${quiet} https://packages.groonga.org/centos/8/groonga-release-latest.noarch.rpm && \
+  dnf install -y --enablerepo=epel groonga && \
```
kou, line 30: Do we need this?

packages/yum/centos-8/Dockerfile
```diff
@@ -26,4 +26,7 @@ RUN \
     ruby \
     tar \
     zlib-devel && \
+  dnf install -y ${quiet} https://packages.groonga.org/centos/8/groonga-release-latest.noarch.rpm && \
```
kou, line 29: Can we merge this into `dnf install -y ${quiet} epel-release 'dnf-command(config-manager)'`?

packages/yum/centos-8/Dockerfile
komainu8: This is needless.
I'll delete it.

komainu8:
@kou 
Thank you for your comments.
I've fixed them.

PR #1133, opened 2020-09-10
Add a tool for upload packages from GitHub release

komainu8:


kou:
I've changed to use existing `apt:release` and `yum:release` for uploading downloaded packages.

komainu8:
> I've changed to use existing `apt:release` and `yum:release` for uploading downloaded packages.

I see. Thank you.

PR #1134, opened 2020-09-15
doc normalizer: add a reference for NormalizerNFKC121

komainu8:
This is a work in progress.


komainu8:
I close this PR.
Because the reference of NormalizerNFKC121 is already an official document.

PR #1138, opened 2020-10-06
Add a tool for upload windows packages from GitHub release

komainu8:


komainu8:
Ah... This is conflict yes.
I'll fix them.

komainu8:
I've fixed them.

packages/windows-helper.rb
```diff
@@ -0,0 +1,52 @@
...
+      "x64-exe",
+      "x86-vs2019",
+      "x86-vs2019-with-vcruntime",
+      "x64-vs2019",
+      "x64-vs2019-with-vcruntime",
+      "x86-vs2017",
+      "x86-vs2017-with-vcruntime",
+      "x64-vs2017",
+      "x64-vs2017-with-vcruntime",
+    ]
```
kou, line 50: How about using the empty array as the default? It seems that this default value is useless. 

```suggestion
    []
```

packages/packages-groonga-org-package-task.rb
```diff
@@ -82,10 +84,15 @@ def download_packages(target_namespace)
...
+        mkdir_p("#{repositories_dir}/windows/groonga/")
+        cp(archive, "#{repositories_dir}/windows/groonga/", {:preserve => true})
+      else
+        cd(repositories_dir) do
+          sh("tar",
+             "xf", archive,
+             "--strip-components=#{built_package_n_split_components}")
+        end
       end
       rm_f(archive)
```
kou, line 97: How about using existing variables instead of literals such as `windows` and `groonga`?

```suggestion
      case target_namespace
      when :apt, :yum
        cd(repositories_dir) do
          sh("tar",
             "xf", archive,
             "--strip-components=#{built_package_n_split_components}")
        end
        rm_f(archive)
      else
        download_dir = "#{repositories_dir}/#{target_namespace}/#{@package}"
        mkdir_p(download_dir)
        mv(archive, download_dir)
      end
```

packages/Rakefile
```diff
@@ -42,6 +42,19 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
...
+  def windows_targets_default
+    [
+      "x86-vs2019",
+      "x86-vs2019-with-vcruntime",
+      "x64-vs2019",
+      "x64-vs2019-with-vcruntime",
+      "x86-vs2017",
+      "x86-vs2017-with-vcruntime",
+      "x64-vs2017",
+      "x64-vs2017-with-vcruntime",
```
kou, line 54: How about using exact file names for targets? It will simplify `built_package_url`.

```suggestion
      "#{@package}-#{@version}-x64-vs2019.zip",
      "#{@package}-#{@version}-x64-vs2019-with-vcruntime.zip",
      "#{@package}-#{@version}-x86-vs2019.zip",
      "#{@package}-#{@version}-x86-vs2019-with-vcruntime.zip",
      "#{@package}-#{@version}-x64-vs2017.zip",
      "#{@package}-#{@version}-x64-vs2017-with-vcruntime.zip",
      "#{@package}-#{@version}-x86-vs2017.zip",
      "#{@package}-#{@version}-x86-vs2017-with-vcruntime.zip",
      "#{@package}-#{@version}-x64.exe",
      "#{@package}-#{@version}-x64.zip",
      "#{@package}-#{@version}-x86.exe",
      "#{@package}-#{@version}-x86.zip",
```

packages/packages-groonga-org-package-task.rb
```diff
@@ -17,10 +17,12 @@
 
 require_relative "../vendor/apache-arrow-source/dev/tasks/linux-packages/package-task"
 require_relative "launchpad-helper"
+require_relative "windows-helper"
```
kou, line 20: It seems that we don't need to split this to separated file because it's not reusable not like `launchpad-helper`.

komainu8:

packages/Rakefile
```diff
@@ -50,8 +67,8 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
     url = "https://github.com/groonga/groonga/releases/download/v#{@version}/"
     if target_namespace == :apt and !target.end_with?("-i386")
       url << "#{target}-amd64.tar.gz"
-    else
-      url << "#{target}.tar.gz"
```
kou, line 54: Does this still work with `:yum` `target_namespace`?

komainu8: Ah...
Yes. This still works with yum and apt for x64.

I added this in e2847366adb9e7819e5a55cc66374dba662e3f3b

kou:
* Updating latest links is missing: https://packages.groonga.org/windows/groonga/XXX-latest.XXX
* Uploading source archives is missing: https://packages.groonga.org/windows/groonga/groonga-10.0.7/

komainu8:
I added updating latest links and uploading source archives.

packages/packages-groonga-org-package-task.rb
```diff
@@ -73,21 +90,77 @@ def built_package_n_split_components
...
+      "groonga-normalizer-mysql",
+      "lz4",
+      "mecab",
+      "msgpack",
+      "rapidjson",
+      "xxHash",
+    ]
+  end
+
+  def download_bundle_packages(download_dir)
```
kou, line 111: We can use sources collected by package build job.

PR #1149, opened 2020-11-30
ci: add a patch that don't check FD for STDIN/STDOUT/STDERR

komainu8:
This is testing now.

### DIFF
--- a/.github/workflows/linux-mingw.yml
+++ b/.github/workflows/linux-mingw.yml
@@ -44,6 +44,8 @@ jobs:
       - name: Generate configure
         run: |
           ./autogen.sh
+      - name: Apply a patch
+          patch -p1 < packages/source/patches/mruby-do-not-check-fd-for-stdin-stdout-stderr.diff
       - name: Configure
         run: |
           ./configure \
@@ -56,7 +58,6 @@ jobs:
           make dist
       - name: Build
         run: |
-          patch -p1 < packages/source/patches/mruby-do-not-check-fd-for-stdin-stdout-stderr.diff
           make \
             -C packages/windows \
             ARCHITECTURES=${{ matrix.architecture }} \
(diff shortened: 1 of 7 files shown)


kou:
The patch should be applied before `configure` or `cmake`.

komainu8:
I see.
I'll fix it.

.github/workflows/linux-mingw.yml
```diff
@@ -44,6 +44,8 @@ jobs:
       - name: Generate configure
         run: |
           ./autogen.sh
+      - name: Apply a patch
+          patch -p1 < packages/source/patches/mruby-do-not-check-fd-for-stdin-stdout-stderr.diff
```
kou, line 48: `run:` is missing. This breaks CI.

.github/workflows/linux.yml
```diff
@@ -88,6 +88,8 @@ jobs:
       - name: Generate configure
         run: |
           ./autogen.sh
+      - name: Apply a patch
```
kou, line 91: `run:` is missing. This breaks CI.

.github/workflows/macos.yml
```diff
@@ -31,6 +31,8 @@ jobs:
           matrix.use-autotools
         run: |
           ./autogen.sh
+      - name: Apply a patch
```
kou, line 34: `run:` is missing. This breaks CI.

.github/workflows/windows-mingw.yml
```diff
@@ -28,10 +28,13 @@ jobs:
             mingw-w64-x86_64-ruby
             mingw-w64-x86_64-xxhash
             mingw-w64-x86_64-zlib
+            patch
       - name: Generate configure
         shell: msys2 {0}
         run: |
           ./autogen.sh
+      - name: Apply a patch
```
kou, line 36: `run:` is missing. This breaks CI.

komainu8:
Thank you for your comment.
I am aware of it. I am fixing them now.

PR #1158, opened 2020-12-24
TokenMecab: Support lock free parsing

naoa:
`GRN_MECAB_LOCK_FREE=yes` environment variable to enable this feature.

It has some overhead, but it is effective for parallel construction of token column.

| patched:env                   | Sequential construction of token column | Parallel construction of token column |
|-------------------------------|-----------------------------------------|---------------------------------------|
| Before                        | 191.22sec                               | 595.12sec                             |
| After                         | 191.68sec                               | 673.04sec                             |
| After:GRN_MECAB_LOCK_FREE=yes | 239.56sec                               | 90.19sec                              |

https://taku910.github.io/mecab/libmecab.html


plugins/tokenizers/mecab.c
```diff
@@ -110,14 +113,19 @@ grn_mecab_init(grn_ctx *ctx,
                grn_mecab *mecab,
                const char *tag)
 {
+  mecab->mecab_model = NULL;
   mecab->mecab = NULL;
-  mecab->mutex = grn_plugin_mutex_open(ctx);
```
kou, line 114: We still always needs the mutex for `mecab_init_mecab()`.
We should avoid creating `mecab_model_t *` in multiple threads.

plugins/tokenizers/mecab.c
```diff
@@ -1083,6 +1109,11 @@ mecab_init(grn_ctx *ctx, grn_tokenizer_query *query)
     GRN_PLUGIN_FREE(ctx, tokenizer);
     return NULL;
   }
+  tokenizer->lattice = mecab_model_new_lattice(tokenizer->mecab->mecab_model);
```
kou, line 1112: This should not be created when `grn_mecab_lock_free == false`. It will improve performance on parallel case.

kou (commented):
Interesting!

I hope that we improve parallel case performance without decreasing sequential case performance and `grn_mecab_lock_free` configuration.

naoa:
Thank you for reviewing.
I've fixed and squashed.

The following is the remeasured.

| patched:env                   | Sequential construction of token column | Parallel construction of token column |
|-------------------------------|-----------------------------------------|---------------------------------------|
| Before                        | 190.41sec                               | 570.44sec                             |
| After                         | 190.82sec                               | 453.11sec                             |
| After:GRN_MECAB_LOCK_FREE=yes | 238.20sec                               | 96.02sec                              |

kou:
It seems that the `After:GRN_MECAB_LOCK_FREE=yes` & `Sequential construction of token column` case's slow down is caused by always creating `mecab_lattice_t`. `mecab_sparse_tostr2()` reuses created `mecab_lattice_t`.

We'll be able to improve the performance of the case by caching `mecab_latitce_t` in `grn_ctx`. I've added `grn_ctx` local variables features into master and pushed a change to use the feature to cache `mecab_lattice_t`.

Could you measure again with the latest commit? You don't need to use `GRN_MECAB_LOCK_FREE=yes` because I removed the variable.

naoa:
Great!
I've measured it.
There was no effect in the sequential case, and it was faster in the parallel case.

patched | Sequential construction of token column | Parallel construction of token column
-- | -- | --
Before |190.48sec | 683.14sec
After | 190.62sec | 70.67sec

kou:
OK. I'll merge this.
Thanks!

PR #1162, opened 2021-01-25
packages windows: fix a wrong repositry directory path for Windows

komainu8:


packages/packages-groonga-org-package-task.rb
```diff
@@ -104,7 +104,12 @@ def built_package_n_split_components
 
   def download_packages(target_namespace)
     base_dir = __send__("#{target_namespace}_dir")
-    repositories_dir = "#{base_dir}/repositories"
+    case target_namespace
+    when :windows
+      repositories_dir = "#{base_dir}/repositories/windows/groonga"
```
kou, line 109: This task is used by packages provided on `packages.groonga.org` such as Mroonga and PGroonga.
If we put Groonga specific logic to here, we can't share this task with Mroonga and PGroonga.

Can we use the use `#{@package}` instead of `groonga` here?

komainu8: Oh...I understand.
I will use #{@package}.

PR #1167, opened 2021-03-08
package apt: add support for arm64 package

komainu8:


kou:
TODO:

  * Reduce build time by ccache
  * Fix test failures caused by floating point number

PR #1172, opened 2021-04-04
Remove references to long-deleted ruby_load function

acook:
There were some old docs I found which talk about removed functionality. 
This PR removes them.

<3


kou:
Good catch!

PR #1174, opened 2021-04-12
packages debian: use the qemu-user-static >= 5.0

komainu8:
Because qemu-user-static has crash bug in version before 5.0.
(See: https://bugs.launchpad.net/qemu/+bug/1749393)
By this bug, host OS may crashes.

TODO: Test


komainu8:
Ah... GitHub Actions doesn't support Ubuntu 20.10 yet.

PR #1175, opened 2021-04-13
expr: add support for multiple token filters with options

komainu8:
Fix mroonga/mroonga#399.

If we specify multiple token filters with options like `--token_filters 'TokenFilterStopWord("column", "ignore"), TokenFilterNFKC130("unify_kana", true)'`, Groonga parses these token filters as below.

```
#<expr
  vars:{
    $1:#<record:pat:Terms id:(no value)>
  },
  codes:{
    0:<push n_args:1, flags:0, modify:3, value:#<proc:token-filter TokenFilterStopWord arguments:[]>>,
    1:<push n_args:1, flags:0, modify:0, value:"column">,
    2:<push n_args:1, flags:0, modify:0, value:"ignore">,
    3:<call n_args:3, flags:0, modify:0, value:(NULL)>,
    4:<push n_args:1, flags:0, modify:3, value:#<proc:token-filter TokenFilterNFKC130 arguments:[]>>,
    5:<push n_args:1, flags:0, modify:0, value:"unify_kana">,
    6:<push n_args:1, flags:0, modify:0, value:true>,
    7:<call n_args:3, flags:0, modify:0, value:(NULL)>,
    8:<comma n_args:2, flags:0, modify:0, value:(NULL)>
  }
```

However, the current implemantation assumes  that a `call` operation appears the end of `codes`.
This commit supports multiple token filters with options.


kou:
Could you rebase on the master?

PR #1176, opened 2021-04-13
package: add a new task for installing qemu-user-static from Groovy

komainu8:
Because qemu-user-static in Ubuntu 20.04 has a crash bug as below.

  * https://bugs.launchpad.net/qemu/+bug/1749393
  * https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=971537

This bug has already fixed in Ubuntu 20.10.
However, this bug isn't fixed in Ubuntu 20.04.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -68,7 +68,7 @@ jobs:
             devscripts \
             python3-pip \
             ruby
-      - name: Install Qemu from Groovy
+      - name: Install QEMU from Groovy
         run: |
           sudo tee /etc/apt/preferences.d/qemu <<EOF
           Package: *
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -67,8 +67,26 @@ jobs:
             autoconf-archive \
             devscripts \
             python3-pip \
-            qemu-user-static \
             ruby
+      - name: Install Qemu from Groovy
+        run: |
+          sudo tee /etc/apt/preferences.d/qemu <<EOF
+          Package: *
+          Pin: release n=focal
+          Pin-Priority: 900
+          Package: *
+          Pin: release n=groovy
+          Pin-Priority: 400
+          EOF
+          cat /etc/apt/preferences.d/qemu
+          sudo tee /etc/apt/sources.list.d/groovy.list <<EOF
+          deb http://jp.archive.ubuntu.com/ubuntu groovy universe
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -67,8 +67,26 @@ jobs:
...
+        run: |
+          sudo tee /etc/apt/preferences.d/qemu <<EOF
+          Package: *
+          Pin: release n=focal
+          Pin-Priority: 900
+          Package: *
+          Pin: release n=groovy
+          Pin-Priority: 400
+          EOF
+          cat /etc/apt/preferences.d/qemu
```
kou, line 81: Why is `cat` needed? `tee` also shows content to stdout.

.github/workflows/package.yml
```diff
@@ -67,8 +67,26 @@ jobs:
...
+          sudo tee /etc/apt/sources.list.d/groovy.list <<EOF
+          deb http://jp.archive.ubuntu.com/ubuntu groovy universe
+          deb http://jp.archive.ubuntu.com/ubuntu groovy-updates universe
+          deb http://security.ubuntu.com/ubuntu groovy-security universe
+          EOF
+          cat /etc/apt/sources.list.d/groovy.list
```
kou, line 87: Ditto.

.github/workflows/package.yml
```diff
@@ -67,8 +67,26 @@ jobs:
             autoconf-archive \
             devscripts \
             python3-pip \
-            qemu-user-static \
             ruby
+      - name: Install Qemu from Groovy
```
kou, line 71:
```suggestion
      - name: Install QEMU from Groovy
```

kou: Could you add a comment that describes why we need this and when we can remove this?

.github/workflows/package.yml
komainu8: Ah... `cat` is needless.
I'll fix it.

.github/workflows/package.yml
kou: @komainu8 Did you miss https://github.com/groonga/groonga/pull/1176#discussion_r612981895 ?

komainu8: No. I confirm it.
I'm fixing now. Please wait a little longer.

komainu8:

PR #1179, opened 2021-04-28
sort_keys: add missing sort flag assignments

komainu8:
GitHub: fix GH-1178.
There are paths that don't assign a sort flag to "key->flags".
By this, sort flag is unexpected vale and Groonga doesn't output sort results.

### DIFF
--- a/lib/sort.c
+++ b/lib/sort.c
@@ -1250,7 +1250,6 @@ grn_table_sort_keys_parse_one(grn_ctx *ctx,
   if (n_codes == 1 &&
       expr->codes[code_start_offset].op == GRN_OP_GET_VALUE &&
       expr->codes[code_start_offset].value) {
-    key->flags = flags;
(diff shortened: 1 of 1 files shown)


lib/sort.c
```diff
@@ -1264,6 +1263,7 @@ grn_table_sort_keys_parse_one(grn_ctx *ctx,
                                           code_end_offset);
     key->key = sliced_expr;
   }
+  key->flags = flags;
```
kou, line 1266: Could you move this to the above line of `if (n_codes == 1 &&`?
We can reduce code range that uses `flags` by the move.

kou (commented):
Good catch!

lib/sort.c
komainu8: I see.
I've fixed it in https://github.com/groonga/groonga/pull/1179/commits/7fdaf663ce31591c4ab4394905e7c03f83262a11.

PR #1191, opened 2021-05-13
Add editorconfig for Visual Studio

HashidaTKS:
Add editorconfig for Visual Studio
  - Automatically generated from the Visual Studio C/C++ format settings
  - Most settings are for Visual Studio only
  - Add some common settings

Strictly it doesn't match the current coding style because of the following reasons.

**No settings ( I couldn't find )**

* Space between label and colon
  *  `exit :` 
  *  `case GRN_TABLE_PAT_KEY :` 
  * In Visual Studio, these are formatted to `exit:` and `case GRN_TABLE_PAT_KEY:`. 

**Not unified on the code side**

* Newline before open brace of function
  * `bool grn_dat_rebuild_trie(grn_ctx *ctx, grn_dat *dat) {`
  * ```
     bool
     grn_dat_open_trie_if_needed(grn_ctx *ctx, grn_dat *dat)
     {
     ```
  * In this editorconfig, always adding newline.

* Indent for namespace
  * ```
    namespace grn {
      namespace bulk {
        template <typename NUMERIC>
        NUMERIC
        ...
    ```
  * ``` 
    namespace grn {
    namespace dat {

     // This class is used to detect an out-of-range access in debug mode.
    template <typename T>
    ...
    ```
  * In this editorconfig, always indenting.
* Indent for access modifier
  * ```
     class GRN_DAT_API Array {
      public:
       Array() : ptr_(NULL), size_(0) {}
     ```
    * It seems that `.hpp` files use this style.
  * ```
     class Array
     {
     public:
       Array(grn_ctx *ctx)
     ```
  * In this editorconfig, always not indenting.

Because of these subtle differences in style, I think it's necessary to re-format manually after auto formatting.
...It may be better to use clang-format instead of this editorconfig so that all editors can use same format and unify these differences.


komainu8:
I can't decide whether we accept this PR  or not.

Because the purpose of this modification is that developers write codes without being conscious of coding style.
But we need to modify the part of the coding style even if we use this. However, I think that mistakes about coding style reduce by this PR.

@kou What do you think about whether we accept this PR  or not?

kou:
Generally, we should consider pros and cons to decide.
In this case, there are some pros but there are no cons:

* Developers who use Visual Studio can reduce manual style adjustments.
* We can unify style configuration for Emacs and Visual Studio.
 
This is not perfect but better than the current situation. So, we can accept this improvement.

kou:
> Space between label and colon

OK. We can switch to `LABEL:` (no space between `LABEL` and `:`) style. Because there is no active developer that likes the current `LABEL :` style.

> Newline before open brace of function

We can unify to newline before open brace of function style. Because there is no active developer that likes no newline before open brace of function style.

> Indent for namespace

We can unity to always indent style. Because there is no active developer that likes no indent for namespace.

> Indent for access modifier

We can unity to no indent for access modifier style. Because there is no active developer that likes indent for access modifier.

We can change the current style but we don't want to do batch re-format. We improve the current style when we touch each file.

kou:
I simplified `editorconfig` a bit and removed style configurations for Emacs. Because we can use `.editorconfig` configuration with Emacs.

kou:
> It may be better to use clang-format instead of this editorconfig so that all editors can use same format and unify these differences.

If we use this approach in the future, we should also configure format check CI job.

kou (approved):

PR #1192, opened 2021-05-21
tool output current n_buffer_segments within database: add

komainu8:
The purpose of this tool is for monitoring the `n_buffer_segments` value.
This tool outputs `n_buffer_segments` to files.

These files are named an index name.
The value of n_buffer_segments and the time when this script is executed is written each file.

```
1622634010.103554, 12
1622634010.115556, 34
1622634010.127578, 35
.
.
.
```

If the error occurs in this script, this script output error logs as `output_n_buffer_segments.error` as below.

```
2021-06-03 14:19:45 +0900: #<Process::Status: pid 32082 exit 1>: [[-22,1622697585.43357,0.0001327991485595703,"invalid command name: dumpAAA",[["grn_ctx_qe_exec","ctx.c",1747]]]]: 
```

`n_buffer_segments` is the number of segments used for “buffer” in the target index column.
“buffer” is used for storing posting lists.

Groonga has the limitation of the number of buffers. (66535)
We can't update the index when it reaches the limitation. Therefore, we want to monitor this value periodically.

This tool useful in such cases.


HashidaTKS (approved):
LGTM

Just curious, what purpose is this command used for?

komainu8:
@HashidaTKS 
Thank you for your comment.
I've added the purpose of this tool in the description of this PR.

HashidaTKS:
Thank you!

tools/output-number-of-buffer-segments.sh
```diff
@@ -0,0 +1,25 @@
...
+        --dump_configs no | \
+  awk '{ print "object_inspect", $2 "." $3 }' | \
+  while read line; do \
+    echo "${line}" | \
+      groonga \
+        --log-path groonga.log \
+        --log-level debug \
+        "${DATABASE_PATH}" | \
+    jq -r '.[1] | {full_name: .full_name, n_buffer_segments: .value.statistics.n_buffer_segments}'
+  done
```
kou, line 25: Do we need this `while`?

```suggestion
  groonga \
    --log-path groonga.log \
    --log-level debug \
    "${DATABASE_PATH}" | \
  jq -r '.[1] | {full_name: .full_name, n_buffer_segments: .value.statistics.n_buffer_segments}'
```

komainu8:
I rewrite this script by ruby.
And I changed to output the value of `n_buffer_segments to a file.

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,31 @@
...
+end
+
+DATABASE_PATH = ARGV[0]
+stdout, stderr, status = Open3.capture3("groonga",
+                                        DATABASE_PATH,
+                                        "dump",
+                                        "--dump_plugins","no",
+                                        "--dump_schema", "no",
+                                        "--dump_records", "no",
+                                        "--dump_configs", "no")
```
HashidaTKS, line 19: How about check the status and output `stdout` and `stderr` if an error occurred?

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,31 @@
...
+stdout.each_line do |line|
+  index_name = "#{line.split()[1]}.#{line.split()[2]}"
+  stdout, stderr, status = Open3.capture3("groonga",
+                                          DATABASE_PATH,
+                                          "object_inspect",
+                                          index_name)
+  unix_time_when_command_executed = JSON.parse(stdout)[0][1]
```
HashidaTKS, line 26: How about check the status and output stdout and stderr if an error occurred?

tools/output-number-of-buffer-segments.rb
komainu8: Thank you for your comment.
I see. I've added them at https://github.com/groonga/groonga/pull/1192/commits/e2273898490c01b39621a408e7c9f18edae9cbb3.

tools/output-number-of-buffer-segments.rb
komainu8: Thank you for your comment.
I see. I've added them at e227389.

HashidaTKS (approved):

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,49 @@
+#!/bin/ruby
+
+require "open3"
+require "json"
+require "date"
+
+if ARGV.size() == 0
```
kou, line 7:
```suggestion
if ARGV.empty?
```

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,49 @@
...
+  puts "Usage: $0 DATABASE_PATH"
+  puts " e.g.: $0 ~/database/db"
+  exit
```
kou, line 10:
```suggestion
  exit(false)
```

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,49 @@
...
+end
+
+stdout, stderr, status = Open3.capture3("groonga",
+                                        DATABASE_PATH,
+                                        "dump",
+                                        "--dump_plugins","no",
+                                        "--dump_schema", "no",
+                                        "--dump_records", "no",
+                                        "--dump_configs", "no")
+if (status.success?)
```
kou, line 29:
```suggestion
if status.success?
```

We can reduce indent by:

```ruby
unless status.success?
  output_error(status, stdout, stderr)
  exit(false)
end

File: tools/output-number-of-buffer-segments.rb [hunk 7]
```diff
@@ -0,0 +1,49 @@
...
+  puts "Usage: $0 DATABASE_PATH"
+  puts " e.g.: $0 ~/database/db"
+  exit
+end
+
+DATABASE_PATH = ARGV[0]
+ERROR_LOG_NAME = "output_n_buffer_segments.error"
+
+def output_error(status, stdout, stderr)
+  File.open(ERROR_LOG_NAME, 'a') do |file|
```
kou, line 17: Why don't you just output error message to the standard error?

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,49 @@
...
+if (status.success?)
+  stdout.each_line do |line|
+    index_name = "#{line.split()[1]}.#{line.split()[2]}"
+    stdout, stderr, status = Open3.capture3("groonga",
+                                            DATABASE_PATH,
+                                            "object_inspect",
+                                            index_name)
+    if (status.success?)
+      unix_time_when_command_executed = JSON.parse(stdout)[0][1]
+      n_buffer_segments = JSON.parse(stdout)[1]["value"]["statistics"]["n_buffer_segments"]
```
kou, line 38: Can we avoid duplicated `JSON.parse`?

tools/output-number-of-buffer-segments.rb
```diff
@@ -0,0 +1,49 @@
...
+stdout, stderr, status = Open3.capture3("groonga",
+                                        DATABASE_PATH,
+                                        "dump",
+                                        "--dump_plugins","no",
+                                        "--dump_schema", "no",
+                                        "--dump_records", "no",
+                                        "--dump_configs", "no")
+if (status.success?)
+  stdout.each_line do |line|
+    index_name = "#{line.split()[1]}.#{line.split()[2]}"
```
kou, line 31:
```suggestion
    index_name = line.split[1..2].join(".")
```

tools/output-number-of-buffer-segments.rb
HashidaTKS: Excuse me for cutting in.

How about exiting with `exit(status.exit status)` if we need return the exit status of this script?

tools/output-number-of-buffer-segments.rb
komainu8: I predict that this tool is executed periodically with crontab and so on.
Therefore, I think that It is easier to use its output to file rather than output to stdout.

komainu8:
Thank you for your comments. I've fixed them.

tools/output-number-of-buffer-segments.rb
kou: ping

tools/output-number-of-buffer-segments.rb
kou: If we use crontab, it can e-mail automatically when command outputs something.
In this case, users don't need to check error log file on remote machine.

tools/output-number-of-buffer-segments.rb
komainu8: Ah...I'm Sorry.
This was a correction omission.

tools/output-number-of-buffer-segments.rb
komainu8: Right.
I've changed that error log output to stdout.

HashidaTKS (approved):

komainu8:
Thank you for your comments. I've fixed them.

PR #1193, opened 2021-05-25
GRN_TOKEN_FORCE_PREFIX: fix documentation

kou:


kou:
Cc: @HashidaTKS

include/groonga/token.h
```diff
@@ -110,7 +110,7 @@ typedef unsigned int grn_token_status;
  */
 #define GRN_TOKEN_SKIP_WITH_POSITION (0x01L<<5)
 /*
- * GRN_TOKEN_FORCE_PREIX that the token is used common prefix search
+ * GRN_TOKEN_FORCE_PREIX means that the token is used common prefix search
```
komainu8, line 113: I found a typo.

```suggestion
 * GRN_TOKEN_FORCE_PREFIX means that the token is used common prefix search
```

HashidaTKS (approved):

include/groonga/token.h
kou: Good catch!

PR #1194, opened 2021-05-25
grn_token_status: use uint32_t instead of unsigned int

kou:
{,u}int{8,16,32,64}_t are suitable raw integer types such as int and
unsigned int. Because they aren't depend on build environment. long
may be 32-bit integer or 64-bit integer.


kou:
Cc: @HashidaTKS

HashidaTKS (approved):

PR #1195, opened 2021-05-25
TokenFilterStem: add support for searching by the original token conditionally

kou:
Users can search by the original token (not stemmed token) conditionally with this.

It's useful when users want to search by a stemmed word generally but users sometimes want to search by a exact (not stemmed) word. For example:

* Searching by a stemmed word returns many results.
* Stemming returns wrong result.
* Users want to find only records that have a exact (not stemmed) word.

### DIFF
--- a/test/command/suite/token_filters/stem/keep_original/add.test
+++ b/test/command/suite/token_filters/stem/keep_original/add.test
@@ -1,0 +1,14 @@
+#@on-error omit
+plugin_register token_filters/stem
+#@on-error default
+
+table_create Memos TABLE_NO_KEY
+column_create Memos content COLUMN_SCALAR ShortText
+
+table_create Terms TABLE_PAT_KEY ShortText \
+  --default_tokenizer TokenBigram \
+  --normalizer NormalizerAuto \
+  --token_filters 'TokenFilterStem("keep_original", true)'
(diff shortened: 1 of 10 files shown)


kou:
Cc: @HashidaTKS

komainu8 (approved):
+1

test/command/suite/token_filters/stem/keep_original/add.test
```diff
@@ -0,0 +1,14 @@
+#@on-error omit
+plugin_register token_filters/stem
+#@on-error default
+
+table_create Memos TABLE_NO_KEY
+column_create Memos content COLUMN_SCALAR ShortText
+
+table_create Terms TABLE_PAT_KEY ShortText \
+  --default_tokenizer TokenBigram \
```
HashidaTKS, line 9: I wondered why the stemming works when the tokenizer is Bigram, but I noticed that `Tokenbigram` uses white-space-separate for ascii when a normalizer is used. 
 (https://groonga.org/docs/reference/tokenizers/token_bigram.html)
This is the reason why it works, right?

HashidaTKS (approved):
LGTM

test/command/suite/token_filters/stem/keep_original/add.test
kou: Right.

PR #1197, opened 2021-05-27
ci: use old red-arrow until the packages of Apache Arrow 4.0.1 are provided

komainu8:
This is testing now.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -108,8 +108,8 @@ jobs:
           sudo gem install \
             groonga-client \
             json \
-            pkg-config \
-            red-arrow
+            pkg-config
+          sudo gem install red-arrow -v 4.0.0
       - name: Set environment variables
         run: |
           echo "COLUMNS=79" >> ${GITHUB_ENV}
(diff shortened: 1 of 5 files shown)


.github/workflows/linux.yml
```diff
@@ -108,8 +108,8 @@ jobs:
           sudo gem install \
             groonga-client \
             json \
-            pkg-config \
-            red-arrow
+            pkg-config
+          sudo gem install red-arrow -v 4.0.0
```
kou, line 112: We don't need this now.

packages/apt/test.sh
```diff
@@ -54,7 +54,7 @@ gem install grntest
 if groonga --version | grep -q apache-arrow; then
   apt install -V -y \
     g++
-  gem install red-arrow
+  gem install red-arrow -v 4.0.0
```
kou, line 57: We don't need this now.

HashidaTKS:
Just asking, are these version specification removed if the packages of Apache Arrow 4.0.1 are provided?

kou:
Yes.
We'll be able to avoid this by improving Apache Arrow's release process. (Uploading Ruby packages after APT, Yum, MSYS2 and Homebrew packages are updated.)

HashidaTKS (approved):

kou:
Apache Arrow packages are updated. We don't need this.

PR #1198, opened 2021-05-28
hash: use GRN_DEFINE_NAME() instead of grn_hash_name()

kou:
GRN_DEFINE_NAME() is a more generic definition than grn_hash_name().


kou:
Cc: @HashidaTKS

HashidaTKS (approved):

PR #1201, opened 2021-05-31
grn_table_get_score: improve getting _score performance

kou:
This reduces needless parent _score check for one level
search (searching from a persistent table not a result set
table). Persistent tables should not have the GRN_OBJ_WITH_SUBREC
flag. Result set tables must have the flag.


kou:
Cc: @HashidaTKS

HashidaTKS (approved):

PR #1202, opened 2021-06-01
test: rename increment/decrement to increase/decrease

kou:
I thought that INC/DEC in GRN_QUERY_ADJ_INC/DEC mean
increment/decrement but I think that they mean increase/decrease now.


kou:
Cc: @HashidaTKS

HashidaTKS (approved):

PR #1203, opened 2021-06-01
Add a missing space between "." and "{"

kou:


kou:
Cc: @HashidaTKS

HashidaTKS:
I'm not familiar with Lemon Parser Generator, but this fix is looks good.

HashidaTKS (approved):

PR #1204, opened 2021-06-01
script syntax: enable "*>", "*<" and "*~" logical operators

kou:
grn_table_selector: add support for weight factor

Weight factor is a factor that is used when logical operator is
processed. The default weight factor is 1.0. It doesn't change weight
of each condition. If weight factor is 2.0, all weights are doubled.

Weight factor is associated with logical
operators (GRN_OP_AND/OR/AND_NOT/ADJUST) but there is no syntax that
uses weight factor. The following commits add it.

New APIs:

  * grn_selector_data_get_weight_factor()
  * grn_table_selector_get_weight_factor()
  * grn_table_selector_set_weight_factor()

"*>", "*<" and "*~" logical operators weren't implemented yet. They
always coursed syntax error.

"*>": Use GRN_OP_AND with weight factor 2.0

    A && B -> 2.0
      A: 1.0
      B: 1.0

    A *> B -> 3.0
      A: 1.0
      B: 2.0

"*>${WEIGHT_FACTOR}": Use GRN_OP_AND with custom weight factor

    A *>3.5 B -> 4.5
      A: 1.0
      B: 3.5

"*<": Use GRN_OP_AND with weight factor 0.5

    A && B -> 2.0
      A: 1.0
      B: 1.0

    A *< B -> 1.5
      A: 1.0
      B: 0.5

"*<${WEIGHT_FACTOR}": Use GRN_OP_AND with custom weight factor.
                      "-${WEIGHT_FACTOR}" is used as the custom weight factor.

    A *<0.5 B -> 0.5
      A: 1.0
      B: -0.5

"*~": Use GRN_OP_ADJUST with weight factor -1.
      GRN_OP_ADJUST doesn't change result set. It just adjusts score of
      records in the result set.

    A *~ B -> 0
      A: 1.0
      B: -1.0

"*~${WEIGHT_FACTOR}": Use GRN_OP_ADJUST with custom weight factor.
                      "-${WEIGHT_FACTOR}" is used as the custom weight factor.

    A *~0.5 B -> 0.5
      A: 1.0
      B: -0.5

query: add support for weight factor


kou:
Cc: @HashidaTKS

lib/expr.c
```diff
@@ -1182,13 +1181,11 @@ grn_expr_append_obj(grn_ctx *ctx, grn_obj *expr, grn_obj *obj, grn_operator op,
     case GRN_OP_AND :
     case GRN_OP_OR :
     case GRN_OP_AND_NOT :
+    case GRN_OP_ADJUST :
       PUSH_CODE(e, op, obj, nargs, code);
       if (nargs != 2) {
         GRN_LOG(ctx, GRN_LOG_WARNING, "nargs(%d) != 2 in relative op", nargs);
       }
-      if (obj) {
-        GRN_LOG(ctx, GRN_LOG_WARNING, "obj assigned to relative op");
```
HashidaTKS, line 1190: Why we can remove this?
Is `if (obj)` never true?

HashidaTKS:
> ```
> A *< B -> 1.5
>  A: 1.0
>  B: 1.5
> ```

Sorry to ask about such minor details, I guess `B` is not 1.5 but 0.5.

lib/db.c
```diff
@@ -4102,6 +4102,7 @@ grn_obj_search(grn_ctx *ctx, grn_obj *obj, grn_obj *query,
 
 typedef struct {
   grn_operator op;
+  float weight_factor;
```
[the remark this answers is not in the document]
kou: Right.

lib/expr.c
kou: We use the `obj` to pass weight factor.

kou:
Good catch!

HashidaTKS (approved):

lib/expr.c
HashidaTKS: I see.

PR #1205, opened 2021-06-01
Initialize grn_search_optarg::query_options

naoa:
The newly added ``grn_search_optarg :: query_options`` pointer may not be initialized with ``NULL`` and may try to parse to invalid addresses. 

https://github.com/groonga/groonga/blob/d883cd887b35947e16a7073b1fd59e685b45c862/plugins/token_filters/stop_word.c#L160-L162

### DIFF
--- a/lib/proc.c
+++ b/lib/proc.c
@@ -3891,6 +3891,7 @@ selector_in_values(grn_ctx *ctx, grn_obj *table, grn_obj *index,
     search_options.proc = NULL;
     search_options.max_size = 0;
     search_options.scorer = NULL;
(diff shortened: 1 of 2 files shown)


kou:
Thanks.

I merge this but I don't thin this is needed because we initialize these variables by `memset()`.

kou:
I think that 26da550 fixes this.

naoa:
Thanks!
I couldn't prepare a case that can be easily reproduced only with Groonga, but there was a case where grn_tokenizer_query_get_options did not become NULL when 6 token filters were set.

PR #1207, opened 2021-06-02
log-path, query-log-path: add support for outputting to stdout/stderr

kou:
We can specify "-" as --log-path/--query-log-path to output stdout and
"+" to output stderr.

This will be useful to run Groonga in Docker.

### DIFF
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -1,6 +1,6 @@
 /*
-  Copyright(C) 2009-2017 Brazil
-  Copyright(C) 2018-2019 Kouhei Sutou <kou@clear-code.com>
+  Copyright(C) 2009-2017  Brazil
+  Copyright(C) 2018-2021  Sutou Kouhei <kou@clear-code.com>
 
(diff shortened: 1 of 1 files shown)


kou:
Cc: @HashidaTKS

lib/logger.c
```diff
@@ -354,40 +420,40 @@ void
 grn_default_logger_set_path(const char *path)
 {
   if (logger_inited) {
```
HashidaTKS, line 422: What referring below is out of scope this PR.

I guess it potentially has a problem if  `grn_default_logger_set_path` and `grn_logger_init` and `grn_logger_fin`  are called concurrently because `logger_inited` can be replaced during execution of `grn_default_logger_set_path`.
However, it seems that these functions are never called concurrently and this problem never surface...

HashidaTKS (approved):

lib/logger.c
kou: It may be occurred but we don't support the situation. Initialization must be done from single thread.

PR #1214, opened 2021-06-18
NormalizerTable: add

kou:
See the added tests for how to use.

It uses Groonga table to normalize text. Users can implement custom
normalization by managing a Groonga table.

Note that users need to rebuild the index that uses NormalizerTable
when they change substitution table for the NormalizerTable. Because
normalization rules are changed by change of the substitution table.

### DIFF
--- a/lib/normalizer.c
+++ b/lib/normalizer.c
@@ -2478,12 +2478,18 @@ table_options_init(grn_ctx *ctx, table_options *options)
 static void
 table_options_fin(grn_ctx *ctx, table_options *options)
 {
+  /* We can't call grn_obj_unlink() safety here. Options are cached
+   * and closed when the table that have this option is closed. It may
+   * be occurred when DB is closed. In the case, options->column or
+   * options->table are already closed. If we call grn_obj_unlink()
+   * with closed column/table, Groonga may be crashed. */
+
   if (options->column) {
-    grn_obj_unlink(ctx, options->column);
+    /* grn_obj_unlink(ctx, options->column); */
     options->column = NULL;
   }
   if (options->table) {
-    grn_obj_unlink(ctx, options->table);
+    /* grn_obj_unlink(ctx, options->table); */
     options->table = NULL;
   }
 }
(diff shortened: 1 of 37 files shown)


kou:
Cc: @HashidaTKS

test/command/suite/normalizers/table/checks.expected
```diff
@@ -0,0 +1,37 @@
...
+    0,
+    0.0,
+    0.0
+  ],
+  {
+    "normalized": ".<あ>。1.",
+    "types": [
+
+    ],
+    "checks": [
```
HashidaTKS, line 23: What does `checks` represent?
I couldn't fully understand it from the code...

lib/normalizer.c
```diff
@@ -2457,6 +2457,578 @@ nfkc130_next(grn_ctx *ctx, int nargs, grn_obj **args, grn_user_data *user_data)
...
+                                             source_current,
+                                             source_end);
+      if (source_char_length == 0) {
+        ERR(GRN_INVALID_ARGUMENT,
+            "%s invalid character in source text: <%.*s>",
+            data->tag,
+            (int)(source_end - source_current),
+            source_current);
+        return;
+        }
```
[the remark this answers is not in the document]
kou: Good catch!

test/command/suite/normalizers/table/checks.expected
kou: It has the number of normalized bytes elements. 1(.) + 1(<) + 3(あ) + 1(>) + 3(。) + 1(1) + 1(.) = 11 in this case.

Each element represents the source character size in bytes.

If there are more normalized characters than source characters, elements of the rest characters are `-1`. `あ` is normalized to `<あ>` in this case. `checks[2]` is for (normalized) `あ` and `checks[5]` is for `>`. They are `-1` because the source character is only `あ`.

If normalized character is a multibyte character, the element of the first byte has the source character size in bytes. The rest elements are `0`. `checks[2]`, `checks[3]` and `checks[4]` are for (normalized) `あ`. `checks[3]` and `checks[4]` are `-1` because they are not for the first byte.

HashidaTKS: I see, Thank you!

> `checks[3]` and `checks[4]` are `-1` because they are not for the first byte.
`-1` is just typo of `0`, right?

kou: Ah, right.

PR #1225, opened 2021-07-05
NormalizerTable: add support for indexed target

kou:
Without this change, the following table is only supported:

    table_create Normalizations TABLE_PAT_KEY ShortText
    column_create Normalizations normalized COLUMN_SCALAR ShortText

With this change, the following table with an index is also supported:

    table_create Normalizations TABLE_NO_KEY
    column_create Normalizations target COLUMN_SCALAR ShortText
    column_create Normalizations normalized COLUMN_SCALAR ShortText

    table_create Terms TABLE_PAT_KEY ShortText
    column_create Terms normalizations_target COLUMN_INDEX \
      Normalizations target

The former can be used with:

    NormalizerTable("normalized", Normalizations.normalized")

The latter can be used with:

    NormalizerTable("normalized", "Normalizations.normalized",
                    "target", "Normalizations.target")

Note that "column" option is deprecated. Use "normalized" option
instead.

### DIFF
--- a/include/groonga/raw_string.h
+++ b/include/groonga/raw_string.h
@@ -74,6 +74,10 @@ GRN_API bool
 grn_raw_string_have_sub_string(grn_ctx *ctx,
                                grn_raw_string *string,
                                grn_raw_string *sub_string);
+GRN_API bool
+grn_raw_string_have_sub_string_cstring(grn_ctx *ctx,
+                                       grn_raw_string *string,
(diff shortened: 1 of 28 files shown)


kou:
Cc: @HashidaTKS

kou:
I merge this because I want to use this with PGroonga soon.
But you can still comment on this.

HashidaTKS (commented):
LGTM

PR #1226, opened 2021-07-06
ii offline: fix a bug that token is normalized twice

kou:


kou:
Cc: @HashidaTKS

lib/ii.c
```diff
@@ -15211,7 +15213,42 @@ grn_ii_builder_flush_term(grn_ctx *ctx, grn_ii_builder *builder,
...
+      break;
+    case GRN_TABLE_DAT_KEY :
+      GRN_TABLE_LOCK_BEGIN(ctx, builder->ii->lexicon) {
+        global_tid = grn_dat_add(ctx,
+                                 (grn_dat *)(builder->ii->lexicon),
+                                 key,
+                                 key_size,
+                                 NULL,
+                                 NULL);
+      } GRN_TABLE_LOCK_END(ctx, table);
```
HashidaTKS, line 15236: Isn't it `builder->ii->lexicon` instead of `table`?

```suggestion
      } GRN_TABLE_LOCK_END(ctx, builder->ii->lexicon);
```

kou: Good catch!

OK. I remove the second argument from `GRN_TABLE_LOCK_END()` because it's an internal macro. We can restore it without braking compatibility when we need it.

PR #1231, opened 2021-07-11
doc reference: add document about new arguments for between and in values

komainu8:


doc/source/reference/functions/between.rst
```diff
@@ -15,10 +15,12 @@ It is often used in combination with :ref:`select-filter` option in :doc:`/refer
 Syntax
 ------
 
-``between`` has three or five parameters::
+``between`` has some arguments from three to six parameters::
```
kou, line 18: Why do you mix "arguments" and "parameters"? We should not mix "arguments" and "parameters" without particular reason.

doc/source/reference/functions/between.rst
```diff
@@ -61,6 +63,12 @@ If it doesn't match the specified range, the ``select`` returns no records becau
 In the above case, it returns all the records because 14 exists in between 13 and 16.
 This behavior is used for checking the specified value exists or not in the table.
 
+How to set options of ``between`` as below.
```
kou, line 66:
```suggestion
Here is a query to ...
```

PR #1234, opened 2021-07-14
grndb recover: close used objects immediately

komainu8:
We can reduce memory usage by this.
This may decrease performance but it will be acceptable.

Note that "grndb check" doesn't close used objects immediately yet.

### DIFF
--- a/lib/mrb/mrb_ctx.c
+++ b/lib/mrb/mrb_ctx.c
@@ -832,6 +832,28 @@ grn_mrb_ctx_to_exception(mrb_state *mrb)
   return mrb_exc_new_str(mrb, error_class, mrb_str_new_cstr(mrb, message));
 }
 
+mrb_value
+ctx_push_temporary_open_space(mrb_state *mrb, mrb_value self)
+{
+  grn_ctx *ctx = (grn_ctx *)mrb->ud;
+
+  grn_ctx_push_temporary_open_space(ctx);
+  grn_mrb_ctx_check(mrb);
+
+  return mrb_nil_value();
+}
+
+mrb_value
+ctx_pop_temporary_open_space(mrb_state *mrb, mrb_value self)
+{
+  grn_ctx *ctx = (grn_ctx *)mrb->ud;
+
+  grn_ctx_pop_temporary_open_space(ctx);
+  grn_mrb_ctx_check(mrb);
+
+  return mrb_nil_value();
+}
+
 void
 grn_mrb_ctx_check(mrb_state *mrb)
 {
@@ -903,5 +925,11 @@ grn_mrb_ctx_init(grn_ctx *ctx)
 
   mrb_define_method(mrb, klass, "opened?", ctx_is_opened,
                     MRB_ARGS_REQ(1));
+
+  mrb_define_method(mrb, klass, "push_temporary_open_space",
+                    ctx_push_temporary_open_space, MRB_ARGS_NONE());
+  mrb_define_method(mrb, klass, "pop_temporary_open_space",
+                    ctx_pop_temporary_open_space, MRB_ARGS_NONE());
+
 }
 #endif
--- a/lib/mrb/mrb_ctx.c
+++ b/lib/mrb/mrb_ctx.c
@@ -309,6 +309,28 @@ ctx_is_opened(mrb_state *mrb, mrb_value self)
   return mrb_bool_value(grn_ctx_is_opened(ctx, mrb_id));
 }
 
+static mrb_value
+ctx_push_temporary_open_space(mrb_state *mrb, mrb_value self)
+{
+  grn_ctx *ctx = (grn_ctx *)mrb->ud;
+
+  grn_ctx_push_temporary_open_space(ctx);
+  grn_mrb_ctx_check(mrb);
+
+  return mrb_nil_value();
+}
+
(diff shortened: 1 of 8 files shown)


kou:
This approach will not work for objects that refer other objects such as an index column. An index column refers its lexicon.

We should use `grn_ctx_push_temporary_open_space()`/`grn_ctx_pop_temporary_open_space()` like `grn_db_recover()` does.

lib/mrb/mrb_object.c
```diff
@@ -254,6 +254,32 @@ grn_mrb_object_is_closed(mrb_state *mrb, mrb_value self)
   return mrb_bool_value(object == NULL);
 }
 
+mrb_value
+grn_mrb_object_push_open_space(mrb_state *mrb, mrb_value self)
```
kou, line 258:
```suggestion
grn_mrb_object_push_temporary_open_space(mrb_state *mrb, mrb_value self)
```

lib/mrb/mrb_object.c
```diff
@@ -254,6 +254,32 @@ grn_mrb_object_is_closed(mrb_state *mrb, mrb_value self)
...
+{
+  grn_ctx *ctx = (grn_ctx *)mrb->ud;
+  grn_bool is_close_opened_object_mode = (grn_thread_get_limit() == 1);
```
kou, line 261: We should not do this in this function.
It should be done by caller.

lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -603,8 +603,13 @@ def recover
           logger.log(:info, "Recovering database: <#{@database.path}>")
           if @force_truncate
             @database.each do |object|
```
kou, line 605: How about the following API?

```suggestion
            @database.each(use_temporary_open_space: true) do |object|
```

lib/mrb/mrb_object.c
```diff
@@ -254,6 +254,32 @@ grn_mrb_object_is_closed(mrb_state *mrb, mrb_value self)
...
+
+  if (is_close_opened_object_mode) {
+    grn_ctx_push_temporary_open_space(ctx);
```
kou, line 264: We should call `grn_mrb_ctx_check(mrb);` after calling any Groonga API.

lib/mrb/scripts/database.rb
```diff
@@ -38,7 +38,11 @@ def each(options=nil)
       context = Context.instance
       each_raw(options) do |id, cursor|
         object = context[id]
-        yield(object) if object
+        if object
+          object.push_temporary_open_space if options[:use_temporary_open_space]
+          yield(object)
+          object.pop_temporary_open_space if options[:use_temporary_open_space]
+        end
```
kou, line 45:
```ruby
each_raw(options) do |id, cursor|
  if options[:use_temporary_open_space]
    context.push_temporary_open_space
    begin
      object = context[id]
      yield(object) if object
    ensure
      context.pop_temporary_open_space
    end
  else
    object = context[id]
    yield(object) if object
  end
end

File: lib/mrb/scripts/command_line/grndb.rb [hunk 6]
```diff
@@ -602,7 +602,7 @@ def initialize(output)
         def recover
           logger.log(:info, "Recovering database: <#{@database.path}>")
           if @force_truncate
-            @database.each do |object|
+            @database.each(use_temporary_open_space: close_opened_object_mode?) do |object|
```
kou, line 605: We can assume that `grndb` doesn't use multiple threads.

lib/mrb/mrb_ctx.c
```diff
@@ -832,6 +832,28 @@ grn_mrb_ctx_to_exception(mrb_state *mrb)
   return mrb_exc_new_str(mrb, error_class, mrb_str_new_cstr(mrb, message));
 }
 
+mrb_value
```
kou, line 835: This should not be exported:

```suggestion
static mrb_value
```

kou: They should be defined right after `ctx_is_opened()`.

kou:
Can you add tests for this case?

komainu8:
Tests of `grndb recover ----force-truncate` and `grndb recover --force-lock-clear` have already exist in https://github.com/groonga/groonga/blob/master/test/command_line/suite/grndb/test_recover.rb#L177-L205 and https://github.com/groonga/groonga/blob/master/test/command_line/suite/grndb/test_recover.rb#L358-L380.

Are these tests not enough?

kou:
If we can reuse existing tests for this case, the existing tests should be failed. Because this change should close more objects.
If the existing tests aren't failed, this case isn't covered by existing tests.

komainu8:
Groonga doesn't output a part of the log(`open existing file ...`) on Linux. Therefore, existing tests don't fail on Linux, but they fail on Windows.
Therefore, I modified expected values for Windows of existing tests.

kou:
`DUMP` log level dumps open/close logs on Linux.
BTW, why do you want to check open log not close log?
Does this change close objects immediately?

kou:
`grndb` wasn't ran in single thread mode.

kou:
I'll merge this because Linux test failures should be unrelated.
This changes only grndb.

PR #1235, opened 2021-07-15
query_expand: add support for weighted increment/decrement/negative

kou:


kou:
Cc: @HashidaTKS

kou:
I merge this because I want to use this soon.
But you can add comments if you want to do.

lib/expr.c
```diff
@@ -5788,16 +5788,23 @@ grn_expr_syntax_expand_query_terms(grn_ctx *ctx,
...
     case GRN_QUERY_PREFIX :
       GRN_TEXT_PUTC(ctx, expanded_query, *cur);
       cur++;
       break;
+    case GRN_QUERY_ADJ_INC :
+    case GRN_QUERY_ADJ_DEC :
+    case GRN_QUERY_ADJ_NEG :
+      {
+        char *end = (char *)query_end;
+        strtof(cur + 1, &end);
```
HashidaTKS, line 5803: Can we correctly treat words that start with a number?
For example, when the word is `1_Rroonga` and the operator is `>-0.2`, the query is `>0.21_Rroonga`, it seems that we treat it as the word is `_Rroonga` and the operator is `>0.21`.

Do we have the way to avoid it like below?

`>0.2'1_Rroonga'`
`>0.2 1_Rroonga`

kou: Good catch!
We have `>-0.2"1_Rroonga"` syntax.
We can add support for the syntax as a follow up task.

PR #1236, opened 2021-07-15
test: add tests for groonga-synonym-generate

kou:


kou:
Cc: @HashidaTKS

HashidaTKS:
The added tests seem to be failing, is that okey?
Sorry, I haven't investigated the reason yet.

kou:
Good catch!
They are failed because groonga-httpd (nginx) rejects the large `load`. I changed `grntest` to increase the acceptable max body size.

HashidaTKS (approved):

HashidaTKS:
Thanks for your explanation!

PR #1237, opened 2021-07-15
increment/decrement: add support for nested

kou:
Parent weights are added to the specified weight.

Negative is also supported but there is no meaningful syntax for
it. For example "~2(~3XXX)" is invalid. "~" requires one or more
preceding words. So nested negative is meaningless.


kou:
Cc: @HashidaTKS

HashidaTKS (approved):

PR #1238, opened 2021-07-16
query query_parallel_or: add support for specifying group of match_columns as vector

naoa:


kou:
I've added this feature to `query()` too.
I'll merge this once tests are passed.

naoa:
Thanks!

PR #1240, opened 2021-07-20
grn_table_selector: add support for section and weight in prefix search

naoa:


kou:
I've added weight vector support to `grn_ii_cursor` and `grn_index_cursor` and use the new APIs for this.

naoa:
Thanks!

PR #1241, opened 2021-07-20
grn_table_selector: add support weight for prefix search

naoa:


kou:
#1240 also resolves this.

PR #1245, opened 2021-07-26
function: add support for object literal in object literal

kou:


kou:
Cc: @HashidaTKS

kou:
I'll merge this because I want to use this soon.
You can add comments after I merge this.

PR #1250, opened 2021-07-29
tool: add a tool for updating bundled Apache Arrow

komainu8:


kou:
Can we use `Makefile` (and shell script) like existing `make update-mruby` and `make update-nginx`?

komainu8:
I see.
I'll remake by using `Makefile` and shell script.

PR #1252, opened 2021-07-29
wal: add support for recovering fixed size column

kou:


kou:
Cc: @HashidaTKS

kou:
I'll add more commits to this and merge this after we release 11.0.5.

PR #1254, opened 2021-08-05
cache: clear all keys on disable

kou:
GitHub: ref GH-1224

Reported by higchi. Thanks!!!


kou:
Cc: @HashidaTKS

kou:
This will be safe because we always lock both read and write access to the hash.
(We need a stress test for this.)

HashidaTKS (approved):

PR #1256, opened 2021-08-20
grn_ctx_at: add missing lock clear

komainu8:
If some threads open the same object around the same time,
threads except for a thread that executes the opening object
at first are waiting for opening the target object.

At this time, threads that are waited for an opening object
take locks, but these locks are not released.

Therefore, these locks remain until Groonga's process is
restarted in the above case.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -11162,8 +11162,10 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
                 "<%u>(<%.*s>)",
                 id,
                 name_size, name);
+            grn_db_value_unlock(ctx, id, vp);
             goto exit;
           }
+          grn_db_value_unlock(ctx, id, vp);
         }
         if (vp->ptr) {
           switch (vp->ptr->header.type) {
--- a/lib/db.c
+++ b/lib/db.c
@@ -11151,7 +11151,8 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
           vp->done = 1;
           GRN_FUTEX_WAKE(&vp->ptr);
         } else {
-          if (!grn_db_value_wait(ctx, id, vp)) {
+          bool success = grn_db_value_wait(ctx, id, vp);
+          if (!success) {
             const char *name;
             uint32_t name_size = 0;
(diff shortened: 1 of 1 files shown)


lib/db.c
```diff
@@ -11162,8 +11162,10 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
                 "<%u>(<%.*s>)",
                 id,
                 name_size, name);
+            grn_db_value_unlock(ctx, id, vp);
             goto exit;
           }
+          grn_db_value_unlock(ctx, id, vp);
```
kou, line 11168: How about the following flow to show we want to ensure unlocking?

```c
bool success = grn_db_value_wait(ctx, id, vp);
if (!success) {
  /* set error */
}
grn_db_value_unlock(ctx, id, vp); /* This sentence is always executed because there is no condition. */
if (!success) {
  goto exit;
}
```

kou: Or:

```c
bool success = grn_db_value_wait(ctx, id, vp);
uint32_t current_lock = vp->lock;
grn_db_value_unlock(ctx, id, vp); /* This sentence is always executed because there is no condition. */
if (!success) {
  /* set error */
  goto exit;
}
```

I've fixed it at https://github.com/groonga/groonga/pull/1256/commits/4154545269ac6986972f74114f28bfc81f94554d.

lib/db.c
```diff
@@ -11162,10 +11163,11 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
                 "<%u>(<%.*s>)",
                 id,
                 name_size, name);
-            grn_db_value_unlock(ctx, id, vp);
+          }
+          grn_db_value_unlock(ctx, id, vp); /* This sentence is always executed because there is no condition. */
```
kou, line 11167: The comment is needless.
It's just for explaining this to you.

komainu8: Oh...
I've removed it.

kou:
https://github.com/groonga/groonga/runs/3380075216?check_suite_focus=true
This breaks reference count mode. We should not unlock when reference count mode and succeeded case.

PR #1257, opened 2021-08-21
grn_ctx_at: don't unlock when reference count mode and succeeded case

komainu8:
Because if reference count mode and succeeded case, vp->lock unlock
when grn_obj_unlink().

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -11152,8 +11152,7 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
           vp->done = 1;
           GRN_FUTEX_WAKE(&vp->ptr);
         } else {
-          bool success = grn_db_value_wait(ctx, id, vp);
-          if (!success) {
+          if (!grn_db_value_wait(ctx, id, vp)) {
             const char *name;
             uint32_t name_size = 0;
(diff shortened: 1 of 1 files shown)


lib/db.c
```diff
@@ -11164,9 +11165,16 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
...
+          if (grn_enable_reference_count) {
+            if (!success) {
+              grn_db_value_unlock(ctx, id, vp);
+              goto exit;
+            }
+          } else {
+            grn_db_value_unlock(ctx, id, vp);
+            if (!success) {
+              goto exit;
+            }
```
kou, line 11177: This logic is difficult to understand. How about the following flow?

```c
if (!grn_db_value_wait(ctx, id, vp)) {
  /* set error */
  grn_db_value_unlock(ctx, id, vp); /* Always unlock on error */
  goto exit;
}
if (!grn_enable_reference_count) {
  grn_db_value_unlock(ctx, id, vp); /* Unlock on non reference count mode. */
}
```

I've fixed at https://github.com/groonga/groonga/pull/1257/commits/37f884893e281ce777e5651de711f58c94be5959.

PR #1258, opened 2021-08-24
grn_obj_flush_only_dependent: improve performance

komainu8:
TODO:
Add a reason for the change into the commit message.
Test.


kou:
This is not acceptable.
Lock is needed for safe WAL clear.

komainu8:
This problem has already resolved at 0ca71c1df8cb6828eeb3c12c3361e0d02f62a8ba.
Therefore, this is close.

PR #1260, opened 2021-08-30
ci: add missing trigger patterns

komainu8:
Because CI doesn't execute even if we push tags or maintenance branch currently.


kou:
Oh, sorry.

PR #1264, opened 2021-09-07
grn_io_seg_map_: fix a resource leak

komainu8:
Currently, If DO_MAP fails, opened fileinfo is not close.
Because unlink does not release file descriptor.

Therefore, If DO_MAP fails, the file descriptor is not released until
Groonga restarts.


kou:
Good catch!

PR #1263, opened 2021-09-02
arrow: use dictionary array for reference text column

HashidaTKS:
## OverView

Make `StreamWriter` write string type values as string type `DictionaryArray`.

Writing `DictionaryArray` on `FileDumper` is not supported because `arrow::FileWriter` does not support dictionary replacement. For writing `DictionaryArray` on `FileDumper`, we should create a single dictionary array for a filed including the case of using delta dictionaries, but it may cause memory exhaustion or exceeding a `ArrowArray` length limitation and so on when the dictionary is large.  
https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format
https://arrow.apache.org/docs/format/Columnar.html#array-lengths

## Behavior

`StreamWriter` writes string type values as string type `DictionaryArray`.
Other type values are written as before because changing to write these types as `DictionaryArray`  does not improve performance/data size.
In order to avoid memory exhaustion or exceeding a `ArrowArray` length limitation and so on when the dictionary becames large, `DictionaryBuilder::ResetFull()` is executed when the dictionary length exceeds a certain size (10000). This size is not set on a special basis. It is set on the assumption that ten times the record batch length (1000) is sufficient.

### Test data

I have created a test file like a dump result (`testdata.grn`) from the attached python file.
(I wanted to attach a compressed dump file but it was over the github size limitation...)
[create_test_data.zip](https://github.com/groonga/groonga/files/7147965/create_test_data.zip)

I have compared performance before and after change.

Table records: 1,000,000
`high_cardinality` column values:  Generated based on random values from 0 to 500,000
`high_cardinality_vector` column values:  Generated based on random values from 0 to 500,000
`middle_cardinality` column values:  Generated based on random values from 0 to 50,000
`middle_cardinality_vector` column values:  Generated based on random values from 0 to 50,000
`low_cardinality` column values:  Generated based on random values from 0 to 5,000
`low_cardinality_vector` column values:  Generated based on random values from 0 to 5,000

Each column is a reference column.

### Result

<details>
<summary>Before log</summary>

```
2021-09-11 21:19:11.791000|000000FCF11FFCD0|>/d/select?table=Logs&command_version=3&output_type=apache-arrow&limit=-1&output_columns=high_cardinality&cache=no
2021-09-11 21:19:11.791000|000000FCF11FFCD0|:000000000000000 select(1000000)
2021-09-11 21:19:12.086000|000000FCF11FFCD0|:000000295000000 output(1000000)
[...]

PR #1265, opened 2021-09-10
Update required apache arrow version to 3.0.0

HashidaTKS:
Update required apache arrow version to 3.0.0 in order to avoid complicating the source code.
Checking arrow version and branching by it are not so complicated for now, but they can be very complicated in the future.


HashidaTKS:
Related: https://github.com/groonga/groonga/pull/1263#discussion_r704744386

HashidaTKS:
Cc: @kou 

Could you please review this?

kou:
Thanks!

PR #1269, opened 2021-09-30
snip: add support for regular expression to find delimiter

kou:
Use case: Use only text in one sentence as snippet by finding "." (for
English) or U+3002 IDEOGRAPHIC FULL STOP (for Japanese).

Note that found delimiters aren't included in snippet. For example,
"." isn't included in snippet with /\./ regular expression. This is
useful for the use case.

### DIFF
--- a/test/command/suite/select/function/snippet/delimiter_regexp/overlap.expected
+++ b/test/command/suite/select/function/snippet/delimiter_regexp/overlap.expected
@@ -1,0 +1,36 @@
+table_create Entries TABLE_NO_KEY
+[[0,0.0,0.0],true]
+column_create Entries content COLUMN_SCALAR ShortText
+[[0,0.0,0.0],true]
+load --table Entries
+[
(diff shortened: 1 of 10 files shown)


kou:
Cc: @HashidaTKS

test/command/suite/select/function/snippet/delimiter_regexp/overlap.expected
```diff
@@ -0,0 +1,35 @@
...
+      ],
+      [
+        [
+          "snippet",
+          null
+        ]
+      ],
+      [
+        [
+          " bbb [ccc] AAA. BBB [CCC] xxx yyy"
```
HashidaTKS, line 30: Delimiters included in the range of  `match_width` do not seem to be recognized as a delimiter, is this intended behavior?

kou: Good catch!
It's intentional for me.
But the user who has this use case wants to be delimited. So I've changed the behavior.

PR #1270, opened 2021-09-30
snippet: add support for cache

kou:
New API: grn_proc_func_generate_cache_key()

We can use the API to support caching in other function.

This commits also add support for extracting keywords from the current
condition like snippet_html(). The current condition is used only when

  1. default_open_tag option is specified
  2. default_close_tag option is specified
  3. No keywords are specified


kou:
Cc: @HashidaTKS

kou:
This uses #1269 as the parent branch. So we should merge #1269 before this.

lib/proc.c
```diff
@@ -477,6 +477,51 @@ grn_proc_options_parsev(grn_ctx *ctx,
...
+      unsigned int key_size;
+      void *value;
+      int value_size = grn_hash_cursor_get_key_value(ctx,
+                                                     cursor,
+                                                     &key,
+                                                     &key_size,
+                                                     &value);
+      GRN_TEXT_PUT(ctx, cache_key, key, key_size);
+      GRN_TEXT_PUTC(ctx, cache_key, '\1');
+      if (value_size == sizeof(grn_obj *)) {
```
HashidaTKS, line 498: I'm not confident but isn't it `grn_obj` , not `grn_obj *` ?

lib/proc/proc_snippet.c
```diff
@@ -89,85 +89,121 @@ snippet_exec(grn_ctx *ctx, grn_obj *snip, grn_obj *text,
...
+  GRN_TEXT_PUTC(ctx, &cache_key, '\0');
+  grn_proc_func_generate_cache_key(ctx, args + 1, nargs - 1, &cache_key);
+
+  grn_obj *snip_ptr = NULL;
+  grn_obj *expression;
+  grn_proc_get_info(ctx, user_data, NULL, NULL, &expression);
+  snip_ptr = grn_expr_get_var(ctx,
+                              expression,
+                              GRN_TEXT_VALUE(&cache_key),
+                                GRN_TEXT_LEN(&cache_key));
```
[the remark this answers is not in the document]
kou: Good catch!

lib/proc.c
kou: Good catch! You're right!

PR #1271, opened 2021-09-30
snippet: document

kou:
It's not cover all features.


kou:
This uses #1270 as the parent branch. So we should merge #1270 before this.

doc/source/reference/functions/snippet.rst
```diff
@@ -0,0 +1,259 @@
...
+..   --query "fast performance"
+
+It returns only one snippet because ``"max_n_results": 1`` is specified.
+
+The max size of a text part is 200byte by default. The unit is bytes
+not characters. The size doesn't include inserted ``[`` and ``[``. You
+can change it by ``width`` option:
+
+.. groonga-command
+.. include:: ../../example/reference/functions/snippet/usage_max_n_results.log
```
HashidaTKS, line 140: Is this example file correct?
Also, it seems that the `width` option isn't used at this usage in the built html.

doc/source/reference/functions/snippet.rst
```diff
@@ -0,0 +1,259 @@
...
+..                             })' \
+..   --match_columns content \
+..   --query "fast performance"
+
+You can detect snippet delimiter with regular expression by
+``delimiter_regexp`` option. You can use ``\.\s*`` to use only text in
+the target sentence. Note that you need to escape ``\`` in string:
+
+.. groonga-command
+.. include:: ../../example/reference/functions/snippet/usage_max_n_results.log
```
HashidaTKS, line 156: Is this example file correct?

doc/source/reference/functions/snippet.rst
kou: Good catch!

doc/source/reference/functions/snippet.rst
kou: Good catch!

PR #1274, opened 2021-10-06
grn_obj_search: add support for auto cast in table search

kou:
Without this, in_values(_key, 10) doesn't work with UInt64 key
table. Because 10 is parsed as Int32.


kou:
Cc: @HashidaTKS

HashidaTKS (approved):

PR #1278, opened 2021-10-11
package: add a task for tagging and uploading source archive

komainu8:
I added a rake task for tagging and uploading the source archive.
We can set tag, make source archive and, upload source archive by the following command.

```
rake archive:tag
```


kou:
https://github.com/groonga/groonga/commit/c63a7c23563c6e44aca2f60fda09520de95b2ea4 build and upload source archive but it doesn't support uploading to packages.groonga.org yet.

PR #1284, opened 2021-10-28
load: add support for ISO 8601 time format

HashidaTKS:
**Specification**

Only support the extended date and time format like below.
(We cannot omit `-` or `/`.)

https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

* YYYY-MM-ddThh:mm:ss.sZ
* YYYY-MM-ddThh:mm:ss.s+10:00
* YYYY-MM-ddThh:mm:ss.s-10:00

The "T" and "Z" characters in this syntax may alternatively be lower case "t" or "z" respectively.
( from https://datatracker.ietf.org/doc/html/rfc3339#section-5.6 )
The `-` character in this syntax may alternatively be `/` because the current date parser of Groonga support it.
When the time-offset is specified, the time is converted to UTC based on the time-offset.
When the time-offset is omitted, the time is regarded as local time and converted to UTC .


HashidaTKS:
This is still a work in progress.

* Test more
* Write a description for this PR

HashidaTKS:
**Specification**

Only support the extended date and time format like below.


* YYYY-MM-ddThh:mm:ss.sZ
* YYYY-MM-ddThh:mm:ss.s+10:00
* YYYY-MM-ddThh:mm:ss.s-10:00

The "T" and "Z" characters in this syntax may alternatively be lower case "t" or "z" respectively.
( from https://datatracker.ietf.org/doc/html/rfc3339#section-5.6 )
The `-` characters in this syntax may alternatively be `/` because the current date parser of groonga support it.
When the zone is omitted, the time is regarded as local time.

I didn't add tests for parsing of local time, including the original time format, because results of `load` depended on the zone of the test environment.
However I tested those cases manually and they worked well.

include/groonga/portability.h
```diff
@@ -204,8 +204,10 @@
 # define grn_time_t __time64_t
 # define grn_time(dest_time) _time64((dest_time))
 # define grn_mktime(tm) _mktime64((tm))
+# define grn_timegm(tm) _mkgmtime64((tm))
 #else /* WIN32 */
 # define grn_time_t time_t
 # define grn_time(dest_time) time((dest_time))
 # define grn_mktime(tm) mktime((tm))
+# define grn_timegm(tm) timegm((tm))
```
HashidaTKS, line 212: If we need to support environments that have no `timegm`, we should create our own `timegm` and use it in those environments.
I think we don't support those environments for now because the code below regard we always have `timegm` in the environments except Windows.

https://github.com/groonga/groonga/blob/master/plugins/functions/time.c#L33

HashidaTKS:
Cc: @kou 

Could you please review this when you have time?

HashidaTKS:
Ah, I will check failed tests.

kou:
These failures will be unrelated.

HashidaTKS:
These failures don't seem to be related to this PR changes, these seem to be caused by environment problems.

- Failing to fetch submodules
- Failing to setup mysy

Sorry but I haven't had much time to investigate, is there any idea to fix them?
Or can we ignore these failures at this PR?

kou:
Could you rebase on master?
Then they will be fixed.

HashidaTKS:
@kou 

Thank you!

I have rebased the branch, but still a check fails during building mecab...

By the way, could I re-run failed actions?
It seems that I have no permission.

komainu8:
I re-ran failed actions.

komainu8:
All tests have passed.
I'll review this PR in a few days.

HashidaTKS:
@komainu8
Thank you!

komainu8 (changes requested):
@HashidaTKS 
I've finished the review.
Could you confirm my comments?

HashidaTKS:
@komainu8 

Thanks for your review!

I have addressed your comments.

komainu8 (approved):
LGTM!

Some tests failed. However, these failures will be unrelated.
Could you rebase on master?
Then they will be fixed.

HashidaTKS:
Rebased on master

test/command/suite/load/time/cast/string/iso8601/normal.test
```diff
@@ -0,0 +1,20 @@
...
+{"created_at": "2000-01-01T08:59:00-01:01"},
+{"created_at": "1999-12-31T22:49:00-11:11"},
+{"created_at": "2000-01-01T11:00:00+01"},
+{"created_at": "2000-01-01T09:00:00-01"},
+{"created_at": "2000-01-01T11:01:00.123+01:01"},
+{"created_at": "2000-01-02T09:00:00+23:00"},
+{"created_at": "2000-01-01T10:59:00+00:59"},
+{"created_at": "2000/01/01T10:00:00Z"}
+]
+select Logs --limit -1
```
kou, line 20: How about showing also input text and formatting time value as ISO format string? It improves debugability.

```text
plugin_register functions/time

table_create Logs TABLE_NO_KEY
column_create Logs created_at COLUMN_SCALAR Time
column_create Logs created_at_text COLUMN_SCALAR ShortText

load --table Logs
[
{"created_at":      "2000-01-01T10:00:00Z",
 "created_at_text": "2000-01-01T10:00:00Z"},
...
]

select Logs --limit -1 --output_columns 'time_format_iso8601(created_at), created_at_text`

Comment (kou) [reply test/command/suite/load/time/cast/string/iso8601/normal.test:20]: How about `valid.test` and `invalid.test` for file names?

**Review by kou — COMMENTED** [review]
We force to use `Asia/Tokyo` timezone for test: https://github.com/groonga/groonga/blob/master/test/command/run-test.sh#L196

Could you update the pull request description with the specification in https://github.com/groonga/groonga/pull/1284#issuecomment-959479320 ?

File: test/command/suite/load/time/cast/string/iso8601/normal.test (hunk 2 above)
Comment (HashidaTKS) [reply test/command/suite/load/time/cast/string/iso8601/normal.test]: I have fixed it.

**Comment by HashidaTKS** [comment]
@kou 

Thank you for the review!

I have addressed your comments.
Please review again.

File: lib/time.c [hunk 3]
```diff
@@ -202,12 +202,58 @@ grn_timeval2str(grn_ctx *ctx, grn_timeval *tv, char *buf, size_t buf_size)
...
+  }
+
+  if ((position + 1) >= end) { return GRN_INVALID_ARGUMENT; }
+
+  position++;
+  hours = grn_atoi(position, end, &position);
+  if (hours < 0 || hours >= 24) { return GRN_INVALID_ARGUMENT; }
+
+  if ((position + 1) < end && *position == ':') {
+    position++;
```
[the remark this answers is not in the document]
kou: This is `*position == ':'` case. Is `2000-01-01T11:00:00+01:` (see the last `:`) a valid input?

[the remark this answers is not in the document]
kou: (`position > end` check is also needed with my suggestion...)

test/command/suite/load/time/cast/string/iso8601/valid.test
```diff
@@ -0,0 +1,26 @@
+plugin_register functions/time
+table_create Logs TABLE_NO_KEY
+column_create Logs created_at COLUMN_SCALAR Time
+column_create Logs created_at_text COLUMN_SCALAR ShortText
+
+
+
+load --table Logs
+[
+{"created_at": "2000-01-01T10:00:00Z","created_at_text": "2000-01-01T19:00:00.000000+09:00"},
```
kou, line 10: Could you specify the same text as the `created_at` to `created_at_text` instead of expected value to view input in the `.expected` file?

(And a space after `,` is missing.)

It may be better that we also have a label what is tested by this data:

```json
{"_key": "timezone: Z", "created_at": "2000-01-01T10:00:00Z", "created_at_text": "2000-01-01T10:00:00Z"}
```

lib/time.c
```diff
@@ -241,19 +287,38 @@ grn_str2timeval(const char *str, uint32_t str_len, grn_timeval *tv)
...
+    if (rc != GRN_SUCCESS) { return rc; }
+
+    if (have_offset) {
+      tv->tv_sec = grn_timegm(&tm);
+      tv->tv_sec -= gm_offset_sec;
+    } else {
+      tv->tv_sec = grn_mktime(&tm);
+      /* tm_yday is set appropriately (0-365) on successful completion. */
+      if (tm.tm_yday == -1) { return GRN_INVALID_ARGUMENT; }
+    }
```
kou, line 319: I think that we also need this for `grn_timegm()` case.

```suggestion
    }
    /* tm_yday is set appropriately (0-365) on successful completion. */
    if (tm.tm_yday == -1) { return GRN_INVALID_ARGUMENT; }
```

lib/time.c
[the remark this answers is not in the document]
HashidaTKS: Ah, sorry, I misunderstood. I have fixed it.

lib/time.c
HashidaTKS: I have fixed it.

test/command/suite/load/time/cast/string/iso8601/valid.test
HashidaTKS: Ah, I understand now. I have fixed it.

lib/time.c
```diff
@@ -202,12 +202,61 @@ grn_timeval2str(grn_ctx *ctx, grn_timeval *tv, char *buf, size_t buf_size)
...
+
+  position++;
+  if (position >= end) { return GRN_INVALID_ARGUMENT; }
+
+  hours = grn_atoi(position, end, &position);
+  if (hours < 0 || hours >= 24) { return GRN_INVALID_ARGUMENT; }
+
+  if (*position == ':') {
+    position++;
+    if(position == end) { return GRN_INVALID_ARGUMENT; }
```
HashidaTKS, line 241: `*end` can be `:`, so if checking `position` with `if(position >= end)`, this line returns `GRN_INVALID_ARGUMENT` even the argument is valid in some cases.
I mean the cases like below.
(This was occurred in `load/time/cast/string/iso8601/valid.test` in my environment)

input value
`2000-01-01T11:01:00.123+01:01`

actual memory
```
start                 end
*                     *
2000-01-01T11:01:00+01:
```

HashidaTKS (commented):
@kou 
Thanks!

I have addressed your second review comments.

lib/time.c
kou: Ah, sorry. I wanted to say it for `if (*position == ':') {` not `if (position == end)`.

HashidaTKS: Ah, I see. 
I have moved `position >= end` check into `if (*position == ':')`  for reducing nests.

lib/time.c
```diff
@@ -202,12 +202,60 @@ grn_timeval2str(grn_ctx *ctx, grn_timeval *tv, char *buf, size_t buf_size)
...
+
+  position++;
+  if (position >= end) { return GRN_INVALID_ARGUMENT; }
+
+  hours = grn_atoi(position, end, &position);
+  if (hours < 0 || hours >= 24) { return GRN_INVALID_ARGUMENT; }
+
+  if (position < end && *position == ':') {
+    position++;
+    if(position == end) { return GRN_INVALID_ARGUMENT; }
```
kou, line 241:
```suggestion
    if (position == end) { return GRN_INVALID_ARGUMENT; }
```

lib/time.c
```diff
@@ -241,19 +289,39 @@ grn_str2timeval(const char *str, uint32_t str_len, grn_timeval *tv)
...
+
+    if (rc != GRN_SUCCESS) { return rc; }
+
+    if (have_offset) {
+      tv->tv_sec = grn_timegm(&tm);
+      tv->tv_sec -= gm_offset_sec;
+    } else {
+      tv->tv_sec = grn_mktime(&tm);
+    }
+
```
kou, line 320:
```suggestion
```

test/command/suite/load/time/cast/string/iso8601/valid.expected
```diff
@@ -0,0 +1,127 @@
...
+{"case": "timezone: +11:11", "created_at": "2000-01-01T21:11:00+11:11", "created_at_text": "2000-01-01T21:11:00+11:11"},
+{"case": "timezone: -01:01", "created_at": "2000-01-01T08:59:00-01:01", "created_at_text": "2000-01-01T08:59:00-01:01"},
+{"case": "timezone: -11:11", "created_at": "1999-12-31T22:49:00-11:11", "created_at_text": "1999-12-31T22:49:00-11:11"},
+{"case": "timezone hour threshold: +23:00", "created_at": "2000-01-02T09:00:00+23:00", "created_at_text": "2000-01-02T09:00:00+23:00"},
+{"case": "timezone minute threshold: +00:59", "created_at": "2000-01-01T10:59:00+00:59", "created_at_text": "2000-01-01T10:59:00+00:59"},
+{"case": "timezone omitting minute: +01", "created_at": "2000-01-01T11:00:00+01", "created_at_text": "2000-01-01T11:00:00+01"},
+{"case": "timezone omitting minute: -01", "created_at": "2000-01-01T09:00:00-01", "created_at_text": "2000-01-01T09:00:00-01"},
+{"case": "timezone: localtime", "created_at": "2000-01-01T19:00:00", "created_at_text": "2000-01-01T19:00:00"},
+{"case": "date separater: /", "created_at": "2000/01/01T10:00:00Z", "created_at_text": "2000/01/01T10:00:00Z"},
```
kou, line 25:
```suggestion
{"case": "compatible: date separater: /", "created_at": "2000/01/01T10:00:00Z", "created_at_text": "2000/01/01T10:00:00Z"},
```

test/command/suite/load/time/cast/string/iso8601/valid.test
```diff
@@ -0,0 +1,25 @@
...
+{"case": "timezone: +11:11", "created_at": "2000-01-01T21:11:00+11:11", "created_at_text": "2000-01-01T21:11:00+11:11"},
+{"case": "timezone: -01:01", "created_at": "2000-01-01T08:59:00-01:01", "created_at_text": "2000-01-01T08:59:00-01:01"},
+{"case": "timezone: -11:11", "created_at": "1999-12-31T22:49:00-11:11", "created_at_text": "1999-12-31T22:49:00-11:11"},
+{"case": "timezone hour threshold: +23:00", "created_at": "2000-01-02T09:00:00+23:00", "created_at_text": "2000-01-02T09:00:00+23:00"},
+{"case": "timezone minute threshold: +00:59", "created_at": "2000-01-01T10:59:00+00:59", "created_at_text": "2000-01-01T10:59:00+00:59"},
+{"case": "timezone omitting minute: +01", "created_at": "2000-01-01T11:00:00+01", "created_at_text": "2000-01-01T11:00:00+01"},
+{"case": "timezone omitting minute: -01", "created_at": "2000-01-01T09:00:00-01", "created_at_text": "2000-01-01T09:00:00-01"},
+{"case": "timezone: localtime", "created_at": "2000-01-01T19:00:00", "created_at_text": "2000-01-01T19:00:00"},
+{"case": "date separater: /", "created_at": "2000/01/01T10:00:00Z", "created_at_text": "2000/01/01T10:00:00Z"},
```
kou, line 21:
```suggestion
{"case": "compatible: date delimiter: /", "created_at": "2000/01/01T10:00:00Z", "created_at_text": "2000/01/01T10:00:00Z"},
```

test/command/suite/load/time/cast/string/iso8601/valid.expected
```diff
@@ -0,0 +1,127 @@
...
+        "2000-01-01T19:00:00.000000+09:00",
+        "2000-01-01T09:00:00-01"
+      ],
+      [
+        "timezone: localtime",
+        "2000-01-01T19:00:00.000000+09:00",
+        "2000-01-01T19:00:00"
+      ],
+      [
+        "date separater: /",
```
kou, line 116:
```suggestion
        "compatible: date delimiter: /",
```

kou:
Thanks!

PR #1288, opened 2021-11-07
groonga admin: Fix advanced search to work properly

HashidaTKS:
#1186 

`$('#table-tab-recordlist-full-checkbox').attr('checked')` returns the initial state of the checkbox, does not changed with the current state of the checkbox.
We should use `prop()` instead of `attr()` for this checkbox.

https://api.jquery.com/attr/

> $( elem ).prop( "checked" ) | true (Boolean) Will change with checkbox state
> --|--
> $( elem ).attr( "checked" )             (1.6+) | "checked" (String) Initial state of the checkbox; does not change

### DIFF
--- a/data/html/admin/assets/groonga-admin.ja.js
+++ b/data/html/admin/assets/groonga-admin.ja.js
@@ -174,7 +174,7 @@ function GroongaAdmin() {
     that.removetable();
   });
   $('#tab-recordlist-form').submit(function() {
-    if ($('#table-tab-recordlist-full-checkbox').attr('checked')) {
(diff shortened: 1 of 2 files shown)


HashidaTKS:
https://github.com/groonga/groonga/blob/master/data/html/admin/assets/groonga-admin.js#L1119

The code above can have the same problem, but as far as I've confirmed, it seems to work correctly, so I have not modified it for now.
I think it is better to also change `attr('disabled', 'disabled')` to `prop('disable', true)` when we modify it.

HashidaTKS:
Cc: @kou, @komainu8 

Could you please review this when you have time?

kou:
Thanks!

PR #1290, opened 2021-11-22
expr: add GRN_EXPR_DISABLE_PREFIX_SEARCH

kou:
This is "DISABLE_PREFIX_SEARCH" in query_flags value.

This has an implementation for not only query syntax but also script
syntax but doesn't have a test for script syntax. Because there is no
API to specify flags for script syntax in the current Groonga
commands.


kou:
I merge this because I want this for the next release.
But anyone can review and leave comments after I merge this.

HashidaTKS (commented):
Sorry for the delay.
LGTM.

Issue #1300, opened 2021-12-09
Can't remove invalid parameter index column


kou:
```text
table_create Statuses TABLE_NO_KEY
column_create Statuses start_time COLUMN_SCALAR UInt16
column_create Statuses end_time COLUMN_SCALAR UInt16

table_create Times TABLE_PAT_KEY UInt16
column_create Times statuses COLUMN_INDEX Statuses start_time,end_time
# [[-22,1639037503.16114,0.003981828689575195,"grn_obj_set_info(): GRN_INFO_SOURCE: multi column index must be created with WITH_SECTION flag: <Times.statuses>",[["grn_obj_set_info_source_validate","../../groonga/lib/db.c",9605],["/tmp/d.grn",6,"column_create Times statuses COLUMN_INDEX Statuses start_time,end_time"]]],false]
table_remove Times
# [[-22,1639037503.16515,0.0005414485931396484,"[object][remove] column is broken: <Times.statuses>",[["remove_columns","../../groonga/lib/db.c",10649],["/tmp/d.grn",8,"table_remove Times"]]],false]
```

kou:
@HashidaTKS Do you want to work on this?

HashidaTKS:
Yes.
I will work on this at this weekend.
Perhaps I will take long time to fix this because I'm busy these days...

kou:
Thanks!
No problem. You don't need to hurry.

HashidaTKS:
Close because #1301 is merged.

PR #1301, opened 2021-12-10
Fix to not break table when column_create fails (for #1300)

HashidaTKS:
**Details of #1300**

Sample  queries
```
table_create Statuses TABLE_NO_KEY
column_create Statuses start_time COLUMN_SCALAR UInt16
column_create Statuses end_time COLUMN_SCALAR UInt16
 
table_create Times TABLE_PAT_KEY UInt16
column_create Times statuses COLUMN_INDEX Statuses start_time,end_time *1
# [[-22,1639037503.16114,0.003981828689575195,"grn_obj_set_info(): GRN_INFO_SOURCE: multi column index must be created with WITH_SECTION flag: <Times.statuses>",[["grn_obj_set_info_source_validate","../../groonga/lib/db.c",9605],["/tmp/d.grn",6,"column_create Times statuses COLUMN_INDEX Statuses start_time,end_time"]]],false]
table_remove Times *2
# [[-22,1639037503.16515,0.0005414485931396484,"[object][remove] column is broken: <Times.statuses>",[["remove_columns","../../groonga/lib/db.c",10649],["/tmp/d.grn",8,"table_remove Times"]]],false]
```

https://github.com/groonga/groonga/blob/master/lib/proc/proc_column.c#L280

`grn->rc` is `GRN_INVALID_ARGUMENT` at the line above when executing the `column_create` (*1), and the `grn_obj_remove` above reaches the line below.

https://github.com/groonga/groonga/blob/master/lib/db.c#L11309

`_grn_obj_remove_spec` returns a return code of `grn_ja_replace` in this case.
(https://github.com/groonga/groonga/blob/master/lib/store.c#L3204)

If `grn_ja_replace` returns the original `grn->rc` when it succeed, it returns `GRN_INVALID_ARGUMENT` in this case.
It means that `_grn_obj_remove_spec` returns `GRN_INVALID_ARGUMENT` and the `grn_obj_delete_by_id` below is skipped.
As a result, the information that should be deleted is not deleted and the column (table) is broken.
The `table_remove` (*2) fails because of this.

https://github.com/groonga/groonga/blob/master/lib/db.c#L11311

**Modification**

- Fix to temporarily reset `ctx->rc` in callers of `grn_obj_remove` in `column_create` and `column_create_similar`.
  - #1300 mentions only to `column_create` but  `column_create_similar` has same problem.
- Add schema and removability checks for invalid cases of `column_create` tests.
  - I felt verbose a little to add to the all cases, but I did so for consistency.

Tests for `column_create_similar` have not be changed because it is difficult to reproduce the problem from command line ( I mean `.test` file ).
I have tested the modification of `column_create_similar` by forcibly changing parameters with debugger.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -927,8 +927,6 @@ grn_column_create_similar_internal(grn_ctx *ctx,
                 "%s base column's source doesn't exist: %u",
                 tag,
                 base_source_column_id);
-            grn_obj_remove(ctx, column);
-            column = NULL;
             break;
           }
           char source_column_name[GRN_TABLE_MAX_KEY_SIZE];
@@ -951,8 +949,6 @@ grn_column_create_similar_internal(grn_ctx *ctx,
                 name,
                 source_column_name_length,
                 source_column_name);
-            grn_obj_remove(ctx, column);
-            column = NULL;
             break;
           }
           GRN_RECORD_PUT(ctx, &source_ids, DB_OBJ(source_column)->id);
@@ -965,7 +961,10 @@ grn_column_create_similar_internal(grn_ctx *ctx,
       }
       GRN_OBJ_FIN(ctx, &base_source_ids);
       if (ctx->rc != GRN_SUCCESS) {
+        grn_rc original_rc = ctx->rc;
+        ctx->rc = GRN_SUCCESS;
         grn_obj_remove(ctx, column);
+        ctx->rc = original_rc;
         return NULL;
       }
     }
--- a/lib/proc/proc_column.c
+++ b/lib/proc/proc_column.c
@@ -277,7 +277,10 @@ command_column_create(grn_ctx *ctx, int nargs, grn_obj **args,
     }
     GRN_OBJ_FIN(ctx, &source_ids);
     if (rc != GRN_SUCCESS) {
+      grn_rc original_rc = ctx->rc;
+      ctx->rc = GRN_SUCCESS;
       grn_obj_remove(ctx, column);
+      ctx->rc = original_rc;
       succeeded = GRN_FALSE;
       goto exit;
     }
--- a/lib/store.c
+++ b/lib/store.c
@@ -2398,8 +2398,6 @@ grn_ja_replace(grn_ctx *ctx, grn_ja *ja, grn_id id,
 {
   const char *tag = "[ja][replace]";
   grn_ja_wal_add_entry_data wal_data = {0};
-  grn_rc original_rc = ctx->rc;
-  ctx->rc = GRN_SUCCESS;
   wal_data.ja = ja;
   wal_data.need_lock = false;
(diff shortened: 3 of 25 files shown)


kou:
Why is the original `ctx->rc` no `GRN_SUCCESS` in this case?

HashidaTKS:
I'll write the details later.

kou:
Thanks.

HashidaTKS:
I have written the details (description).

kou:
OK. Then we should reset `ctx->rc` in `proc_column.c` not in `store.c`.

HashidaTKS:
I agree that it is enough to reset `ctx->rc` in `proc_column.c` for this case, but considering the general cases, can `grn_ja_replace` return the original `ctx->rc` ?
E.g. `ctx->rc` is set `GRN_CANCEL` while executing `_grn_obj_remove_index` by a cancel request, the same problem may be occurs.

kou:
Generally, Groonga API should return immediately (with error handling) and doesn't call other API when `ctx->rc` isn't `GRN_SUCCESS`.
(The current implementation misses `ctx->rc` check in many places, though...)

For `_grn_obj_remove_index()` case, it should remove the index entirely as much as possible (we may need to reset `ctx->rc` temporary) or doesn't remove the index at all. We should not remove the index partially. But it's not a responsibility of `grn_ja_replace()`. It's a responsibility of  `_grn_obj_remove_index()`.

lib/db.c
```diff
@@ -927,8 +927,6 @@ grn_column_create_similar_internal(grn_ctx *ctx,
                 "%s base column's source doesn't exist: %u",
                 tag,
                 base_source_column_id);
-            grn_obj_remove(ctx, column);
-            column = NULL;
```
HashidaTKS, line 931: We need not execute `grm_obj_remove`  at this line because it is also executed at line 968 ( before fixed ).

We should not substitute `NULL` for `column because  `grm_obj_remove` ( at line 968 ) can not handle null pointer.

HashidaTKS:
@kou 
Thank you for the comment!

Eventually, I have decided to temporarily reset `ctx->rc` in callers of `grn_obj_remove`.
Would you please review this when you have time?

lib/db.c
HashidaTKS: The comment above is based on an assumption that `GRN_OBJ_FIN(ctx, &base_source_ids)` does not change ctx->rc to `GRN_SUCCESS`, so if the assumption is incorrect, I consider another modification.

kou: Generally, `GRN_OBJ_FIN()` resets` ctx->rc`.
But it's not occurred in code between `GRN_API_ENTER`/`GRN_API_RETURN`.

kou (approved):

kou:
Thanks!

PR #1303, opened 2021-12-13
Add a missing error check for grn_ii_posting_add_float()

komainu8:
For example, if we execute ``request_cancel`` while we execute ``sub_filter``, Groonga may return ``GRN_SUCCESS`` even if we execute ``request_cancel`` depending on the timing of executing ``request_cancel``.

Because Groonga doesn't confirm return code of ``grn_ii_posting_add_float()`` currently.

This issue occurs in the search with an accessor.


kou:
Can you add a test case?
Or can you describe when is this occurred?

komainu8:
I added when this issue will occur.

kou:
Another checks are pushed.
This is needless.

PR #1306, opened 2021-12-24
ci package: use qemu-user-static on Impish 

komainu8:
Because qemu-user-static in Ubuntu 20.04 has a crash bug as below.

  * https://bugs.launchpad.net/qemu/+bug/1749393
  * https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=971537

This bug has been already fixed in Ubuntu 20.10.
However, Ubuntu 20.10 already has been EOL.
Therefore, we use qemu-user-static on Ubuntu 21.10.


komainu8:
This PR passed all tests at https://github.com/groonga/groonga/actions/runs/1617886585.

Issue #1309, opened 2021-12-31
grn_snip_add_cond(): accept 32 or more keywords


kou:
The current implementation has fixed size (32) `snip_cond` buffer as `grn_snip::cond`. See `lib/grn_snip.h` for details.

We can accept 32 or more keywords by allocating the buffer dynamically with `GRN_REALLOC()` like `ctx->impl->stack` in `expr.c`.

kou:
@HashidaTKS You may want to work on this.

HashidaTKS:
Sure. I will work on this.

kou:
Done by #1313.

PR #1313, opened 2022-01-10
snippet: Accept 32 or more keywords 

HashidaTKS:
* Modify `snippet` to accept 32 or more keywords
* Fix a bug that Groonga crashes when `grn_expr_get_or_add_var` fails in `func_snippet`

This patch just changed the input keywords to variable length, did not change the result length (the length of `snip_result`).
If we need to add support for the variable length result, I will address it in another PR.


HashidaTKS:
#1309

lib/proc/proc_snippet.c
```diff
@@ -126,6 +126,9 @@ func_snippet(grn_ctx *ctx, int nargs, grn_obj **args, grn_user_data *user_data)
                                        expression,
                                        GRN_TEXT_VALUE(&cache_key),
                                        GRN_TEXT_LEN(&cache_key));
+    if (ctx->rc != GRN_SUCCESS) {
+      goto exit;
+    }
     GRN_OBJ_FIN(ctx, snip_ptr);
     GRN_PTR_INIT(snip_ptr, GRN_OBJ_OWN, GRN_DB_OBJECT);
```
HashidaTKS, line 133: When `grn_expr_get_or_add_var` fails, `snip_ptr` is NULL and null pointer access occur at this line.
I have added the `ctx->rc` check for the above reason.

HashidaTKS:
@kou, @komainu8 

Would you please review this when you have time ?

lib/snip.c
```diff
@@ -436,6 +436,20 @@ grn_snip_get_delimiter_regexp(grn_ctx *ctx, grn_obj *snip, size_t *length)
   return snip_->delimiter_pattern;
 }
 
+grn_rc
```
kou, line 439:
```suggestion
static grn_rc
```

lib/snip.c
```diff
@@ -436,6 +436,20 @@ grn_snip_get_delimiter_regexp(grn_ctx *ctx, grn_obj *snip, size_t *length)
...
+grn_snip_expand_cond(grn_ctx *ctx, grn_snip *snip)
+{
+  size_t cond_capacity = snip->cond_capacity * 2;
```
kou, line 442: How about using more meaningful name?

```suggestion
  size_t new_capacity = snip->cond_capacity * 2;
```

lib/snip.c
```diff
@@ -583,6 +603,22 @@ grn_snip_open(grn_ctx *ctx, int flags, unsigned int width,
     GRN_API_RETURN(NULL);
   }
 
+  ret->cond_capacity = DEFAULT_SNIP_COND_CAPACITY;
+  ret->cond = (snip_cond *)GRN_MALLOC(sizeof(snip_cond) * ret->cond_capacity);
+
+  if (!ret->cond) {
+    if (copy_tag) {
+      if(ret->defaultopentag){
```
kou, line 611:
```suggestion
      if (ret->defaultopentag) {
```

test/command/suite/select/function/snippet/many_keywords.test
```diff
@@ -0,0 +1,42 @@
...
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
+  "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", "SQL", \
```
kou, line 37: Could you use different keywords to confirm all specified keywords are used?

test/command/suite/select/function/snippet/invalid/too_long_key.actual
```diff
@@ -0,0 +1,37 @@
...
+          null
+        ]
+      ],
+      [
+        "[hash][add][(temporary)] too long key"
+      ]
+    ]
+  ]
+]
+#|e| [hash][add][(temporary)] too long key
```
kou, line 37: Why is this occurred? What is the problem of this case?

test/command/suite/select/function/snippet/invalid/too_long_key.actual
```diff
@@ -0,0 +1,37 @@
+table_create Entries TABLE_NO_KEY
```
kou, line 1: `.actual` -> `.expected`.

lib/snip.c
```diff
@@ -662,6 +698,9 @@ grn_snip_close(grn_ctx *ctx, grn_snip *snip)
        cond < cond_end; cond++) {
     grn_snip_cond_close(ctx, cond);
   }
+  if (snip->cond) {
```
kou, line 701: Is this needed? It seems that we always have non-NULL `snip->cond`.

lib/snip.c
```diff
@@ -449,10 +463,16 @@ grn_snip_add_cond(grn_ctx *ctx, grn_obj *snip,
   grn_snip *snip_;
 
   snip_ = (grn_snip *)snip;
-  if (!snip_ || !keyword || !keyword_len || snip_->cond_len >= MAX_SNIP_COND_COUNT) {
+  if (!snip_ || !keyword || !keyword_len) {
     return GRN_INVALID_ARGUMENT;
   }
 
+  if (snip_->cond_len >= snip_->cond_capacity) {
+    if ((rc = grn_snip_expand_cond(ctx, snip_))) {
```
kou, line 471: Could you avoid assigning in `if`? I don't like this style except `while ((x = xxx_next()))` case because it causes long `if (...)` code. It's difficult to read.

```suggestion
    rc = grn_snip_expand_cond(ctx, snip_);
    if (rc) {
```

kou:
Could you rebase on master to recover CI?

test/command/suite/select/function/snippet/invalid/too_long_key.actual
HashidaTKS: I thought it was the intended behavior without giving it much thought, but on second thought, it seems a bug.

proc_snippet.c
```
    grn_proc_func_generate_cache_key(ctx, "snippet", args + 1, nargs - 1, &cache_key);
    ...
    snip_ptr = grn_expr_get_or_add_var(ctx,
                                       expression,
                                       GRN_TEXT_VALUE(&cache_key),
                                       GRN_TEXT_LEN(&cache_key));
```

`grn_proc_func_generate_cache_key` adds all arguments as string to `cache_key` and `GRN_TEXT_LEN` returns the length  including these arguments.
`GRN_TEXT_LEN(&cache_key))`  is greater than the max hash key size (4096) on this case.

I am not confident whether passing `GRN_TEXT_LEN(&cache_key)` to `grn_expr_get_or_add_var` is correct or not.
Do you have any good idea to fix it?
Anyway, I will create a new issue or PR.

HashidaTKS: > I am not confident whether passing GRN_TEXT_LEN(&cache_key) to grn_expr_get_or_add_var is correct or not.

I guess this intends to cache function results including arguments, but in other cases, simply passing a name and its length, what is the difference?

proc_highlight.c
```
    highlighter_ptr =
      grn_expr_get_or_add_var(ctx, expression,
                              GRN_FUNC_HIGHLIGHT_HTML_CACHE_NAME,
                              strlen(GRN_FUNC_HIGHLIGHT_HTML_CACHE_NAME));
```

kou: Ah, I understand.
We should not use cache when the cache key is too large:

```diff
diff --git a/lib/proc/proc_snippet.c b/lib/proc/proc_snippet.c
index d101d502d..60f61b8aa 100644
--- a/lib/proc/proc_snippet.c
+++ b/lib/proc/proc_snippet.c
@@ -115,10 +115,12 @@ func_snippet(grn_ctx *ctx, int nargs, grn_obj **args, grn_user_data *user_data)
   grn_obj *snip_ptr = NULL;
   grn_obj *expression;
   grn_proc_get_info(ctx, user_data, NULL, NULL, &expression);
-  snip_ptr = grn_expr_get_var(ctx,
+  if (GRN_TEXT_LEN(&cache_key) <= GRN_TABLE_MAX_KEY_SIZE) {
+    snip_ptr = grn_expr_get_var(ctx,
                               expression,
                               GRN_TEXT_VALUE(&cache_key),
                               GRN_TEXT_LEN(&cache_key));
+  }
   if (snip_ptr) {
     snip = GRN_PTR_VALUE(snip_ptr);
   } else {
```

Cache is just for performance. We should not use cache if we don't provide normal feature with cache.

We can use a static cache key (`GRN_FUNC_HIGHLIGHT_HTML_CACHE_NAME`) for `highlight_html` because it doesn't have any arguments. It means that all `highlight_html` call should return the same result. So we can cache it with the static cache key.

HashidaTKS: Thank you.
I have modified to check `GRN_TEXT_LEN(&cache_key)`.

test/command/suite/select/function/snippet/many_keywords.test
HashidaTKS: Fixed.

lib/snip.c
HashidaTKS: Fixed.

lib/snip.c
HashidaTKS: Fixed.

lib/snip.c
HashidaTKS: Fixed.

test/command/suite/select/function/snippet/invalid/too_long_key.actual
HashidaTKS: I have renamed and modified the test contents.

lib/snip.c
HashidaTKS: Fixed.

HashidaTKS:
I will fix a test failed with `Request-URI Too Large`

HashidaTKS:
The test will be fixed if https://github.com/groonga/grntest/pull/4 is merged.

https://github.com/groonga/groonga/blob/master/test/command/run-test.sh#L76
Maybe it's better to change to execute `git -C "$grntest_dir" pull` at the line above because `grntest` doesn't seem to be updated if `grntest_dir` already exists.

HashidaTKS:
Ah, do we need to update the gem of `grntest`?
May I update it?

kou:
Ah, yes.
Do you have RubyGems account?

HashidaTKS:
No, I will create an account.

lib/snip.c
```diff
@@ -449,13 +463,21 @@ grn_snip_add_cond(grn_ctx *ctx, grn_obj *snip,
...
+    if (rc) {
+      return rc;
+    }
+  }
+
   cond = snip_->cond + snip_->cond_len;
-  if ((rc = grn_snip_cond_init(ctx, cond, keyword, keyword_len,
-                               snip_->encoding, snip_->normalizer, snip_->flags))) {
+  rc = grn_snip_cond_init(ctx, cond, keyword, keyword_len,
+                               snip_->encoding, snip_->normalizer, snip_->flags);
```
kou, line 479:
```suggestion
                          snip_->encoding, snip_->normalizer, snip_->flags);
```

kou (approved):
+1

komainu8 (approved):
+1

kou:
Could you rebase on master to fix Windows MSVC tests?

lib/snip.c
```diff
@@ -436,6 +436,20 @@ grn_snip_get_delimiter_regexp(grn_ctx *ctx, grn_obj *snip, size_t *length)
   return snip_->delimiter_pattern;
 }
 
+static grn_rc
+grn_snip_expand_cond(grn_ctx *ctx, grn_snip *snip)
+{
+  size_t new_cond_capacity = snip->cond_capacity * 2;
+  snip_cond *new_cond = (snip_cond *)GRN_REALLOC(snip->cond,
+                                          sizeof(snip_cond) * new_cond_capacity);
```
kou, line 444: Indentation is broken.

```suggestion
  size_t new_cond_size = sizeof(snip_cond) * new_cond_capacity;
  snip_cond *new_cond = (snip_cond *)GRN_REALLOC(snip->cond, new_cond_size);
```

HashidaTKS: Right, I have fixed it.
(Sorry for my carelessness)

HashidaTKS:
> Could you rebase on master to fix Windows MSVC tests?

Sure, I have rebased.

kou:
Thanks!

PR #1322, opened 2022-02-03
ci: add a new step for detecting update of nginx 

komainu8:
GitHub Actions detect whether we need to update bundled nginx or not
by comparing the latest version of nginx of upstream and the version
of bundled nginx.

Currently, we confirm the latest version of nginx manually.
However, in this method, we often forget to update nginx that is
bundled with Groonga.


komainu8:
When the version of bundled nginx is 1.21.5 and the version of upstream nginx is 1.21.6, "Detect nginx update" step fails as below.
https://github.com/komainu8/groonga/runs/5046273728?check_suite_focus=true#step:6:6

When the same version the bundled nginx and the upstream nginx, "Detect nginx update" step successes as below.
https://github.com/komainu8/groonga/runs/5046313623?check_suite_focus=true

kou:
We stopped to bundle nginx.

PR #1325, opened 2022-02-08
ci: increase timeout for HTTP chunked test

HashidaTKS:
HTTP chunked test is slower than others. We should improve performance for HTTP chunked
but we increase timeout for now.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -163,7 +163,6 @@ jobs:
         run: |
           bash -x test/command/run-test.sh \
             --n-retries=3 \
-            --timeout=10 \
             --read-timeout=30 \
             --reporter=mark
       - name: "Test: HTTP: load: Apache Arrow"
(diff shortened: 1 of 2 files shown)


kou:
Do we need more timeout not only for chunked case?

HashidaTKS:
Honestly, I don't decide whether we need `timeout` option for all `http` cases.
Recently, only the chunked case seems to time out but I guess it may occur in other `http` cases.
Does the chunked case do something special to occur timeout?

kou:
> Does the chunked case do something special to occur timeout?

We need to comparing slower case deeply to identify it but the chunked case is slower than others obviously:

https://github.com/groonga/groonga/runs/5110173990?check_suite_focus=true

* `Test: stdio`: 7m10s
* `Test: HTTP: load: Apache Arrow`: 7m5s
* `Test: HTTP: command: POST`: 7m2s
* `Test: HTTP: command: chunked`: 8m13s
* `Test: HTTP: groonga-httpd`: 6m52s
* `Test: HTTP: reference count`: 7m49s

We may should improve performance of the chunked case as the real fix.

HashidaTKS:
I see, I have removed the `timeout` option from not chunked cases.

Issue #1326, opened 2022-02-15
Add support for Visual Studio 2022


kou:
It's failed to build `mrbc` without error message:

https://github.com/groonga/groonga/runs/5190836525?check_suite_focus=true#step:10:245

```text
 cl.exe /c /nologo /W3 /we4013 /Zi /MD /O2 /D_CRT_SECURE_NO_WARNINGS /DMRB_STACK_EXTEND_DOUBLING /DMRB_DEBUG /DMRBGEM_MRUBY_BIN_MRBC_VERSION=0.0.0 /I"D:\a\groonga\groonga\vendor\mruby-source\include" /I"D:\a\groonga\groonga\vendor\onigmo-source" /Fo"D:\a\groonga\groonga-build\vendor\mruby-build\host\mrbgems\mruby-bin-mrbc\tools\mrbc\mrbc.obj" "D:\a\groonga\groonga\vendor\mruby-source\mrbgems\mruby-bin-mrbc\tools\mrbc\mrbc.c"
rake aborted!
Command failed with status (127): [cl.exe /c /nologo /W3 /we4013 /Zi /MD /O2 /D_CRT_SECURE_NO_WARNINGS /DMRB_STACK_EXTEND_DOUBLING /DMRB_DEBUG /DMRBGEM_MRUBY_BIN_MRBC_VERSION=0.0.0 /I"D:\a\groonga\groonga\vendor\mruby-source\include" /I"D:\a\groonga\groonga\vendor\onigmo-source" /Fo"D:\a\groonga\groonga-build\vendor\mruby-build\host\mrbgems\mruby-bin-mrbc\tools\mrbc\mrbc.obj" "D:\a\groonga\groonga\vendor\mruby-source\mrbgems\mruby-bin-mrbc\tools\mrbc\mrbc.c"]
```

HashidaTKS:
I guess it's because calling `%VC_PREFIX%\Auxiliary\Build\vcvarsall.bat` fail at the first step.
Visual studio seems to be installed to `Program Files` not `Program Files (x86)` from 2022.

https://github.com/groonga/groonga/runs/5190836525?check_suite_focus=true#step:10:49

We should fix `VC_PREFIX` to support that case in `windows-msvc.yms`
https://github.com/groonga/groonga/blob/master/.github/workflows/windows-msvc.yml#L125

I will create a patch for this.

PR #1328, opened 2022-02-25
ci: Add support for Visual Studio 2022

HashidaTKS:
Visual studio 2022 is installed to `Program Files` not `Program Files (x86)`


HashidaTKS:
#1326

kou:
Good catch!

PR #1330, opened 2022-03-03
logical_range_filter: Add support for immediate un-reference of shards

HashidaTKS:
Add support for immediate un-reference of shards.

### Behavior before patched

All shards are un-referred at the end of the command.

### Behavior after patched

Shards that are no longer referred are un-referred immediately.

PR #1333, opened 2022-03-17
Add necessary unref to grn_obj_get_accessor

HashidaTKS:
This is a patch for https://github.com/groonga/groonga/pull/1330/files#r827579558

An opened table is not closed in cases below.

1. using `time_classify_xxx` with dynamic column
2. using `window` function with dynamic column
3. specifying a non existent column on `output_column` with specifying `filter`

`select` command also has these issues so I guess this is not the matter of `logical_range_filter`.
The commands to reproduce these are below.

**common**
```
plugin_register functions/time
table_create Logs_20170315 TABLE_NO_KEY
column_create Logs_20170315 timestamp COLUMN_SCALAR Time
column_create Logs_20170315 price COLUMN_SCALAR UInt32
load --table Logs_20170315
[
{"timestamp": "2017/03/15 00:00:00", "price": 1000},
{"timestamp": "2017/03/15 11:00:00", "price":  900},
{"timestamp": "2017/03/15 12:00:00", "price":  300},
{"timestamp": "2017/03/15 13:00:00", "price":  200}
]
log_level --level dump
```

**Case 1: using `time_classify_xxx` with dynamic column**
```
select Logs_20170315   --columns[day].stage filtered   --columns[day].type Time   --columns[day].flags COLUMN_SCALAR   --columns[day].value 'time_classify_day(timestamp-36000000000)'   --output_columns _id
[[0,0.0,0.0],[[[4],[["_id","UInt32"]],[1],[2],[3],[4]]]]
#|-| [obj][open] <272>(<Logs_20170315>):<51>(<table:no_key>)
#|-| [obj][open] <273>(<Logs_20170315.timestamp>):<64>(<column:fix_size>)
#|-| [obj][close] <273>(<Logs_20170315.timestamp>):<64>(<column:fix_size>)
```

**Case 2: using `window` function with dynamic column**
```
select Logs_20170315   --columns[price_per_day].stage filtered   --columns[price_per_day].type UInt32   --columns[price_per_day].flags COLUMN_SCALAR   --columns[price_per_day].value 'window_sum(price)'   --columns[price_per_day].window.group_keys 'timestamp'   --output_columns _id
[[0,0.0,0.0],[[[4],[["_id","UInt32"]],[1],[2],[3],[4]]]]
#|-| [obj][open] <272>(<Logs_20170315>):<51>(<table:no_key>)
#|-| [obj][open] <274>(<Logs_20170315.price>):<64>(<column:fix_size>)
#|-| [obj][open] <273>(<Logs_20170315.timestamp>):<64>(<column:fix_size>)
#|-| [obj][close] <274>(<Logs_20170315.price>):<64>(<column:fix_size>)
#|-| [obj][close] <273>(<Logs_20170315.timestamp>):<64>(<column:fix_size>)
```

**Case 3: specifying a non existent column with `output_column`**
```
select Logs_20170315   --filter 'price >= 300'   --output_columns _id,non_existent
[[0,0.0,0.0],[[[3],[["_id","UInt32"]],[1],[2],[3]]]]
#|-| [obj][open] <272>(<Logs_20170315>):<51>(<table:no_key>)
#|-| [obj][open] <274>(<Logs_20170315.price>):<64>(<column:fix_size>)
#|-| [obj][close] <274>(<Logs_20170315.price>):<64>(<column:fix_size>)
```

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -7242,6 +7242,9 @@ grn_obj_get_accessor(grn_ctx *ctx, grn_obj *obj, const char *name, unsigned int 
           next_obj_id = obj->header.domain;
           if (!next_obj_id) {
             // ERR(GRN_INVALID_ARGUMENT, "no such column: <%s>", name);
+            if (obj_is_referred) {
+              grn_obj_unref(ctx, obj);
+            }
             if (!is_chained) {
               grn_obj_close(ctx, (grn_obj *)res);
             }
--- a/test/command/suite/select/columns/stage/filtered/time_reference_count.expected
+++ b/test/command/suite/select/columns/stage/filtered/time_reference_count.expected
@@ -1,0 +1,26 @@
+plugin_register functions/time
+[[0,0.0,0.0],true]
+table_create Logs TABLE_NO_KEY
+[[0,0.0,0.0],true]
+column_create Logs timestamp COLUMN_SCALAR Time
+[[0,0.0,0.0],true]
+column_create Logs price COLUMN_SCALAR UInt32
+[[0,0.0,0.0],true]
+load --table Logs
+[
+{"timestamp": "2017/03/15 00:00:00", "price": 1000},
+{"timestamp": "2017/03/15 11:00:00", "price":  900},
+{"timestamp": "2017/03/15 12:00:00", "price":  300},
+{"timestamp": "2017/03/15 13:00:00", "price":  200}
+]
+[[0,0.0,0.0],4]
+log_level --level dump
+[[0,0.0,0.0],true]
+select Logs   --columns[day].stage filtered   --columns[day].type Time   --columns[day].flags COLUMN_SCALAR   --columns[day].value 'time_classify_day(timestamp-36000000000)'   --output_columns _id
+[[0,0.0,0.0],[[[4],[["_id","UInt32"]],[1],[2],[3],[4]]]]
+#|-| [obj][open] <266>(<Logs>):<51>(<table:no_key>)
+#|-| [obj][open] <267>(<Logs.timestamp>):<64>(<column:fix_size>)
+#|-| [obj][close] <267>(<Logs.timestamp>):<64>(<column:fix_size>)
+#|-| [obj][close] <266>(<Logs>):<51>(<table:no_key>)
+log_level --level notice
+[[0,0.0,0.0],true]
--- a/test/command/suite/select/columns/stage/filtered/time_reference_count.test
+++ b/test/command/suite/select/columns/stage/filtered/time_reference_count.test
@@ -1,0 +1,25 @@
+#$GRN_ENABLE_REFERENCE_COUNT=yes
+plugin_register functions/time
+table_create Logs TABLE_NO_KEY
+column_create Logs timestamp COLUMN_SCALAR Time
+column_create Logs price COLUMN_SCALAR UInt32
+load --table Logs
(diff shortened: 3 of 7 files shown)


HashidaTKS:
I would like to merge this after #1330 is merged because tests related to this PR are added at #1330.

kou:
Could you add a test for this with `select`?

HashidaTKS:
Sure

kou:
Are all of 3 cases caused by the same pattern? Or there are multiple causes?

If there are multiple causes, could you split this pull request?

HashidaTKS:
As far as I confirmed, these 3 cases are caused by the same pattern.

However, I have noticed that a similar but not the same issue exists.
I am investigating that...

HashidaTKS:
I have added tests and rebased on #1330 because some expected test results are updated by this change.
So this patch include the change of #1330.
Please review this after #1330 is merged.

kou:
In the description, you said we can reproduce this problem with `select` not `logical_range_filter`.
Why do you use `logical_range_filter` not `select` for tests for this?

HashidaTKS:
The tests for this added the commit below and they use `select`.

https://github.com/groonga/groonga/pull/1333/commits/984c3c158eab66a74efe2e752a0151292191696a

Some tests using `logical_range_filter` did not closed shards by the same reason, so the tests are also updated by this change because they are not closed

kou:
OK. We don't need #1330 to test this problem, right?
If it's true, could you rebase on master not #1330 for easy to review?

HashidaTKS:
OK, I have rebased this on master.

lib/db.c
```diff
@@ -7242,6 +7237,9 @@ grn_obj_get_accessor(grn_ctx *ctx, grn_obj *obj, const char *name, unsigned int
           next_obj_id = obj->header.domain;
           if (!next_obj_id) {
             // ERR(GRN_INVALID_ARGUMENT, "no such column: <%s>", name);
+            if (obj_is_referred) {
+              grn_obj_unref(ctx, obj);
+            }
```
kou, line 7242: The `time_classify_day(timestamp-36000000000)` case only needs this.

kou: All other cases are also need only this.
Do we really need other changes?

HashidaTKS: Yes, they need only this change.
However, I thought it was better to also fix other not `unref` cases because it was the same bug in the same function.

kou: Can you create tests for other changes?

HashidaTKS: I guess they are tested at the CI tests for the case `GRN_ENEBLE_REFERENCE_COUNT=yes`.

kou: How can I confirm it?

kou: If tests for other cases already exist, CI with `GRN_ENABLE_REFERENCE_COUNT=yes` is red.

HashidaTKS: I mean if CI with `GRN_ENABLE_REFERENCE_COUNT=yes` is not red, at least we don't `unref` overly.

HashidaTKS: Ok, it is not enough to check CI with GRN_ENABLE_REFERENCE_COUNT=yes is not red,
I have separated the other changes into another PR and I will work on this to add test them.

#1335

kou:
Thanks.

PR #1334, opened 2022-03-17
Add necessary unref to accessor.c and report.c

HashidaTKS:
Related: #1333,  #1330

When executing `grn_accessor_execute` or `grn_report_table`, a referred object is not un-referred  in some cases.
I have added necessary `unref`s for those.

Please see the added `reference_count.test` for reproducing the issue.
This affect to #1330.


lib/report.c
```diff
@@ -87,6 +87,9 @@ grn_report_table(grn_ctx *ctx,
       GRN_TEXT_PUT(ctx, &description, name, name_size);
       GRN_TEXT_PUTS(ctx, &description, ">");
     }
+    if (target != table) {
+      grn_obj_unref(ctx, target);
```
kou, line 91: This is invalid.
We use `target` after this `grn_obj_unref` for `target = grn_ctx_at(ctx, target->header.domain)`. If `target` is closed by this `grn_obj_unref()`, `target->header.domain` touches invalid memory.

test/command/suite/select/column/stage/initial/reference_count.test
```diff
@@ -0,0 +1,30 @@
...
+{"_key": "Pen",   "price":  510},
+{"_key": "Food",  "price":  500},
+{"_key": "Drink", "price":  300}
+]
+
+#@add-important-log-levels dump
+#@add-ignore-log-pattern /\A\[io\]/
+log_level --level dump
+select Items \
+  --column[price_with_tax].stage initial \
```
kou, line 22: Could you use `columns` instead of `column` because `column` is deprecated?

lib/report.c
HashidaTKS: Right, I have fixed it.

test/command/suite/select/column/stage/initial/reference_count.test
HashidaTKS: I have fixed it.

lib/report.c
```diff
@@ -65,7 +65,7 @@ grn_report_table(grn_ctx *ctx,
                  grn_obj *table)
 {
   grn_obj description;
-  grn_obj *target;
+  grn_obj *target, *previous_target = NULL;
```
kou, line 68: Could you declare one variable per line?

lib/report.c
```diff
@@ -87,6 +87,13 @@ grn_report_table(grn_ctx *ctx,
       GRN_TEXT_PUT(ctx, &description, name, name_size);
       GRN_TEXT_PUTS(ctx, &description, ">");
     }
+    if (previous_target && previous_target != table) {
+      grn_obj_unref(ctx, previous_target);
+    }
+    previous_target = target;
```
kou, line 93: Could you move this to `for (...)` to indicating that this is update in each loop?

HashidaTKS: I have fixed it.

lib/report.c
HashidaTKS: I have fixed it.

HashidaTKS:
I have noticed that `grn_report_column` and `grn_report_accessor` seem to have the same issue.
I will fix them another PR.

kou:
It makes sense.

PR #1338, opened 2022-03-24
logical_range_filter: separate window from logical_range_filter into an independent file

HashidaTKS:
Separate `window` from `logical_range_filter` into an independent file.

This is a preparation for modification of `logical_count` to un-refer shards immediately like `logical_range_filter`.


plugins/sharding/window.rb
```diff
@@ -0,0 +1,236 @@
...
+        [@year, @month, @day, @hour, @minute, @second, @microsecond]
+      end
+      
+      def <=>(other)
+        (values + [@border == :include ? 1 : 0]) <=>
+          (other.values + [other.border == :include ? 1 : 0])
+      end
+    end
+  end
+end
```
kou, line 236: Could you add a new line?

kou:
Could you also update `plugins/sharding/sources.am`?

plugins/sharding/window.rb
HashidaTKS: Sorry for forgetting it so many times...
I have fixed it.

HashidaTKS:
@kou 

Thank you for the review.

> Could you also update plugins/sharding/sources.am?

I have added `window.rb` into `plugins/sharding/sources.am`.

kou:
Thanks.

PR #1341, opened 2022-03-29
logical_count: improve memory usage

HashidaTKS:
Overall, I have changed `logical_count` to immediately `unref` shards or `close` temporary objects that are no longer referred like `logical_range_filter`.

TODO:

- [x] Refactor
  * Reconsider name of method, class and so on
  * Separate common method with `logical_range_filter` into a common class
    * e.g. `detect_time_classify_types`
    * If they have not match in common, I will skip this.
- [x] More test
    * Especially performance test

PR #1342, opened 2022-03-29
logical_range_filter: Separate a base of ExecuteContext into another class

HashidaTKS:
* Separate a base of ExecuteContext into another class `ExecuteContextBase`.
* Change `ExecuteContext` to inherite `ExecuteContextBase`.

This is a preparation for  #1341.


plugins/sharding/execute_context_base.rb
```diff
@@ -0,0 +1,87 @@
+module Groonga
+  module Sharding
+    class ExecuteContextBase
```
kou, line 3: How about `StreamExecuteContext` or something because this isn't suitable for non stream based execution such as `logical_select`?

HashidaTKS: It make sense.
I have renamed it.

PR #1346, opened 2022-03-31
logical_range_filter: Make result_sets enumeration sharable

HashidaTKS:
logical_range_filter: Make result_sets enumeration sharable on `StreamShardExecutor`

### DIFF
--- a/plugins/sharding/logical_range_filter.rb
+++ b/plugins/sharding/logical_range_filter.rb
@@ -239,6 +239,7 @@ module Groonga
 
           enumerate_result_sets do |result_set|
             sort_result_set(result_set)
+            return if @window && @context.current_limit.zero?
           end
         end
 
--- a/plugins/sharding/logical_range_filter.rb
+++ b/plugins/sharding/logical_range_filter.rb
@@ -239,7 +239,7 @@ module Groonga
 
           each_result_set do |result_set|
             sort_result_set(result_set)
-            return if @window && @context.current_limit.zero?
+            return if @context.current_limit.zero?
           end
         end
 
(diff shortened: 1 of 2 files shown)


plugins/sharding/stream_shard_executor.rb
```diff
@@ -43,6 +43,28 @@ def have_record?
         true
       end
 
+      protected
+      def enumerate_result_sets
```
kou, line 47: `each_result_set` is better for Ruby.

plugins/sharding/stream_shard_executor.rb
```diff
@@ -43,6 +43,28 @@ def have_record?
         true
       end
 
+      protected
```
kou, line 46: Why is it needed?
Generally, `protected` isn't used in Ruby.

HashidaTKS: > Generally, protected isn't used in Ruby.

I see. (I didn't know that.)

I have removed this method access modifier.

plugins/sharding/stream_shard_executor.rb
HashidaTKS: I have renamed it.

plugins/sharding/logical_range_filter.rb
```diff
@@ -237,23 +237,9 @@ def execute
...
-              end
-            end
-          else
-            apply_filtered_dynamic_columns
-            @filtered_result_sets.each do |result_set|
-              sort_result_set(result_set)
-            end
+          each_result_set do |result_set|
+            sort_result_set(result_set)
+            return if @window && @context.current_limit.zero?
```
kou, line 242: Is `@window &&` needed?

HashidaTKS: I had added this just because originally `@context.current_limit.zero?` was executed only when `@window`.
However, even removing `@window &&`, all `logical_range_filter` tests were passed.
So I have removed this `@window &&`.

PR #1347, opened 2022-03-31
logical_range_filter: Move order to StreamExcuteContext

HashidaTKS:
Move the `order` attribute to `StreamExcuteContext` from `ExecuteContext`

### DIFF
--- a/plugins/sharding/stream_execute_context.rb
+++ b/plugins/sharding/stream_execute_context.rb
@@ -8,6 +8,7 @@ module Groonga
       attr_reader :time_classify_types
       attr_reader :order
       def initialize(command_name, input)
+        @command_name = command_name
         @input = input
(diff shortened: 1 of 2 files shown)


plugins/sharding/stream_execute_context.rb
```diff
@@ -56,6 +58,23 @@ def need_look_ahead?
...
+        return :ascending if order.nil?
+
+        case order
+        when "ascending"
+          :ascending
+        when "descending"
+          :descending
+        else
+          message =
+            "[logical_range_filter] #{name} must be " +
```
kou, line 72: We should use `command_name` here.

HashidaTKS: Sorry, I have overlooked.
I have fixed it.

PR #1361, opened 2022-05-25
hash: reduce the size of the table for setting the result set in advance

komainu8:
Groonga improves performance for setting results set by making the table for setting the result set in advance.

However, if the number of match records is few, the cost of making the table is high.
Because the size of the table same as a table of the search targets.

This modification is made the table for setting the result set in advance in two steps.

1. Groonga makes the table the size of half a table of the search targets before the search execution.

2. If the size of the result sets is over the size of half a table of the search targets while the search execution, Groonga expands the size of the table for setting the result set in advance.

By doing so, we improve performance if the number of match records is few.

### DIFF
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -3622,9 +3622,12 @@ grn_hash_add_table_cursor(grn_ctx *ctx,
   }
 
   /* lock */
+  grn_obj *table = grn_table_cursor_table(ctx, cursor);
+  const uint32_t target_table_size = grn_table_size(ctx, table);
+  const uint32_t reset_threshold =
+    ((target_table_size / 2) == 0) ? target_table_size : (target_table_size / 2);
+
   if (op == GRN_OP_OR) {
-    grn_obj *table = grn_table_cursor_table(ctx, cursor);
-    const uint32_t reset_threshold = grn_table_size(ctx, table);
     rc = grn_hash_ensure_rehash(ctx, hash, reset_threshold, tag);
     if (rc != GRN_SUCCESS) {
       return rc;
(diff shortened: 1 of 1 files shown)


kou:
Could you add reporter information of reporter such as the following?

```text
Fix [groonga-dev,04992]

Reported by Koya Masuda. Thanks!!!
```

komainu8:
I added reporter information in the commit message ( https://github.com/groonga/groonga/pull/1361/commits/be8afdf04ba3a689ee185b4ce4f2eac29a6d73c6 ).

komainu8:
I measure this modification.

### Versions

* CentOS 7
* Mroonga 12.02
* Groonga 12.0.3(before) -> 12.04(after)( https://github.com/groonga/groonga/suites/6687591521/artifacts/254050280 )
* MariaDB 10.4.24

### Tables and the number of records

CREATE TABLE `urls` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `url` mediumtext DEFAULT NULL,
  PRIMARY KEY (`id`),
  FULLTEXT KEY `fulltext_url` (`url`)
) ENGINE=Mroonga DEFAULT CHARSET=utf8mb4;

The number of records is ten million(10,000,000).

### Result

The result of the measurement is the following.

(Before modification)

3618 rows in set (3.491 sec)
3618 rows in set (3.129 sec)

(After modification)

3618 rows in set (0.423 sec)
3618 rows in set (0.316 sec)

kou:
Data and `SELECT` are missing to reproduce this.

komainu8:
Ah...Sorry.
I added Data and SELECT in https://github.com/groonga/groonga/pull/1361#issuecomment-1141800121.

kou:
It seems that this case is faster without this change. Could you try with https://github.com/groonga/groonga/suites/6737636090/artifacts/257176572 in  https://github.com/groonga/groonga/actions/runs/2418613573 ?

komainu8:
I tried this case with https://github.com/groonga/groonga/suites/6737636090/artifacts/257176572.


### Result

The result of the measurement is the following.

3618 rows in set (0.419 sec)
3618 rows in set (0.257 sec)
3618 rows in set (0.277 sec)

komainu8:
I close this PR.
Because this modification already merged to master by 6bd5a6bb916adc429bb2036334d6b6e2d970f4b8.

PR #1362, opened 2022-05-26
munin: fix a wrong regular expression of thread id

komainu8:
Currently, the thread id has "0x" as the prefix in the query log of Groonga.

### DIFF
--- a/data/munin/groonga_query_performance_
+++ b/data/munin/groonga_query_performance_
@@ -169,7 +169,7 @@ elapsed_times = []
 File.open(target_query_log_path) do |log_file|
   ReverseLineReader.new(log_file).each do |line|
     case line
-    when /\A(\d+)-(\d+)-(\d+) (\d+):(\d+):(\d+)\.(\d+)\|([\da-f])+\|<(\d+) rc=0$/
+    when /\A(\d+)-(\d+)-(\d+) (\d+):(\d+):(\d+)\.(\d+)\|0x([\da-f])+\|<(\d+) rc=0$/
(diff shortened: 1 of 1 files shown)


data/munin/groonga_query_performance_
```diff
@@ -169,7 +169,7 @@ elapsed_times = []
 File.open(target_query_log_path) do |log_file|
   ReverseLineReader.new(log_file).each do |line|
     case line
-    when /\A(\d+)-(\d+)-(\d+) (\d+):(\d+):(\d+)\.(\d+)\|([\da-f])+\|<(\d+) rc=0$/
+    when /\A(\d+)-(\d+)-(\d+) (\d+):(\d+):(\d+)\.(\d+)\|0x([\da-f])+\|<(\d+) rc=0$/
```
kou, line 172: `0x` should be included in `context` and `0x` doesn't exist on Windows.

```suggestion
    when /\A(\d+)-(\d+)-(\d+) (\d+):(\d+):(\d+)\.(\d+)\|((?:0x)?[\da-f])+\|<(\d+) rc=0$/
```

kou: This is not a thread ID. It's an address of `grn_ctx`.

The original expression has a bug... The first character is only used for `context`...

```suggestion
    when /\A(\d+)-(\d+)-(\d+) (\d+):(\d+):(\d+)\.(\d+)\|((?:0x)?[\da-f]+)\|<(\d+) rc=0$/
```

PR #1366, opened 2022-06-07
group: support single reference weight vector

naoa:


kou:
Thanks!

naoa:
Thanks for your reviewing and merged!

PR #1367, opened 2022-06-08
group: support reference vector via deep accessor

naoa:


lib/group.c
```diff
@@ -931,15 +932,21 @@ grn_table_group_single_key_records(grn_ctx *ctx, grn_obj *table,
...
+        }
+        if (a->next->action == GRN_ACCESSOR_GET_COLUMN_VALUE &&
+            a->next->obj && !a->next->next) {
+          real_key = a->next->obj;
+          data.need_resolved_id = true;
+        }
       }
     }
-    group_key_init(ctx, &(data.key), real_key);
+    group_key_init(ctx, &(data.key), real_key, key);
```
kou, line 949: Should we pass `NULL` instead of `key` when `key` isn't an accessor?

lib/group.c
```diff
@@ -931,15 +932,21 @@ grn_table_group_single_key_records(grn_ctx *ctx, grn_obj *table,
...
+           a->next;
+           a = a->next) {
+        if (a->action != GRN_ACCESSOR_GET_KEY) {
+          break;
+        }
+        if (a->next->action == GRN_ACCESSOR_GET_COLUMN_VALUE &&
+            a->next->obj && !a->next->next) {
+          real_key = a->next->obj;
+          data.need_resolved_id = true;
+        }
```
kou, line 946: Can we simplify this?

e.g.:

```c
      grn_accessor *a;
      for (a = (grn_accessor *)key;
           a->action == GRN_ACCESSOR_GET_KEY && a->next;
           a = a->next) {
        /* do nothing */
      }
      if (a->next &&
          a->next->action == GRN_ACCESSOR_GET_COLUMN_VALUE &&
          a->next->obj &&
          !a->next->next) {
        real_key = a->next->obj;
        data.need_resolved_id = true;
      }
```

lib/group.c
```diff
@@ -673,9 +676,7 @@ grn_table_group_single_key_records_resolve_id(grn_ctx *ctx,
                                               grn_id id)
 {
   if (data->need_resolved_id) {
-    void *key;
-    grn_table_cursor_get_key(ctx, cursor, &key);
-    return *((grn_id *)key);
+    return grn_accessor_resolve_id(ctx, data->key.accessor, id);
```
kou, line 679: This may be slow for no deep case. Can you measure this with no deep case? (I hope that we have continuous benchmark mechanism...)

naoa: I see.
I will try the benchmark later.

lib/group.c
kou: Yes. `grn_obj_column()` returns a column itself not an accessor for the column with a column name of a table.

lib/group.c
kou: Thanks!
If this decreases performance, we can use the current logic only for no deep case.

naoa: I benchmarked with a drilldown of 3,144,060 reference vectors of results that no deep case.
There is a slight speed degradation.
I changed to switch them depends to the depth of the accessor.

Case 1. Use grn_accessor_resolve_id
8.912397623062134 sec
7.918306112289429 sec
7.794296026229858 sec
7.789118528366089 sec
7.804325819015503 sec

Case 2. Use grn_table_cursor_get key
8.694790124893188 sec
7.541107654571533 sec
7.652536869049072 sec
7.514059543609619 sec
7.609168767929077 sec

Case 3. Switch grn_table_cursor_get/grn_accessor_resolve_id  depends to the depth of the accessor.
8.73039174079895 sec
7.665851831436157 sec
7.721802234649658 sec
7.707949638366699 sec
7.606190919876099 sec

kou:
Thanks!

PR #1374, opened 2022-06-29
select: Include "grn_ctx.hpp" even if in an environment without Apache Arrow

komainu8:
Because we use "grn::ChildCtxReleaser" in "proc_select.cpp".
ChildCtxReleaser class is defined in "grn_ctx.hpp".
Therefore, if we don't include "grn_ctx.hpp", a build error occurs in an environment without Apache Arrow.


kou:
Good catch!

PR #1386, opened 2022-08-10
doc: Remove `slices.table` description

HashidaTKS:
The feature `slices.table` is not implemented.
Maybe it was intended `drilldowns.table`.


HashidaTKS:
Cc: @kou

PR #1392, opened 2022-08-22
db: add a new condition for stopping result set marge when fail to add the result set

komainu8:
Currently, Groonga doesn't stop making a result set when failing to
add the result set of a OR search.

Therefore, when the number of results of the OR search is many and a
query has many OR conditions, Groonga may response slow with the
"request_cancel" command.

if we execute the "request_cancel" command while Groonga adds the
result set of the OR search, the addition to the result set fails.

Therefore, Groonga output  a error of "failed to add:".
Groonga may output this error a lot because this error occurs every
failing to add the result set.

If Groonga output this error a lot, Groonga's response is very slow.

This modification is stopping making the result set on the way when
failing to add the result set.
Thereby, the command responds quickly.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -6081,23 +6081,22 @@ grn_table_setoperation_or(grn_ctx *ctx,
         } else {
           break;
         }
-      } else {
-        if (added) {
-          grn_memcpy(value_dest, value_src, data->value_size);
-          grn_rset_recinfo *ri_dest = value_dest;
-          ri_dest->score *= data->weight_factor;
-        } else {
-          grn_rset_recinfo *ri_dest = value_dest;
-          grn_rset_recinfo *ri_src = value_src;
-          grn_table_add_subrec(ctx,
-                               data->table_dest,
-                               ri_dest,
-                               ri_src->score * data->weight_factor,
-                               NULL,
-                               0);
-        }
-        grn_table_setoperation_merge_columns(ctx, data, id_dest, id_src);
       }
+      if (added) {
+        grn_memcpy(value_dest, value_src, data->value_size);
+        grn_rset_recinfo *ri_dest = value_dest;
+        ri_dest->score *= data->weight_factor;
+      } else {
+        grn_rset_recinfo *ri_dest = value_dest;
+        grn_rset_recinfo *ri_src = value_src;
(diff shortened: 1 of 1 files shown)


lib/db.c
```diff
@@ -6075,7 +6075,9 @@ grn_table_setoperation_or(grn_ctx *ctx,
                                               key_src_size,
                                               &value_dest,
                                               &added);
-      if (id_dest != GRN_ID_NIL) {
+      if (id_dest == GRN_ID_NIL) {
+        break;
```
kou, line 6079: Could you also check `ctx->rc`? I'm not sure we can break for all "failed to add" cases.

```c
if (id_dest == GRN_ID_NIL) {
  if (ctx->rc == GRN_SUCCESS) {
    continue;
  } else{
    break;
  }
}
```

BTW, do we need the same logic for `!dat->have_subrec` case too?

lib/db.c
```diff
@@ -6075,7 +6075,13 @@ grn_table_setoperation_or(grn_ctx *ctx,
...
                                               &value_dest,
                                               &added);
-      if (id_dest != GRN_ID_NIL) {
+      if (id_dest == GRN_ID_NIL) {
+        if (ctx->rc == GRN_SUCCESS) {
+          continue;
+        } else {
+          break;
+        }
+      } else {
```
kou, line 6084:
```suggestion
      }

```

Could you also remove this `else` and unindent the current `else` block contents?

lib/db.c
komainu8: I see. I also add a check of `ctx->rc`.

> BTW, do we need the same logic for !data->have_subrec case too?
Sure.
We need this logic for `!data->have_subrec` case. I added at d4a013b40bdefdbc4eab25f49f2333ae71da0aa5.

kou:
Do we need similar codes for other `grn_table_setoperation_*()` too?

komainu8:
> Do we need similar codes for other `grn_table_setoperation_*()` too?

No.
Because `grn_table_setoperation_and`, `grn_table_setoperation_and_not`, and `grn_table_setoperation_adjust` are not output error logs even if we execute "request_cancle" while Groonga execute loop.
Therefore, this problem does not occur in these functions.

PR #1404, opened 2022-09-02
ci: workflows except Package workflow ignore the modification of documents

komainu8:
Because tests of linux, macos, windows-mingw, and windows-msvc
are not affected by the modification of documents.

Currently, when we modify documents, many ci are executed.
Thereby a queue of GitHub Actions is full.
This is inefficient.

Therefore, we enable ignore the modification of documents in these workflows.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,7 +6,11 @@ on:
       - 'maintenance/**'
     tags:
       - '*'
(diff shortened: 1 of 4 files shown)


kou:
Can we add a CI job that only build documents and detect documentation related problems? (`make dist` is OK for it.)
If it's true, we can introduce this.

PR #1414, opened 2022-09-20
package: Add support for AlmaLinux 9

HashidaTKS:
Add support for AlmaLinux 9.

Note: 

We need to install `apache-arrow-release-latest.rpm` on AlmaLinux 9 because `packages.groonga.org` don't provide Apache Arrow for AlmaLinux 9+.

We provide Apache Arrow from `packages.groonga.org` for compatibility of users who didn't install Apache Arrow, but currently there is no user who use Groonga on AlmaLinux 9 because AlmaLinux 9 was recently released and we add support for it from this patch. So we don't need to consider the compatibility (of Apache Arrow) for the AlmaLinux 9 users. This is because that `packages.groonga.org` don't provide Apache Arrow for AlmaLinux 9+.


HashidaTKS:
@kou @komainu8 

kou:
We also need to add support for AlmaLinux 9 aarch64.

PR #1421, opened 2022-09-28
ii: remove redudant normalize on search as much as possible

komainu8:
Some normalizes such as `NormalizerTable` aren't idempotent. So we should not normalize the input
multiple times.

The current `token_info_open()` related codes normalize the input multiple times. This change avoids
multiple normalizations as much as possible. If a tokenizer enables `report_source_location`, we can
avoid multiple normalizations all time. But if a tokenizer doesn't enable `report_source_location`, we may
normalize the input multiple times.

See the added tests for problem cases.


komainu8:
@kou @HashidaTKS Could you review this PR?

lib/db.c
```diff
@@ -3748,6 +3748,34 @@ grn_table_add_by_key(grn_ctx *ctx, grn_obj *table, grn_obj *key, int *added)
   return id;
 }
 
+grn_id
+grn_table_get_without_normalize(grn_ctx *ctx,
```
kou, line 3752: I don't want to add this. If we add this, we want to add `grn_table_cursor_open_without_nromalize()`.

I think that this is a `ii` problem. `ii` should not re-get token from token id. `ii` should use source of a token instead of re-getting token. We may be able to use `grn_token_get_source_offset()` and `grn_token_get_source_length()` for it. (We can add `grn_token_get_source()` for convenient.)

PR #1423, opened 2022-09-28
doc: Add description about installing AlmaLinux 9

HashidaTKS:
Add description about installing AlmaLinux 9


HashidaTKS:
@kou @komainu8 

PR #1438, opened 2022-10-31
ci linux: use ubuntu-22.04

HashidaTKS:
Because Apache Arrow requires C++17.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -42,7 +42,7 @@ jobs:
             configure-options: >-
               CC=clang
               CXX=clang++
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


kou:
I'm not against using Ubuntu 22.04 but why is it related? `g++`/`clang++` on Ubuntu 20.4 should support C++17.

HashidaTKS:
> I'm not against using Ubuntu 22.04 but why is it related? g++/clang++ on Ubuntu 20.4 should support C++17.

Just because Ubuntu 22.04 seems to use g++11.1* as default and it uses C++17 as default.
So I guessed we can fix this problem just using Ubuntu 22.04.
But I wan not have confident about it and try it with this PR. 
(I had to create this PR as draft and add descriptions about it, Sorry).

komainu8:
We close this PR.
Because we have already used `ubuntu-latest`.

Currently, `ubuntu-latest` is Ubuntu 24.04. So, this problem has already resolved.

PR #1439, opened 2022-10-31
ci linux: specify c++ version at gem install

HashidaTKS:
Because Apache Arrow requires C++17

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -304,7 +304,8 @@ jobs:
           sudo env MAKEFLAGS=-j$(nproc) gem install \
             grntest \
             pkg-config \
-            red-arrow
+            red-arrow \
+            -- --with-cxxflags=-std=c++17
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -304,7 +304,8 @@ jobs:
           sudo env MAKEFLAGS=-j$(nproc) gem install \
             grntest \
             pkg-config \
-            red-arrow
+            red-arrow \
+            -- --with-cxxflags=-std=c++17
```
kou, line 308: This should be fixed when red-arrow 10.0.0 is published.
So we don't need to do anything here.

kou:
Ah, is this a blocker for a new release?

HashidaTKS:
> Ah, is this a blocker for a new release?

Packages are created successfully, so I think we can release new Groonga.
But I think it is better to check all tests are passed for checking newly bugs.

PR #1447, opened 2022-11-07
package: add support for CentOS Stream 9

komainu8:


kou:
We don't add support for CentOS Stream.

PR #1451, opened 2022-11-14
packages: download latest source archive in dist task

HashidaTKS:
By this change, we can download the latest source archive from the github release page into the top of the repository by executing "cd packages && rake dist".

### DIFF
--- a/packages/Rakefile
+++ b/packages/Rakefile
@@ -42,6 +42,8 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
   end
 
   def define_archive_task
+    source_archive_url = built_package_url(nil, @archive_base_name)
+    download(source_archive_url, "..")
     file @original_archive_name do
       File.symlink("../#{@original_archive_name}",
                    @original_archive_name)
--- a/packages/Rakefile
+++ b/packages/Rakefile
@@ -42,9 +42,11 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
   end
 
   def define_archive_task
-    file @original_archive_name do
+    file "../#{@original_archive_name}" do
       source_archive_url = built_package_url(nil, @archive_base_name)
       download(source_archive_url, "..")
+    end
+    file @original_archive_name => "../#{@original_archive_name}" do
       File.symlink("../#{@original_archive_name}",
                    @original_archive_name)
     end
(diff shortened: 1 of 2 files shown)


packages/packages-groonga-org-package-task.rb
```diff
@@ -40,6 +41,14 @@ def detect_release_time
     end
   end
 
+  def define_source_task
+    desc "Download the latest source archive"
+    task :source do
+      source_archive_url = built_package_url(nil, @archive_base_name)
+      download(source_archive_url, "..")
+    end
```
kou, line 49: Can we rewrite `define_archive_task` instead?

HashidaTKS: Adressed. (Also changed the description of this PR.)

packages/Rakefile
```diff
@@ -42,6 +42,8 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
   end
 
   def define_archive_task
+    source_archive_url = built_package_url(nil, @archive_base_name)
+    download(source_archive_url, "..")
```
kou, line 46: If we write this here, this is always processed. (`rake --help` also processes this.)
We should put this into `file do ... end`.

packages/packages-groonga-org-package-task.rb
kou: We need to update title too.

packages/Rakefile
HashidaTKS: Fixed.

HashidaTKS:
@kou 

Would you re-review this PR?

kou:
How about this?

```diff
diff --git a/packages/Rakefile b/packages/Rakefile
index bc7939453..7c5704cfc 100644
--- a/packages/Rakefile
+++ b/packages/Rakefile
@@ -42,9 +42,14 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
   end
 
   def define_archive_task
-    file @original_archive_name do
-      File.symlink("../#{@original_archive_name}",
-                   @original_archive_name)
+    file "../#{@original_archive_name}" do
+      source_archive_url = built_package_url(:source, @archive_base_name)
+      download(source_archive_url, "..")
+    end
+
+    file @original_archive_name => "../#{@original_archive_name}" do
+      ln_s("../#{@original_archive_name}",
+           @original_archive_name)
     end
   end
 
```

HashidaTKS:
@kou 

Thanks, I have changed the code as you indicated, and it worked fine as far as I tested.

PR #1484, opened 2023-01-11
doc: add bootstrap.js/css into files.am

HashidaTKS:
`bootstrap.js` and `bootstrap.css` are necessary to the latest document site.
They are copied (moved) into result files of `make update-document` by adding them into files.am.

The reason why we need to add `bootstrap.js` and `bootstrap.css`  is because `pydata-sphinx-theme` split bootstrap and theme styling.

https://github.com/pydata/pydata-sphinx-theme/pull/994

### DIFF
--- a/doc/files.am
+++ b/doc/files.am
@@ -4291,9 +4291,11 @@ html_files_relative_from_locale_dir = \
 	html/_static/minus.png \
 	html/_static/plus.png \
 	html/_static/pygments.css \
+	html/_static/scripts/bootstrap.js \
 	html/_static/scripts/pydata-sphinx-theme.js \
 	html/_static/searchtools.js \
 	html/_static/sphinx_highlight.js \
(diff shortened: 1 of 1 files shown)


HashidaTKS:
@kou @komainu8 

komainu8:
+1

BTW, why do this modification to be become necessary?
Is it because the documentation was generated with Sphinx 6.0.x?

HashidaTKS:
> BTW, why do this modification to be become necessary?
> Is it because the documentation was generated with Sphinx 6.0.x?

Because `pydata-sphinx-theme` split bootstrap and theme styling.

https://github.com/pydata/pydata-sphinx-theme/pull/994

komainu8:
I see.
Could you add this reason to commit message?

HashidaTKS:
> Could you add this reason to commit message?

Sure, I have added.

PR #1505, opened 2023-02-01
packages yum: add support for Oracle Linux

komainu8:


komainu8:
@kou @HashidaTKS Could you review this PR?

packages/yum/test.sh
```diff
@@ -13,26 +13,38 @@ case ${os} in
...
+        ;;
+      8)
+        DNF="dnf --enablerepo=powertools"
+        ;;
+      *)
+        DNF="dnf --enablerepo=crb"
+        ${DNF} install -y \
+          https://apache.jfrog.io/artifactory/arrow/${os}/${version}/apache-arrow-release-latest.rpm
+        ;;
+    esac
```
HashidaTKS, line 47: [nits]
`;;`  is not required in the last case, but `;;` is added for the last cases in this file, so I think it is prefer to add `;;`.

```suggestion
    esac
    ;;
```

PR #1506, opened 2023-02-02
NormalizerNFKC: add `unify_katakana_trailing_o` option

HashidaTKS:
When `unify_katakana_trailing_o` is specified, `NormalizerNFKC*` normalize characters as below.

オオ -> オウ
コオ -> コウ
ソオ -> ソウ
トオ -> トウ
ノオ -> ノウ
ホオ -> ホウ
モオ -> モウ
ヨオ -> ヨウ
ロオ -> ロウ
ゴオ -> ゴウ
ゾオ -> ゾウ
ドオ -> ドウ
ボオ -> ボウ
ポオ -> ポウ

Usage: 

```
normalize \
  'NormalizerNFKC130("unify_katakana_trailing_o", true, \
                     "report_source_offset", true)' \
  "オオコオソオトオノオホオモオヨオロオゴオゾオドオボオポオ" \
  WITH_CHECKS|WITH_TYPES
```


HashidaTKS:
@kou 

lib/normalizer.c
```diff
@@ -1783,6 +1783,84 @@ grn_nfkc_normalize_unify_katakana_g_sounds(grn_ctx *ctx,
   return current;
 }
 
+static const unsigned char *
+grn_nfkc_normalize_unify_katakana_trailing_o(grn_ctx *ctx,
+                                           const unsigned char *start,
```
kou, line 1788: Could you fix indent?

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
```diff
@@ -0,0 +1,129 @@
...
+      "katakana",
+      "katakana",
+      "katakana",
+      "katakana",
+      "katakana",
+      "katakana",
+      "katakana",
+      "katakana"
+    ],
+    "checks": [
```
HashidaTKS, line 26: I think you told me what checks are before, but now I'm not confident with that these `checks` results are correct or not...
`-1` corresponds to `オウ`'s `ウ`, for example, right?
And `-1` means that the character is the second or subsequent character of normalized by one definition that normalizes multiple characters at once?

lib/normalizer.c
```diff
@@ -1783,6 +1783,84 @@ grn_nfkc_normalize_unify_katakana_g_sounds(grn_ctx *ctx,
...
+
+    next_char_length = (size_t)grn_charlen_(ctx, next, end, GRN_ENC_UTF8);
+    /* U+30AA KATAKANA LETTER O */
+    if (next_char_length == 3 &&
+        next[0] == 0xe3 &&
+        next[1] == 0x82 && 
+        next[2] == 0xaa) {
+      unified_buffer[(*n_unified_bytes)++] = current[0];
+      unified_buffer[(*n_unified_bytes)++] = current[1];
+      unified_buffer[(*n_unified_bytes)++] = current[2];
```
kou, line 1846:
```suggestion
      unified_buffer[(*n_unified_bytes)++] = current[2];
      (*n_unified_characters)++;
```

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
```diff
@@ -0,0 +1,129 @@
...
+      0,
+      0,
+      6,
+      0,
+      0,
+      -1,
+      0,
+      0
+    ],
+    "offsets": [
```
HashidaTKS, line 112: Also I'm not confident with that these `offsets` results are correct or not...

```
0,
0,
```

means `オ` and `ウ` of `オウ` are normalized at offset `0`?
Or, Is this a wrong result?
Have I mis-implemented this normalizer?

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
kou: The results are incorrect.
https://github.com/groonga/groonga/pull/1506#discussion_r1096872298 will fix them.

> `-1` corresponds to `オウ`'s `ウ`, for example, right?

Right.

> And does `-1` mean that the character is the second or subsequent character normalized by one definition that normalizes multiple characters at once?

No. `-1` means that the character (`ウ`) doesn't have corresponding character in the source (`オオ`). But the second `オ` is the corresponding character of `ウ`.

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
kou: Wrong result.

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
HashidaTKS: Hmm, the result applied [#1506 (comment)](https://github.com/groonga/groonga/pull/1506#discussion_r1096872298) seems wrong...
Even normalized result (`"normalized": "オウオソウオノウオモウオロウオゾウオボウオ",`) is wrong.
I will re-check it.

lib/normalizer.c
HashidaTKS: Fixed.

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
HashidaTKS: `*n_used_bytes` and `*n_used_characters` are already added first character's value, so they should have not add the first character's value again at that line.

HashidaTKS: But check result still have `-1`...

HashidaTKS: Current implementation normalized two bytes at once, that means when normalizing `コオ` -> `コウ`, we normalize not only `オ` but `コオ` it self. Is that reason why the check result of `オ` is `-1` ...?

HashidaTKS: I think in order to normalize only `オ`of `コオ`, we need to know the previous character in `grn_nfkc_normalize_unify_katakana_trailing_o`.
In order to know the previous character in `grn_nfkc_normalize_unify_katakana_trailing_o`, we need to pass data about the previous character as the `user_data` argument like https://github.com/groonga/groonga/blob/8f04e6d4425a756db6418190ffb5411374a8b884/lib/normalizer.c#L1888
Is it a good idea?

kou: Let's try.

test/command/suite/normalizers/nfkc121/unify_katakana_trailing_o.expected
HashidaTKS: I see. Thanks!

HashidaTKS:
@kou 

Thank you for your comments.
I have addressed you comments.
Would you re-review this when you have time?

lib/normalizer.c
```diff
@@ -1783,6 +1783,77 @@ grn_nfkc_normalize_unify_katakana_g_sounds(grn_ctx *ctx,
...
+  char_length = (size_t)grn_charlen_(ctx, current, end, GRN_ENC_UTF8);
+
+  *n_used_bytes = char_length;
+  *n_used_characters = 1;
+
+  if (*need_trailing_check &&
+      /* U+30AA KATAKANA LETTER O */
+      char_length == 3 &&
+      current[0] == 0xe3 &&
+      current[1] == 0x82 &&
```
kou, line 1810:
```suggestion
      current[1] == 0x82 &&
```

PR #1519, opened 2023-02-09
[WIP] Near search: modify interval calculation

HashidaTKS:
WIP

### DIFF
--- a/lib/ii.c
+++ b/lib/ii.c
@@ -17,6 +17,7 @@
 */
 
 #include "grn.h"
+#include "groonga/token.h"
 
 #include <fcntl.h>
 #include <float.h>
@@ -82,6 +83,9 @@
 #define CHUNK_SPLIT_THRESHOLD    0x60000
 
 #define MAX_N_ELEMENTS           5
+#define my_printf(...) //printf(__VA_ARGS__)
+#define my_grn_p(...) //grn_p(__VA_ARGS__)
+
 
 #ifndef S_IRUSR
 # define S_IRUSR 0400
@@ -9781,7 +9785,8 @@ grn_inline static grn_rc
 token_info_build_phrase(grn_ctx *ctx,
                         grn_ii_select_data *data,
                         const char *phrase,
-                        unsigned int phrase_len)
+                        unsigned int phrase_len,
+                        int tokenizer_mode)
 {
   token_info *ti;
   grn_rc rc = GRN_END_OF_DATA;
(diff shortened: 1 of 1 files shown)


HashidaTKS:
I think this is not a good approach to adjust n_token_infos, but I don't have any other good idea...

lib/ii.c
```diff
@@ -10217,7 +10231,7 @@ token_info_build_near_phrase(grn_ctx *ctx,
     }
 
     uint32_t n_before = data->n_token_infos;
-    rc = token_info_build_phrase(ctx, data, phrase, phrase_len);
+    rc = token_info_build_phrase(ctx, data, phrase, phrase_len, GRN_TOKENIZE_ADD);
```
kou, line 10234: We should not use `GRN_TOKENIZE_ADD` in search. It may add new tokens.

HashidaTKS: Hmm... I see.

A current problem is that when a type of tokenizer is N-gram, `n_token_infos` seems be not expected value.

The `interval` for `near` series are calculated as:

`interval` = `interval between tops of two words` - `n_token_infos of the left word` + 1.

For example, thinking about `abcdefg`.
And if `* NP "abc ef"` is executed:

`interval between tops of two words`: interval between a and e = 4
`n_token_infos of the left word`: token num of `abc` = num of `ab bc` by Bi-gram and `GET_TOKENIZE_GET` = 2

as a result, `interval` is 3.

But...  `abcdefg` is normalized as `ab bc cd de ef fg` by Bi-gram.
I think `abc` should be considered as having token `cd` for `interval` calculation because it contains part of `abc` (`c`), but `abc` is normalized `ab bc` by Bi-gram and `GET_TOKENIZE_GET` .
The reason why I use `GET_TOKENIZE_ADD` is because when `GET_TOKENIZE_ADD` specified, `abc` is normalized as `ab bc c` and the number of token matches expected. It's just that as a result, it doesn't seem to be semantically correct though...

Is there the way to tokenize `abc` to `ab bc c` by Bi-gram and `GET_TOKENIZE_GET`?
I would like to get suffix tokens shorter than N-gram's N (e.g. N = 2 when Bi-gram, and N = 3 when Tri-gram).

HashidaTKS: Or, do you have any other good idea for calculating `interval` ?
I can't think of any good ideas...

kou: We should not use `c` for search because it's redundant. It causes extra needless search. It has a performance penalty.

`grn_token_have_overlap()` may help us...

HashidaTKS:
Close because this is a no use WIP PR.

PR #1527, opened 2023-02-21
ordered_near_phrase: fix max interval caluculation with multiple tokens in a phrase

HashidaTKS:
When `max_element_intervalals` is specified and a specified word containes two or more tokens in it, the interval is miscalculated.

How to reproduce.
```
table_create Entries TABLE_NO_KEY
[[0,0.0,0.0],true]
column_create Entries content COLUMN_SCALAR Text
[[0,0.0,0.0],true]
table_create Terms TABLE_PAT_KEY ShortText   --default_tokenizer 'TokenNgram("unify_alphabet", false,                                   "unify_digit", false)'   --normalizer NormalizerNFKC121
[[0,0.0,0.0],true]
column_create Terms entries_content COLUMN_INDEX|WITH_POSITION Entries content
[[0,0.0,0.0],true]
load --table Entries
[
{"content": "abcXYZdef"},
{"content": "abebcdXYZdef"},
{"content": "abcdef"},
{"content": "defXYZabc"},
{"content": "XYZabc"},
{"content": "abc123456789def"},
{"content": "abc12345678def"},
{"content": "abc1de2def"}
]
[[0,0.0,0.0],8]
select Entries --filter 'content *ONP-1,0,3 "abc def"' --output_columns '_score, content'
[
  [
    0,
    0.0,
    0.0
  ],
  [
    [
      [
        5
      ],
      [
        [
          "_score",
          "Int32"
        ],
        [
          "content",
          "Text"
        ]
      ],
      [
        1,
        "abcXYZdef"
      ],
      [
        1,
        "abcdef"
      ],
      [
        1,
        "abc123456789def"
      ],
      [
        1,
        "abc12345678def"
      ],
      [
        1,
        "abc1de2def"
      ]
    ]
  ]
]
```

The expected result of `*ONP-1,0,3 "abc def"` is only `"abcdef"`, but the query above hits all records that contains `abc` and `def`.
This is because interval is regarded as 0.

For example, in case of target is `abcXYZdef` and query is `*ONP-1,0,3 "abc def"`.

`data->token_infos[0]`: `ab` (pos = 0, offset = 0)
`data->token_infos[1]`: `bc` (pos = 0, offset = 1)
`data->token_infos[2]`: `de` (pos = 6, offset = 0)
`n`: 2 (because `if (n > n_max_element_intervals + 1)` is satisfied.)

As a result, the interval is calculated as only `data->token_infos[1].pos - data->token_infos[0].pos = 0`.
This is because `abcXYZdef` is matched.


HashidaTKS:
The interval of `abc` and `def` for `abcdef`  should be `0`, but currently it's calculated as `3`.
However, that is out of scope of this PR.
I plan to work to resolve it with #1519.

HashidaTKS:
@kou @komainu8 

Would you review this?

lib/ii.c
```diff
@@ -11768,14 +11768,18 @@ grn_ii_select_data_check_near_element_intervals(grn_ctx *ctx,
...
     uint32_t i;
     uint32_t n = data->n_token_infos;
-    if (n > n_max_element_intervals + 1) {
-      n = n_max_element_intervals + 1;
-    }
     int32_t previous_pos = data->token_infos[0]->pos;
-    for (i = 1; i < n; i++) {
+    uint32_t n_checked_elements = 0;
+
+    for (i = 1; i < n && n_checked_elements < n_max_element_intervals; i++) {
```
kou, line 11774: We should compare the first token in the `i-1`th phrase and the first token in the `i`th phrase.
Does this change do it?

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/ngram.test
```diff
@@ -0,0 +1,22 @@
...
+load --table Entries
+[
+{"content": "abcXYZdef"},
+{"content": "abebcdXYZdef"},
+{"content": "abcdef"},
+{"content": "defXYZabc"},
+{"content": "XYZabc"},
+{"content": "abc123456789def"},
+{"content": "abc12345678def"},
+{"content": "abc1de2def"}
```
kou, line 19: Can we use more meaningful test data?
For example, if we focus on the max interval `3`, we should use border data such as interval `2`, `3` and `4` data.

lib/ii.c
HashidaTKS: Sounds great!
I didn't noticed that we have `phrase_id`.

kou: Ah, the code is buggy. `n_max_element_intervals + 1` should be `n_max_element_intervals`.

HashidaTKS: I have added a little modification about `n_phrases`, and it works fine!
Thanks!

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/ngram.test
HashidaTKS: Fixed.

HashidaTKS:
@kou 

Thank you for your comments!
I have addressed your comments, would you please re-review this?

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/ngram.expected
```diff
@@ -0,0 +1,48 @@
...
+column_create Terms entries_content COLUMN_INDEX|WITH_POSITION Entries content
+[[0,0.0,0.0],true]
+load --table Entries
+[
+{"content": "abc1def"},
+{"content": "abc12def"},
+{"content": "abc123def"}
+]
+[[0,0.0,0.0],3]
+select Entries --filter 'content *ONP-1,0,5 "abc def"' --output_columns '_score, content'
```
kou, line 16: Could you also add a separated test for `n_phrases > n_max_element_intervals` case?

HashidaTKS: This case satisfies `n_phrases > n_max_element_intervals` because `n_phrases` is 2 (`abc` and `def`), `n_max_element_intervals` is `1` (`5` of `*ONP-1,0,5`).

Did you mean `n_phrases <= n_max_element_intervals` case?

HashidaTKS: Hmm, I feel like `n_phrases > n_max_element_intervals + 1` is correct in implementation.

HashidaTKS: Added tests for `n_phrases > n_max_element_intervals + 1` and `n_phrases == n_max_element_intervals + 1` and `n_phrases < n_max_element_intervals + 1`.

HashidaTKS:
@kou

Thank you.
I have addressed your comment.

lib/ii.c
```diff
@@ -11766,21 +11766,31 @@ grn_ii_select_data_check_near_element_intervals(grn_ctx *ctx,
...
-    if (n > n_max_element_intervals + 1) {
-      n = n_max_element_intervals + 1;
-    }
+    uint32_t i_token_info;
+    uint32_t n_token_infos = data->n_token_infos;
+    uint32_t i_phrase;
+    uint32_t n_phrases = data->n_phrases;
+    if (n_phrases > n_max_element_intervals + 1) {
+      n_phrases = n_max_element_intervals;
+    }
```
kou, line 11775: Ah, we should improve variable name. How about this?

```suggestion
    uint32_t i_interval;
    uint32_t n_intervals = data->n_phrases - 1;
    if (n_intervals > n_max_element_intervals) {
      n_intervals = n_max_element_intervals;
    }
```

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_token/all.expected
```diff
@@ -0,0 +1,62 @@
+table_create Entries TABLE_NO_KEY
```
kou, line 1: Could you use `multi_tokens`?

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_token/all.expected
```diff
@@ -0,0 +1,62 @@
...
+{"content": "abc1def123ghi"},
+{"content": "abc12def1ghi"},
+{"content": "abc12def12ghi"},
+{"content": "abc12def123ghi"},
+{"content": "abc1def123ghi"},
+{"content": "abc12def123ghi"},
+{"content": "abc123def123ghi"}
+]
+[[0,0.0,0.0],9]
+select Entries --filter 'content *ONP-1,0,5|5 "abc def ghi"' --output_columns '_score, content'
```
kou, line 22: Could you use different values for max element intervals to check suitable value is really use? For example, `5|6`?

lib/ii.c
HashidaTKS: I adopted your suggestion.

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_token/all.expected
HashidaTKS: Renamed.

test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_token/all.expected
HashidaTKS: Changed to use `5|6` (and fix test cases)

HashidaTKS:
@kou 

Thanks, I have addressed your comments.

lib/ii.c
kou: Could you push it?

HashidaTKS: Sorry, I pushed it.

PR #1532, opened 2023-03-03
packages: expand the maximum value of per page of the list artifacts

komainu8:
The maximum value of per page of the list artifacts is 30 in default.
However, the number of packages of Mroonga is 35.
The number of per page of the list artifacts in default is not enough.

### DIFF
--- a/packages/packages-groonga-org-package-task.rb
+++ b/packages/packages-groonga-org-package-task.rb
@@ -154,7 +154,7 @@ class PackagesGroongaOrgPackageTask < PackageTask
                                                   workflow_file_name,
                                                   branch: branch)
     workflow_runs_response.workflow_runs.each do |workflow_run|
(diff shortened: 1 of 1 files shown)


packages/packages-groonga-org-package-task.rb
```diff
@@ -154,7 +154,7 @@ def detect_built_package_url_github_actions(target_namespace,
                                                   workflow_file_name,
                                                   branch: branch)
     workflow_runs_response.workflow_runs.each do |workflow_run|
-      artifacts_response = client.get(workflow_run.artifacts_url)
+      artifacts_response = client.get(workflow_run.artifacts_url, {per_page: 50})
```
kou, line 157:
```suggestion
      artifacts_response = client.get(workflow_run.artifacts_url, per_page: 50)
```

PR #1539, opened 2023-03-07
NormalizerNFKC*: Add tests for mixing katakana options

HashidaTKS:
Added tests for mixing katakana options.
And fixed `unify_katakana_trailing_o` in order to work fine for the following cases.

1. `ョー -> ヨー -> ヨオ -> ヨウ`
  
   `ョー` was normalized to `ヨオ` (should be `ヨウ`) because `unify_katakana_trailing_o` was applied before `unify_kana_case` and `unify_katakana_trailing_o` is not normalized `オ` after small letters.
  I have added `オ` after `ォ` or `ョ` as a target of `unify_katakana_trailing_o`.

2. `ヺー -> ヺオ -> ヺウ`

   `ヺー` was normalized to `ヺオ` because  `unify_katakana_trailing_o` didn't target `ヺ`.
   I have added `ヺ` as a target of `unify_katakana_trailing_o`.

PR #1541, opened 2023-03-08
test: fix indent of test cases and remove .reject files

HashidaTKS:
This is a follow up task of #1539


HashidaTKS:
I will merge this if all tests are passed.

kou:
I merge this because I want to change tests now.

test/command/suite/select/slices/function/sub_filter/pre_filter.expected
```diff
@@ -1,25 +0,0 @@
-table_create Users TABLE_HASH_KEY ShortText
```
HashidaTKS, line 1: Note: `pre_filter.test` doesn't exist. we don't need `pre_filter.test`  or `pre_filter.test`.

PR #1542, opened 2023-03-08
doc normalizers: Add description about newly added options

HashidaTKS:
Add description about the following options.

`unify_katakana_gu_small_sounds`
`unify_katakana_di_sound`
`unify_katakana_wo_sound`
`unify_katakana_zu_small_sounds`
`unify_katakana_du_sound`
`unify_katakana_trailing_o`
`unify_katakana_du_small_sounds`
`unify_kana_prolonged_sound_mark`
`unify_kana_hyphen`


HashidaTKS:
We need to add similar description for token_filters, but I think it is better to work about it in another PR.

HashidaTKS:
I feel like description about options are separated into a single page because all NormalizerNFKC* uses the same options...
Adding similar descriptions for each normalizerNFKCs make writing and reviewing difficult...

kou:
It makes sense.

Issue #1548, opened 2023-03-30
Multiple adjusts use wrong weight


kou:
### What happend?

If we have 2 adjusts, the latter adjust also uses the former weight.

### How to reproduce it

```text
table_create Memos TABLE_NO_KEY
column_create Memos title COLUMN_SCALAR ShortText
column_create Memos content COLUMN_SCALAR ShortText

table_create Terms TABLE_PAT_KEY ShortText \
  --normalizer NormalizerAuto \
  --default_tokenizer TokenNgram
column_create Terms memos_index COLUMN_INDEX|WITH_POSITION|WITH_SECTION \
  Memos title,content

load --table Memos
[
{"title": "PGroonga = PostgreSQL + Groonga", "content": "Very fast full text search extension."},
{"title": "PostGIS", "content": "GIS extension."}
]

select Memos \
  --query '>5title:@Groonga >3content:@full' \
  --output_columns 'title, content, _score'
```

```json
[
  [
    0,
    0.0,
    0.0
  ],
  [
    [
      [
        1
      ],
      [
        [
          "title",
          "ShortText"
        ],
        [
          "content",
          "ShortText"
        ],
        [
          "_score",
          "Int32"
        ]
      ],
      [
        "PGroonga = PostgreSQL + Groonga",
        "Very fast full text search extension.",
        12
      ]
    ]
  ]
]
```

### Expected behavior

```json
[
  [
    0,
    0.0,
    0.0
  ],
  [
    [
      [
        1
      ],
      [
        [
          "title",
          "ShortText"
        ],
        [
          "content",
          "ShortText"
        ],
        [
          "_score",
          "Int32"
        ]
      ],
      [
        "PGroonga = PostgreSQL + Groonga",
        "Very fast full text search extension.",
        8
      ]
    ]
  ]
]
```

### Environment

- OS: Debian GNU/Linux
- OS Version: sid
- Groonga version: 13.0.0

### Additional context

See also: https://github.com/pgroonga/pgroonga/issues/313#issuecomment-1488176488

PR #1554, opened 2023-04-07
expr adjust: fix invalid weight with multiple adjusts

komainu8:
GitHub: fix GH-1548

The weight value that a user specify is pushed in stack.
This value pop in `PARSE()`.
Currently, If we specify multiple weights in search query, Groonga pop the latest weight immediately after Groonga it push in stack.

Therefore, the latest weight doesn't apply result.
As a result, the previous weight is multiple applied.

In this modification, multiple weight are applied correctly by pushing weight in stack after Groonga execute `PARSE()`.


kou:
Hmm...
We should pop in `.lemon` because nest information should be processed by Lemon.

It seems that we just need the following change:

```diff
diff --git a/lib/expr.c b/lib/expr.c
index 29fa00cea..cd7a13e70 100644
--- a/lib/expr.c
+++ b/lib/expr.c
@@ -4431,8 +4434,8 @@ parse_query_accept_adjust(grn_ctx *ctx,
                           float weight)
 {
   if (!(q->flags & GRN_EXPR_QUERY_NO_SYNTAX_ERROR)) {
-    parse_query_push_weight(ctx, q, weight);
     PARSE(token);
+    parse_query_push_weight(ctx, q, weight);
     return;
   }
```

komainu8:
I've fixed it and I added tests for this modification.

kou:
Could you add why/when this problem is happen and how to fix this problem to the pull request description?

komainu8:
I added he pull request description.

PR #1557, opened 2023-05-02
doc news: add news topics for 13.0.2

komainu8:


komainu8:
Groonga 13.0.2 has already released.

PR #1568, opened 2023-05-26
highlight: use more suitable name

komainu8:
Because "cycled" has a meaning that a return to the initial position.
However, "cycled_class_tag" isn't behavior of the return to the initial position.
Therefore, we use sequential "instead" of "cycled".


kou:
We need to update tests too.

komainu8:
@kou 
Thank you for your comments.
I updated tests.

PR #1571, opened 2023-05-31
Fix typo in grnslap

abetomo:
procotol ->
protocol

The example run results in the documentation was out of date, so I have updated it.


kou:
Good catch!

FYI: We want to drop support for grnslap because there is no user for it...

PR #1573, opened 2023-06-07
ii near-phrase-family: fix interval calculation when we use additional_last_interval

komainu8:
Currently, interval calculation is wrong as below when we use additional_last_interval.

* Input: "AaaxxxBbbcdefghi."
* query: '*NP3,-1"aaa bbb .$"'

Expected:

  The number of tokens between "Aaa" and "Bbb" is 3.

Before:

  The number of tokens between "Aaa" and "Bbb" is 7.

Current implementation expecte min_without_last_token's potision shows the potition of the last character of "Aaa" and max_without_last_token's potision shows the potition of the last character of "Bbb".

However, in fact, min_without_last_token's potision shows "A"'s potition and max_without_last_token's potision shows "B"'s potition.

### DIFF
--- a/lib/ii.c
+++ b/lib/ii.c
@@ -13122,8 +13122,8 @@ grn_ii_select_data_is_matched_near_phrase_real(grn_ctx *ctx,
     return (interval <= data->max_interval);
   }
 
-  token_info *max_token_info = NULL;
   token_info *min_token_info = NULL;
+  token_info *max_token_info = NULL;
   int min_without_last_token = -1;
   int max_without_last_token = -1;
   if (data->mode == GRN_OP_ORDERED_NEAR_PHRASE_PRODUCT) {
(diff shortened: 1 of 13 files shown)


lib/ii.c
```diff
@@ -13175,9 +13179,10 @@ grn_ii_select_data_is_matched_near_phrase_real(grn_ctx *ctx,
     return true;
   }
 
-  int interval_without_last_token = max_without_last_token -
-                                    min_without_last_token -
-                                    (max_token_info->n_tokens_in_phrase - 1);
+  int interval_without_last_token =
+    max_without_last_token - min_without_last_token;
+  interval_without_last_token -= min_token_info->n_tokens_in_phrase;
+
```
kou, line 13185:
```suggestion
  int interval_without_last_token = max_without_last_token -
                                    min_without_last_token -
                                    min_token_info->n_tokens_in_phrase;
```

PR #1574, opened 2023-06-08
ci macOS: add a missing dependency package

komainu8:
This modification is for solving the following error in CI for macOX

```
-- Could NOT find Protobuf (missing: Protobuf_LIBRARIES Protobuf_INCLUDE_DIR) 
CMake Error at CMakeLists.txt:368 (find_package):
-- Configuring incomplete, errors occurred!
  Found package configuration file:

    /usr/local/lib/cmake/Arrow/ArrowConfig.cmake

  but it set Arrow_FOUND to FALSE so package "Arrow" is considered to be NOT
  FOUND.  Reason given by package:

  Arrow could not be found because dependency ProtobufAlt could not be found.
```

See: https://github.com/groonga/groonga/actions/runs/5198222266/jobs/9378606104


kou:
Why is this needed?
Please write the error message you want to solve to the pull request description.

kou:
It's not a dependency problem.
See https://github.com/apache/arrow/issues/35987 for details.

komainu8:
Thanks!

The cause of this problem is upstream's problem.
Therefore, I don't need to fix Groonga's codes.

PR #1580, opened 2023-06-15
package debian: add support for Debian 12 (bookworm)

komainu8:


packages/apt/debian-bookworm/Dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG FROM=debian:bullseye
```
kou, line 1:
```suggestion
ARG FROM=debian:bookworm
```

packages/apt/debian-bookworm/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    debconf-set-selections
+
+ARG DEBUG
+
+RUN \
+  quiet=$([ "${DEBUG}" = "yes" ] || echo "-qq") && \
+  apt update ${quiet} && \
+  apt install -y -V ${quiet} \
+    wget && \
+  wget https://packages.groonga.org/debian/groonga-apt-source-latest-bookworm.deb && \
```
kou, line 15: We need to create this in packages.groonga.org.

packages/apt/debian-bookworm/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+  apt install -y -V ${quiet} ./groonga-apt-source-latest-bookworm.deb && \
+  rm ./groonga-apt-source-latest-bullseye.deb && \
```
kou, line 17:
```suggestion
  rm ./groonga-apt-source-latest-bookworm.deb && \
```

PR #1581, opened 2023-06-16
db: optimize built-in objects reference

komainu8:
Skip lock/unlock built-in objects because built-in objects must be kept opening by the current design.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -12774,6 +12774,17 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
       if (!(vp = grn_tiny_array_at(&s->values, id))) {
         goto exit;
       }
+      /* If reference count mode enable, Groonga maintain lock of
+       * a object until the object is gone refrence.
+       * However, if we many object open,
+       * the number of lock count may be over flow.
+       *
+       * Therefore, we unload unreserved objects.
+       */
+      if (id <= GRN_N_RESERVED_TYPES && vp->ptr) {
+        res = vp->ptr;
+        goto exit;
+      }
       if (grn_enable_reference_count) {
         if (!grn_db_value_lock(ctx, id, vp, &lock)) {
           const char *name;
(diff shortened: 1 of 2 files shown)


lib/db.c
```diff
@@ -13028,8 +13028,13 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
             grn_ja_unref(ctx, &iw);
           }
           if (grn_enable_reference_count) {
-            if (!vp->ptr) {
+            if (vp->ptr->header.type == GRN_TYPE
```
kou, line 13031: We should not touch `vp->ptr->XXX` before we check whether `vp->ptr` isn't `NULL`.

lib/db.c
```diff
@@ -13028,8 +13028,13 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
...
+                || vp->ptr->header.type == GRN_EXPR) {
```
kou, line 13032: We should add a comment why we need to have special case for them.

BTW, I think that we don't need `GRN_EXPR` here because we don't use permanent `GRN_EXPR` now. Our `GRN_EXPR` is always temporal now.

kou:
We may want to check this by whether reserved ID or not instead of `GRN_TYPE` or not because reserved objects aren't unloaded:

```diff
diff --git a/lib/db.c b/lib/db.c
index e1e9705ad..7575f9314 100644
--- a/lib/db.c
+++ b/lib/db.c
@@ -12774,6 +12774,10 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
       if (!(vp = grn_tiny_array_at(&s->values, id))) {
         goto exit;
       }
+      if (id <= GRN_N_RESERVED_TYPES && vp->ptr) {
+        ret = vp->ptr;
+        goto exit;
+      }
       if (grn_enable_reference_count) {
         if (!grn_db_value_lock(ctx, id, vp, &lock)) {
           const char *name;
```

komainu8:
Thanks.
I added comment.
I checked whether reserved ID or not.

kou:
Could you update the title and description?

BTW, how did you test this?

kou:
Could you update the title and description?

BTW, how did you test this?

lib/db.c
```diff
@@ -12774,6 +12774,17 @@ grn_ctx_at(grn_ctx *ctx, grn_id id)
       if (!(vp = grn_tiny_array_at(&s->values, id))) {
         goto exit;
       }
+      /* If reference count mode enable, Groonga maintain lock of
+       * a object until the object is gone refrence.
+       * However, if we many object open,
+       * the number of lock count may be over flow.
+       *
+       * Therefore, we unload unreserved objects.
+       */
```
kou, line 12783: We can skip opened check with lock for reserved types because reserved types are opened on DB open and never closed until DB closed at least with the current design. So the object is opened (vp->ptr != NULL, vp->ptr is NULL only when no associated object), we can safely return the opened object directly without lock.

PR #1620, opened 2023-09-26
mruby windows: enable presym

komainu8:
Because the performance of Groonga degrades when presym disable.
In addition, the mruby's issues about escape has arleady fixed in upstream probably.
Therefore, we use the master branch of mruby in this PR.


kou:
Could you open this PR from your fork instead of groonga/groonga?
Because groonga/groonga's GitHub Actions capacity is full now.

Issue #1638, opened 2023-12-08
Use bool instead of grn_bool


kou:
### What is your problem?

We introduced `grn_bool` for compatibility. But we can always use `bool` on all supported environments now.

We should do this step by step. We must not open one large PR that includes all changes. We can't review it.

abetomo:
We will fix it with small changes such as file by file.

### Target files.

`grep -lr grn_bool *`

- [ ] benchmark/bench-query-optimizer.c
- [ ] benchmark/bench-range-select.c
- [ ] doc/source/reference/api/grn_ctx.rst
- [ ] doc/source/reference/api/grn_obj.rst
- [ ] doc/source/news/1.0.rst
- [ ] doc/locale/en/LC_MESSAGES/news/1.0.po
- [ ] doc/locale/ja/LC_MESSAGES/news/1.0.po
- [ ] include/groonga/plugin.h
- [ ] include/groonga/array.h
- [ ] include/groonga/expr.h
- [ ] include/groonga/time.h
- [ ] include/groonga/config.h
- [ ] include/groonga/tokenizer_query_deprecated.h
- [ ] include/groonga/id.h
- [ ] include/groonga/ii.h
- [ ] include/groonga/operator.h
- [ ] include/groonga/output.h
- [ ] include/groonga/command.h
- [ ] include/groonga/token.h
- [ ] include/groonga/table.h
- [ ] include/groonga/request_canceler.h
- [ ] include/groonga/type.h
- [ ] include/groonga/obj.h
- [ ] include/groonga/tokenizer.h
- [ ] include/groonga/cast.h
- [ ] include/groonga/groonga.h
- [ ] include/groonga/db.h
- [ ] lib/pat.c
- [ ] lib/sort.c
- [ ] lib/group.c
- [ ] lib/output.c
- [ ] lib/grn_pat.h
- [ ] lib/window_function.c
- [ ] lib/io.c
- [ ] lib/grn_ecmascript.lemon
- [ ] lib/grn_scanner.h
- [ ] lib/index_column.c
- [ ] lib/grn_ecmascript.c
- [ ] lib/expr_executor.cpp
- [ ] lib/grn_ctx_impl.h
- [ ] lib/logger.c
- [ ] lib/options.c
- [ ] lib/time.c
- [ ] lib/expr.c
- [ ] lib/grn_load.h
- [ ] lib/mrb.c
- [ ] lib/command.c
- [ ] lib/grn_util.h
- [ ] lib/grn_io.h
[...]

PR #1661, opened 2024-01-11
Change `grn_id_is_builtin` to return bool instead of grn_bool

abetomo:
Related `grn_obj_is_builtin` was also changed to return bool.


abetomo:
#1638

PR #1699, opened 2024-02-13
Add `n_workers` to output of `status`

abetomo:


lib/proc.c
```diff
@@ -722,6 +722,8 @@ proc_status(grn_ctx *ctx, int nargs, grn_obj **args, grn_user_data *user_data)
       GRN_OUTPUT_UINT64(0);
     }
   }
+  GRN_OUTPUT_CSTR("n_workers");
+  GRN_OUTPUT_INT32(grn_ctx_get_n_workers(ctx));
```
kou, line 726: We must increment `n_elements` too.

It's better that we append a new element to the last. Theoretically adding a new element into a map is backward compatible, but I'm not sure that all clients (for all output formats) are implemented as expected.

abetomo: I didn't notice `n_elements`. Thanks for the info.
I also moved `n_workers` to last.

kou:
Could you update https://groonga.org/docs/reference/commands/status.html#return-value too?

abetomo:
I have updated the return value in the documentation.

kou (commented):
Ah, we want to add `default_n_workers` not `n_workers`.

kou:
We can add `n_workers` too but `default_n_workers` is required.
(Like `command_version` and `default_command_version`.)

kou:
One of this use cases is that we use it to confirm whether the `GRN_N_WORKERS_DEFAULT` environment variable is passed correctly or not.

doc/source/reference/commands/status.rst
kou, line 1: Could you also update the example output at https://groonga.org/docs/reference/commands/status.html#return-value ?

abetomo:
I added  `default_n_workers`.

doc/source/reference/commands/status.rst
```diff
@@ -72,7 +72,10 @@ The command returns the current status as an object::
       "uptime": UPTIME,
       "version": VERSION,
       "features": FEATURES,
-      "apache_arrow": APACHE_ARROW_INFORMATION
+      "apache_arrow": APACHE_ARROW_INFORMATION,
+      "n_workers": N_WORKERS,
+      "default_n_workers": DEFAULT_N_WORKERS
+
```
kou, line 78:
```suggestion
```

doc/source/reference/commands/status.rst
```diff
@@ -188,3 +191,11 @@ values:
              "version_patch": 0,
              "version": "2.0.0"
           }
+
+   * - ``n_workers``
+     - The value of ``n_workers`` set in this context.
+     - ``0``
+
+   * - ``default_n_workers``
+     - The default ``n_workers`` value of the Groonga process.
```
kou, line 200: TODO: Add a document for the common `n_workers` parameter to https://github.com/groonga/groonga/tree/main/doc/source/reference/command like https://github.com/groonga/groonga/blob/main/doc/source/reference/command/command_version.rst (but in English) and link from here. The document must describe the `GRN_N_WORKERS_DEFAULT` environment variable too.

PR #1722, opened 2024-03-04
doc contribusion/documentation: Use `.. code-block:: console` for visiblity

otegami:


doc/source/contribution/documentation/introduction.rst
```diff
@@ -15,43 +15,53 @@ Groonga uses Sphinx_ as documentation tool.
...
 
-CentOS, Fedora::
+   % sudo apt-get install -V -y python3-pip gettext
+   % sudo pip install -r doc/requirements.txt
+   % (cd doc && bundle install)
 
-  % sudo yum install -y python-pip gettext
-  % sudo pip install -r doc/requirements.txt
-  % (cd doc && bundle install)
+CentOS, Fedora:
```
kou, line 26: We should use AlmaLinux instead of CentOS in a separated PR.

doc/source/contribution/documentation/introduction.rst
```diff
@@ -15,43 +15,53 @@ Groonga uses Sphinx_ as documentation tool.
...
+   % sudo pip install -r doc/requirements.txt
+   % (cd doc && bundle install)
+
+OS X:
+
+.. code-block:: console
+
+   % brew install python
+   % brew install gettext
+   % export PATH=`brew --prefix gettext`/bin:$PATH
```
kou, line 40: We should use `$(...)` in a separated PR.

```suggestion
   % export PATH=$(brew --prefix gettext)/bin:$PATH
```

doc/source/contribution/documentation/introduction.rst
otegami: Sure. I didn't know that the CentOS users would use the AlmaLinux because of EOL.

otegami: fix: https://github.com/groonga/groonga/pull/1723 I also fixed the name from OS X to macOS.

PR #1728, opened 2024-03-04
doc contribution/documentation: add `How to fork and clone Groonga repository` step

otegami:


doc/source/contribution/documentation/introduction.rst
```diff
@@ -6,6 +6,21 @@ Introduction
 This documentation describes about how to write, generate and manage
 Groonga documentation.
```
otegami, line 8: Before writing this section, should we add the contribution flow like [`Translation flow`](https://groonga.org/docs/contribution/documentation/i18n.html#translation-flow) on the `contribution/documentation/i18n.rst` ?

kou: It may be better...?

BTW, it seems that we have duplicated content in here and https://groonga.org/docs/contribution/documentation/i18n.html#how-to-clone-groonga-repository . We may want to unify some contents including the flow.

otegami: I think so too. So I've just created the issue https://github.com/groonga/groonga/issues/1731 about this point.
After writing the introduction document, I'm thinking I will try this issue.

PR #1729, opened 2024-03-05
ubuntu: Install Sphinx because it's used for building document

otegami:


setup.sh
```diff
@@ -69,5 +69,6 @@ esac
 case "${distribution}-${code_name}" in
   debian-*|ubuntu-*)
     ${SUDO} apt install -y -V "${package_names[@]}"
+    ${SUDO} pip install -r doc/requirements.txt
```
kou, line 72: We can't assume that the current directory is the top of the source directory here.

We need to detect the top of the source directory something like:
https://github.com/groonga/groonga/blob/76dd2e9ca2a8ed41a1087474fe2a17462be185a8/tools/parse-backtrace.sh#L11

otegami: fix: [a2f7be1](https://github.com/groonga/groonga/pull/1729/commits/a2f7be1c00e704ae67f6e4650837fcb4cd094579) Thanks, I couldn't think of that case.

otegami:
I will close this PR once because I want to solve this point later with this issue https://github.com/groonga/groonga/pull/1729.

PR #1757, opened 2024-04-10
Add a script to help load the maximum number of records for TABLE_HASH_KEY

abetomo:


tools/generate-max-records-data.rb
```diff
@@ -0,0 +1,33 @@
...
+puts(<<-COMMANDS)
+table_create Data TABLE_HASH_KEY Int64
+COMMANDS
+
+puts(<<-LOAD)
+load --table Data
+[
+LOAD
+(1..N_RECORDS).each do |id|
+  puts("{\"_key\":#{id}}")
```
kou, line 29:
```suggestion
N_RECORDS.times do |i|
  puts(",") unless i.zero?
  print("{\"_key\": #{i}}")
```

tools/generate-max-records-data.rb
```diff
@@ -0,0 +1,33 @@
...
+end
+puts(<<-LOAD)
```
kou, line 31:
```suggestion
puts
puts(<<-LOAD)
```

PR #1767, opened 2024-04-26
ja: Add grn_ja_is_empty()

abetomo:
We will add `grn_obj_is_empty` after this.
We will use it there.
`grn_obj_is_empty` is used to improve checking for empty values.


abetomo:
Formatted in #1766.

PR #1787, opened 2024-06-09
ci cmake windows: remove unnecessary step for installing red-arrow

otegami:
Because it has already been installed at the previous step.


kou:
Good catch!

PR #1793, opened 2024-06-19
Add GroongaReleaseTask class

komainu8:
Because we want to also use this class in Mroonga.


komainu8:
I close this PR.
Because this is needless.

PR #1831, opened 2024-07-26
NormalizerNFKC: add `unify_alphabet_diacritical_mark` option

otegami:
GitHub: ref GH-1755

This PR adds the `unify_alphabet_diacritical_mark` option to NormalizerNFKC.
This option will allow the removal of diacritical marks to normalize them to alphabets in the following PRs.

In this PR, we are not implementing the actual logic, but just introducing the option interface.

### DIFF
--- a/lib/grn_nfkc.h
+++ b/lib/grn_nfkc.h
@@ -35,7 +35,7 @@ typedef struct {
   grn_nfkc_compose_func compose_func;
   grn_bool include_removed_source_location;
   grn_bool report_source_offset;
-  grn_bool unify_alphabet_diacritical_mark;
+  bool unify_alphabet_diacritical_mark;
   grn_bool unify_kana;
   grn_bool unify_kana_case;
   grn_bool unify_kana_voiced_sound_mark;
--- a/lib/grn_nfkc.h
+++ b/lib/grn_nfkc.h
@@ -35,6 +35,7 @@ typedef struct {
   grn_nfkc_compose_func compose_func;
   grn_bool include_removed_source_location;
   grn_bool report_source_offset;
+  grn_bool unify_alphabet_diacritical_mark;
   grn_bool unify_kana;
   grn_bool unify_kana_case;
   grn_bool unify_kana_voiced_sound_mark;
--- a/lib/normalizer.c
+++ b/lib/normalizer.c
@@ -1254,7 +1254,7 @@ grn_nfkc_normalize_unify_stateless(grn_ctx *ctx,
   i_byte = 0;
   i_character = 0;
   while (current < end) {
-    unsigned char unified_alpha[2];
+    unsigned char unified_alphabet[2];
     unsigned char unified_kana[3];
     unsigned char unified_kana_case[3];
     unsigned char unified_kana_voiced_sound_mark[3];
@@ -1283,7 +1283,7 @@ grn_nfkc_normalize_unify_stateless(grn_ctx *ctx,
         unified_char_length == 2) {
       unifying =
         grn_nfkc_normalize_unify_alphabet_diacritical_mark(unifying,
-                                                           unified_alpha);
+                                                           unified_alphabet);
     }
 
     if (before && data->options->unify_kana &&
--- a/lib/normalizer.c
+++ b/lib/normalizer.c
@@ -849,6 +849,7 @@ grn_inline static const unsigned char *
 grn_nfkc_normalize_unify_alphabet_diacritical_mark(
   const unsigned char *utf8_char, unsigned char *unified)
 {
+  // TODO: Implement function to return `unified` instead of `utf8_char`.
   return utf8_char;
 }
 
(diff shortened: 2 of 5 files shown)


test/command/suite/normalizers/nfkc/unify_alphabet_diacritical_mark.test
```diff
@@ -0,0 +1,6 @@
+#@add-substitution /NormalizerNFKC(\d*)/ "NormalizerNFKC#{ENV['NFKC'] || '100'}" "NormalizerNFKC"
+normalize \
+  'NormalizerNFKC("unify_alphabet_diacritical_mark", true)' \
+  "ÀÁÂ" \
```
abetomo, line 4: They look like the same characters, but they are different.

abetomo (approved):

lib/grn_nfkc.h
```diff
@@ -35,6 +35,7 @@ typedef struct {
   grn_nfkc_compose_func compose_func;
   grn_bool include_removed_source_location;
   grn_bool report_source_offset;
+  grn_bool unify_alphabet_diacritical_mark;
```
kou, line 38: We want to use `bool` not `grn_bool` for newly written code. See also: #1638

```suggestion
  bool unify_alphabet_diacritical_mark;
```

lib/normalizer.c
```diff
@@ -1247,6 +1254,7 @@ grn_nfkc_normalize_unify_stateless(grn_ctx *ctx,
   i_byte = 0;
   i_character = 0;
   while (current < end) {
+    unsigned char unified_alpha[2];
```
kou, line 1257: * Could you use `alphabet` not `alpha` like other added code?
* Do we need 2 bytes for unified alphabet? I think that 1 byte is enough because an unified alphabet is an ASCII character.

lib/normalizer.c
```diff
@@ -1270,6 +1278,14 @@ grn_nfkc_normalize_unify_stateless(grn_ctx *ctx,
       char_type = data->options->char_type_func(current);
     }
 
+    if (before && data->options->unify_alphabet_diacritical_mark &&
+        GRN_CHAR_TYPE(char_type) == GRN_CHAR_ALPHA &&
+        unified_char_length == 2) {
+      unifying =
+        grn_nfkc_normalize_unify_alphabet_diacritical_mark(unifying,
+                                                           unified_alpha);
```
kou, line 1286: We also need to update `unified_char_length` when we we unify `unifying` because `unified_alpha` is 1 byte character.
```suggestion
      unifying =
        grn_nfkc_normalize_unify_alphabet_diacritical_mark(unifying,
                                                           unified_alpha);
      if (unifying == unified_alpha) {
        unified_char_length = sizeof(unified_alpha);
      }
```

lib/normalizer.c
```diff
@@ -845,6 +845,13 @@ grn_nfkc_normalize_expand(grn_ctx *ctx,
                                     "");
 }
 
+grn_inline static const unsigned char *
+grn_nfkc_normalize_unify_alphabet_diacritical_mark(
+  const unsigned char *utf8_char, unsigned char *unified)
+{
+  return utf8_char;
```
komainu8, line 852:
```suggestion
  // TODO: We should return `unified` from this function. However, this function does not implement yet.
  //              I will use `unified` instead of `utf8_char` when I implement this function.
  return utf8_char;
```

lib/grn_nfkc.h
otegami: fix: [20dff36](https://github.com/groonga/groonga/pull/1831/commits/20dff364377347d458dd94b874914a1c0df2c3e4) Thanks!

lib/normalizer.c
otegami: > Could you use alphabet not alpha like other added code?
fix: [bba09c9](https://github.com/groonga/groonga/pull/1831/commits/bba09c934036cabbc5adc58fc265470cf69ba1a0) Sure.

> Do we need 2 bytes for unified alphabet? I think that 1 byte is enough because an unified alphabet is an ASCII character.

fix: [9d05eed](https://github.com/groonga/groonga/pull/1831/commits/9d05eed1511ea76d7ca31da08a856bce0ca166c9) I misunderstood the size of ASCII characters.

```ruby
irb(main):001> 'À'.bytesize
=> 2
irb(main):002> 'A'.bytesize
=> 1
```

lib/normalizer.c
otegami: fix: [9d05eed](https://github.com/groonga/groonga/pull/1831/commits/9d05eed1511ea76d7ca31da08a856bce0ca166c9) Thanks.
📝 If the character isn't unified, we should keep the size of characters 2 bytes.

lib/normalizer.c
otegami: fix: [946d5fd](https://github.com/groonga/groonga/pull/1831/commits/946d5fdaa4230dd8aedd6957584c77071809a5eb) Thanks!

otegami:
Thank you for reviewing. I've just fixed them.

Issue #1847, opened 2024-07-31
[Document] Difference in TokenFilterStem descriptions


Watson1978:
### What happend?

In English, TokenFilterStem is described as follows:
```
If you want to use TokenFilterStem as a token filter, install groonga-token-filter-stem package.

Install groonga-token-filter-stem package:

sudo apt -V -y install groonga-token-filter-stem
```

https://groonga.org/docs/install/ubuntu.html

However, in Japanese, it is described as follows:
```
TokenFilterStemをトークンフィルターとして使いたいときはgroonga-tokenizer-filter-stemパッケージをインストールします。

groonga-token-filter-stemパッケージをインストールします。

sudo apt -V -y install groonga-token-filter-stem
```

https://groonga.org/ja/docs/install/ubuntu.html

I think it should be described as `TokenFilterStemをトークンフィルターとして使いたいときはgroonga-token-filter-stemパッケージをインストールします。`, 
because `groonga-tokenizer-filter-stem` is not distributed.

### How to reproduce it

See:
- https://groonga.org/docs/install/ubuntu.html 
- https://groonga.org/ja/docs/install/ubuntu.html

### Expected behavior

Fix document

### Environment

- OS: [e.g. Windows]
- OS Version: [e.g. 11]
- Groonga version: [e.g. 12.0.1]

PR #1849, opened 2024-08-02
doc install: Fix typo in Japanese Debian/Ubuntu install guide

Watson1978:
This patch will change to correct package name in Japanese Debian/Ubuntu install guide

Fix GH-1847


Watson1978:
@kou @otegami @abetomo @komainu8 Could you review this?

otegami (approved):
LGTM 👍🏾 

---

I have two comments about the description.

1. We add the prefix of the PR like `doc install/debian` when we edit Groonga document because it's easy to find which documents.
1. We write the reference to the issue like the following because we might use different tools to manage issues or ticket.
```
GitHub: fix GH-xx
```

komainu8 (approved):

PR #1853, opened 2024-08-02
packages/apt/env.sh.in: remove unused file

Watson1978:
Seems  env.sh.in is unused.
So, this patch will remove env.sh.in file.


Watson1978:
@kou @otegami @abetomo @komainu8 Could you review this?

otegami (commented):
Thanks! Do we need `/packages/yum/env.sh` in gitignore too?

komainu8 (approved):

.gitignore
```diff
@@ -88,7 +88,6 @@
 /packages/*/*.log
 /packages/apt/.vagrant/
 /packages/apt/build.sh
-/packages/apt/env.sh
```
kou, line 91: We don't need to remove this entry because it's still auto generated by https://github.com/apache/arrow/blob/bae2908b3142f1141941d740d42c5d403c01749f/dev/tasks/linux-packages/package-task.rb#L337-L343 .

kou (commented):
Sorry, could you rebase on main?
I changed `configure.ac` recently.

.gitignore
Watson1978: Reverted .gitignore changing.
Thanks

Watson1978:
Rebased on main.

otegami (approved):
LGTM 👍🏾

PR #1871, opened 2024-08-15
logger: Fix crash bug when log rotate is enabled

abetomo:
This bug occurred in the following conditions.

* When logging to a file
  *  `--log-path <path>`
  * `--query-log-path <path>`
* Log rotate is enabled
  * `--log-rotate-threshold-size <threshold>`
  * `--query-log-rotate-threshold-size <threshold>`
* Process ID log output is enabled
  * `--log-flags process_id`

When the log is rotated, `output->file` is closed and set to `NULL`.
The crash occurs because of `flock(fileno(output->file), LOCK_UN);` in that state.
This unlocking was only executed when outputting process ID to the log, so it occurred only when outputting process ID.

### DIFF
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -313,6 +313,9 @@ grn_logger_output_end(grn_ctx *ctx,
         output->file != stderr &&
         (output->rotate_threshold_size > 0 &&
          output->size >= output->rotate_threshold_size)) {
+      if (output->need_flock) {
+        flock(fileno(output->file), LOCK_UN);
+      }
       fclose(output->file);
       output->file = NULL;
       grn_logger_output_rotate(ctx, output);
@@ -321,7 +324,7 @@ grn_logger_output_end(grn_ctx *ctx,
     }
   }
 #ifndef _WIN32
-  if (output->need_flock) {
+  if (output->need_flock && output->file) {
     flock(fileno(output->file), LOCK_UN);
   }
 #endif
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -313,9 +313,11 @@ grn_logger_output_end(grn_ctx *ctx,
         output->file != stderr &&
         (output->rotate_threshold_size > 0 &&
          output->size >= output->rotate_threshold_size)) {
+#ifndef _WIN32
       if (output->need_flock) {
         flock(fileno(output->file), LOCK_UN);
       }
+#endif
       fclose(output->file);
       output->file = NULL;
       grn_logger_output_rotate(ctx, output);


lib/logger.c
```diff
@@ -321,7 +326,7 @@ grn_logger_output_end(grn_ctx *ctx,
     }
   }
 #ifndef _WIN32
-  if (output->need_flock) {
+  if (output->need_flock && output->file) {
```
otegami, line 329: Could we call `flock(fileno(output->file), LOCK_UN);` like the following?
Because I'm no sure whether we need to call it when `written` is <= 0 or not mm.

```c
    if (output->file != stdout &&
        output->file != stderr &&
        (output->rotate_threshold_size > 0 &&
         output->size >= output->rotate_threshold_size)) {
#ifndef _WIN32
      if (output->need_flock) {
        flock(fileno(output->file), LOCK_UN);
      }
#endif
      fclose(output->file);
      output->file = NULL;
      grn_logger_output_rotate(ctx, output);
    } else {
      fflush(output->file);
      if (output->need_flock) {
        flock(fileno(output->file), LOCK_UN);
      end
    }
```

kou: We need `flock(LOCK_UN)` even when `written == 0`.

kou: `flock(LOCK_UN)` must be balanced with `flock(LOCK_EX)`.

lib/logger.c
```diff
@@ -313,6 +313,11 @@ grn_logger_output_end(grn_ctx *ctx,
         output->file != stderr &&
         (output->rotate_threshold_size > 0 &&
          output->size >= output->rotate_threshold_size)) {
+#ifndef _WIN32
+      if (output->need_flock) {
```
komainu8, line 317: Is `&& output->file` needless in this condition?

Because if we don't check `output->file != NULL` in this condition, I think that Groonga will crash at [`flock(fileno(output->file), LOCK_UN)`](https://github.com/groonga/groonga/pull/1871/files#diff-4300374c82a4008ddcc5c3784fbba41e8f8582ff220cabb9ea8581445b9d95a3R318).

lib/logger.c
otegami: Thank you very much. I understand that we need to unlock it even if the write operation results in 0, because the file has already been locked.

otegami (approved):
LGTM 👍🏾

lib/logger.c
abetomo: I see, it seems safe to do a `NULL` check.
It seems that `fclose()` also causes a crash, so I think the fix would be something like this.

```c
if (output->file != NULL) {
#ifndef _WIN32
  if (output->need_flock) {
    flock(fileno(output->file), LOCK_UN);
  }
#endif
  fclose(output->file);
  output->file = NULL;
}
```

kou: Is `output->file == NULL` happen here?

abetomo: It should not be `NULL` here.
I thought it would be safe to check just in case, but is it unnecessary?

kou: I don't want to add a needless check to keep code simple as much as possible.

abetomo: I got it

@komainu8  I decided not to add the check since it is not `NULL` here.

komainu8: I see.
I agree with it.

komainu8 (approved):

abetomo:
Thanks!

PR #1872, opened 2024-08-15
docs: Add `\ref` to `GRN_SUCCESS`

abetomo:
It will be a link when the document is generated by Doxygen.


include/groonga/groonga.h
```diff
@@ -59,6 +59,7 @@ typedef __bf16 grn_bfloat16;
 #define GRN_FALSE  false
 
 typedef enum {
+  /// Success
```
abetomo, line 62: The generated document is this.
Screenshot from 2024-08-15 23-40-43

PR #1883, opened 2024-08-20
Remove unnecessary definition

abetomo:
The same definition is in `include/groonga/table.h`.


abetomo:
https://github.com/groonga/groonga/blob/f00b0829962907d988c0339f0db11be31b7231dd/include/groonga/table.h#L687-L691

PR #1894, opened 2024-08-23
docs: Add a description of the flags used in grn_type_create()

abetomo:


include/groonga/groonga.h
```diff
@@ -479,9 +494,10 @@ typedef uint32_t grn_column_flags;
 
 /* Don't use (0x01<<12) because it's used internally. */
 
-#define GRN_OBJ_NO_SUBREC    (0x00 << 13)
-#define GRN_OBJ_WITH_SUBREC  (0x01 << 13)
+#define GRN_OBJ_NO_SUBREC   (0x00 << 13)
+#define GRN_OBJ_WITH_SUBREC (0x01 << 13)
 
+/// variable size
```
otegami, line 500:
```suggestion
/// Variable size
```

otegami (approved):
LGTM 👍🏾

include/groonga/groonga.h
```diff
@@ -436,14 +436,29 @@ typedef uint32_t grn_column_flags;
...
+/// Unsigned integer
+/// (Used to create data types. Used to determine sorting method.)
+#define GRN_OBJ_KEY_UINT (0x00 << 3)
+/// Signed integer
+/// (Used to create data types. Used to determine sorting method.)
+#define GRN_OBJ_KEY_INT (0x01 << 3)
+/// Float
+/// (Used to create data types. Used to determine sorting method.)
+#define GRN_OBJ_KEY_FLOAT (0x02 << 3)
+/// Latitude and Longitude (\ref grn_geo_point)
```
kou, line 450:
```suggestion
/// Latitude and longitude (\ref grn_geo_point)
```

include/groonga/groonga.h
```diff
@@ -436,14 +436,29 @@ typedef uint32_t grn_column_flags;
...
+/// (Used to create data types. Used to determine sorting method.)
+#define GRN_OBJ_KEY_GEO_POINT (0x03 << 3)
+/// Semi Infinite String
+/// Used in backward matching search. (The tokens are huge.)
```
kou, line 454:
```suggestion
/// Enable semi-infinite string support.
/// This is only available with \ref GRN_OBJ_TABLE_PAT_KEY.
/// You can use efficient suffix search with this but it requires more storage/memory
/// size. Because it generates many additional data for efficient suffix search implicitly.
```

include/groonga/groonga.h
```diff
@@ -479,9 +494,10 @@ typedef uint32_t grn_column_flags;
 
 /* Don't use (0x01<<12) because it's used internally. */
 
-#define GRN_OBJ_NO_SUBREC    (0x00 << 13)
-#define GRN_OBJ_WITH_SUBREC  (0x01 << 13)
+#define GRN_OBJ_NO_SUBREC   (0x00 << 13)
+#define GRN_OBJ_WITH_SUBREC (0x01 << 13)
 
+/// Variable size
```
kou, line 500:
```suggestion
/// It shows that the key is variable size not fixed size.
/// If you're using DB API such as `grn_table_create()`, you don't need to use this.
/// You can use type object such as `ShortText` instead.
/// If you're using a low-level table such as `grn_hash` and `grn_pat` directly, you can use this to show
/// that the table uses variable size key.
```

PR #1947, opened 2024-09-05
packages ubuntu noble: unify dependency packages

otegami:
Currently, building on Launchpad during release was a one-shot process, and dealing with build failures is challenging. To mitigate this, we're setting up a CI environment that mimics the Launchpad environment for testing.

However, the dependency packages listed in control.in and Dockerfile are different, making the test ineffective. This change unifies the dependency packages in Dockerfile to match those in control.in to ensure consistency between the test and actual build environments.

### DIFF
--- a/packages/apt/ubuntu-noble/Dockerfile
+++ b/packages/apt/ubuntu-noble/Dockerfile
@@ -23,11 +23,9 @@ RUN \
     liblz4-dev \
     libmecab-dev \
     libmsgpack-dev \
-    libjemalloc-dev \
     libsimdjson-dev \
     libssl-dev \
     libstemmer-dev \
-    libthrift-dev \
     libxxhash-dev \
     libzmq3-dev \
     libzstd-dev \


otegami:
📝 
```console
$ colordiff packages/apt/ubuntu-focal/Dockerfile packages/apt/ubuntu-noble/Dockerfile
1c1
< ARG FROM=ubuntu:focal
---
> ARG FROM=ubuntu:noble
25a26
>     libjemalloc-dev \
28a30
>     libthrift-dev \
```

kou:
In general, could you use "unified diff" format not "context diff" format for readability?

FYI: You can use `diff --color` instead of `colordiff`.

abetomo:
We have similar changes in #1946 and #1943, are there likely to be others?
It may be better to have a system that automatically detects or automatically generates a list.

otegami:
> We have similar changes in https://github.com/groonga/groonga/pull/1946 and https://github.com/groonga/groonga/pull/1943, are there likely to be others?
It may be better to have a system that automatically detects or automatically generates a list.

@abetomo Thanks, I've just created a issue about this.

Issue #1817, opened 2024-07-23
docs: Generate C API documentation with Doxygen


abetomo:
### What is your problem?

The C API documentation is not well maintained.
https://groonga.org/docs/reference/api.html

It would be better to use Doxygen to generate it automatically.
https://www.doxygen.nl/

abetomo:
For documents that have been created as Sphinx documents, we will remove them once and transcribe them to a header file for each file.

### Target files

`find doc/source/reference/api/ -type f`

- [x] doc/source/reference/api/global_configurations.rst
- [x] doc/source/reference/api/grn_cache.rst
- [x] doc/source/reference/api/grn_column.rst
- [x] doc/source/reference/api/grn_command_version.rst
- [x] doc/source/reference/api/grn_content_type.rst
- [x] doc/source/reference/api/grn_ctx.rst
- [x] doc/source/reference/api/grn_db.rst
- [x] doc/source/reference/api/grn_encoding.rst
- [x] doc/source/reference/api/grn_expr.rst
- [x] doc/source/reference/api/grn_geo.rst
- [x] doc/source/reference/api/grn_hook.rst
- [x] doc/source/reference/api/grn_ii.rst
- [x] doc/source/reference/api/grn_index_cursor.rst
- [x] doc/source/reference/api/grn_info.rst
- [x] doc/source/reference/api/grn_inspect.rst
- [x] doc/source/reference/api/grn_match_escalation.rst
- [x] doc/source/reference/api/grn_obj.rst
- [x] doc/source/reference/api/grn_proc.rst
- [x] doc/source/reference/api/grn_search.rst
- [x] doc/source/reference/api/grn_table.rst
- [x] doc/source/reference/api/grn_table_cursor.rst
- [ ] doc/source/reference/api/grn_thread.rst
- [x] doc/source/reference/api/grn_type.rst
- [x] doc/source/reference/api/grn_user_data.rst
- [ ] doc/source/reference/api/overview.rst
- [ ] doc/source/reference/api/plugin.rst

abetomo:
todo: Build with CMake.

otegami:
I will take `doc/source/reference/api/grn_obj.rst`.

otegami:
@abetomo
I've just fined this part: `doc/source/reference/api/grn_obj.rst`.
May I ask you to check it on TODO list?

otegami:
I will take `doc/source/reference/api/global_configurations.rst`.

abetomo:
[...]

PR #1951, opened 2024-09-10
docs: api/grn_table: move grn_table_difference() reference to the header file

abetomo:
GitHub: GH-1817

### DIFF
--- a/include/groonga/table.h
+++ b/include/groonga/table.h
@@ -710,7 +710,7 @@ grn_table_setoperation(grn_ctx *ctx,
                        grn_obj *res,
                        grn_operator op);
 /**
- * \brief Delete duplicate records in `table1` and `table2`.
+ * \brief Delete duplicated records in `table1` and `table2`.
  *
  * \param ctx The context object.
  * \param table1 The table.
--- a/include/groonga/table.h
+++ b/include/groonga/table.h
@@ -709,6 +709,18 @@ grn_table_setoperation(grn_ctx *ctx,
                        grn_obj *table2,
                        grn_obj *res,
                        grn_operator op);
+/**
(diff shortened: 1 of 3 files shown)


include/groonga/table.h
```diff
@@ -709,6 +709,18 @@ grn_table_setoperation(grn_ctx *ctx,
...
+ * \param table1 The table.
+ * \param table2 The table.
+ * \param res1 Specify `table1`.
+ * \param res2 Specify `table2`.
+ *
+ * \return \ref GRN_SUCCESS on success, \ref GRN_INVALID_ARGUMENT on
+ *         `table1 != res1 || table2 != res2`.
+ */
 GRN_API grn_rc
 grn_table_difference(
```
otegami, line 725: Is the grn_table_difference function no longer in use? If that's the case, should we consider adding a deprecation notice?

include/groonga/table.h
```diff
@@ -709,6 +709,18 @@ grn_table_setoperation(grn_ctx *ctx,
                        grn_obj *table2,
                        grn_obj *res,
                        grn_operator op);
+/**
+ * \brief Delete duplicate records in `table1` and `table2`.
```
kou, line 713:
```suggestion
 * \brief Delete duplicated records in `table1` and `table2`.
```

include/groonga/table.h
kou: We don't need to do it so far.

We may want to do it when we add `GRN_OP_DIFFERENCE` and add support for it in `grn_table_setoperation()`. (But in-place difference can't be done with `grn_table_setoperation()` interface...)

otegami: I got it, thank you!

PR #1952, opened 2024-09-11
logger: Change grn_logger_output_close() not to be run before grn_logger_init()

abetomo:
Crash on Windows when `grn_logger_output_close()` is run before `grn_logger_init()`.
Therefore, we changed it so that it does not close() when not init().  

PGroonga uses a usage where close() is run before init().
(That is the behavior when setting the logger with `grn_logger_set()`.)


otegami (commented):
LGTM 👍🏾

PR #1964, opened 2024-09-26
docs: api/grn_table: move grn_table_setoperation() reference to the header file

abetomo:
GitHub: GH-1817

### DIFF
--- a/include/groonga/table.h
+++ b/include/groonga/table.h
@@ -748,6 +748,26 @@ grn_table_group_keys_parse(grn_ctx *ctx,
                            const char *raw_sort_keys,
                            int32_t raw_sort_keys_size,
                            uint32_t *n_keys);
+/**
+ * \brief The result of the set operation of `table1` and `table2` according to
+ *        the specification of `op` is stored in `res`.
+ *
+ * \attention The table specified in `res` will be destructed.
+ *
+ * \param ctx The context object.
+ * \param table1 The table.
+ * \param table2 The table.
+ * \param res Specify `table1` or `table2`.
+ * \param op Type of operation.
+ *           - \ref GRN_OP_OR
+ *           - \ref GRN_OP_AND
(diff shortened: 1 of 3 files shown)


include/groonga/table.h
```diff
@@ -748,6 +748,26 @@ grn_table_group_keys_parse(grn_ctx *ctx,
...
+/**
+ * \brief The result of the set operation of `table1` and `table2` according to
+ *        the specification of `op` is stored in `res`.
+ *
+ * \attention The table specified in `res` will be destructed.
+ *
+ * \param ctx The context object.
+ * \param table1 The table.
+ * \param table2 The table.
+ * \param res Specify `table1` or `table2`.
```
otegami, line 760: Can we specify the other table except for `table1` or `table2`, although the current use-cases are handling only table1 or table2?

abetomo: If it is not `table1` or `table2`, the error `GRN_INVALID_ARGUMENT` occurs.

otegami: Ahh, I see. I missed it, Thanks you!

https://github.com/groonga/groonga/blob/5fcf66bd731ae6e79806447f6582c5bfbba1849b/lib/db.c#L5211

otegami (approved):
LGTM :pray:

include/groonga/table.h
```diff
@@ -748,6 +748,26 @@ grn_table_group_keys_parse(grn_ctx *ctx,
                            const char *raw_sort_keys,
                            int32_t raw_sort_keys_size,
                            uint32_t *n_keys);
+/**
+ * \brief The result of the set operation of `table1` and `table2` according to
+ *        the specification of `op` is stored in `res`.
```
kou, line 753:
```suggestion
 * \brief Execute the set operation of `table1` and `table2` according to
 *        the specification of `op` and store the result into `res`.
```

PR #1980, opened 2024-10-03
docs: api/grn_db: move grn_db_unmap() reference to the header file

abetomo:
GitHub: GH-1817

### DIFF
--- a/include/groonga/db.h
+++ b/include/groonga/db.h
@@ -48,8 +48,6 @@ grn_db_recover(grn_ctx *ctx, grn_obj *db);
  *        Normally, this API is needless. Because resources used by opened
  *        tables and columns are managed by OS automatically.
  *
- * \note This is an experimental API.
- *
  * \attention This is a thread unsafe API. You can't touch the database while
  *            this API is running.
  *
--- a/include/groonga/db.h
+++ b/include/groonga/db.h
@@ -42,6 +42,24 @@ GRN_API void
 grn_db_touch(grn_ctx *ctx, grn_obj *db);
 GRN_API grn_rc
 grn_db_recover(grn_ctx *ctx, grn_obj *db);
+/**
+ * \brief Unmaps all opened tables and columns in the passed database.
+ *        Resources used by these opened tables and columns are freed.
+ *        Normally, this API isn't useless. Because resources used by opened
+ *        tables and columns are managed by OS automatically.
+ *
+ * \note This is an experimental API.
+ *
+ * \attention This is a thread unsafe API. You can't touch the database while
+ *             this API is running.
+ *
+ * \since 5.0.7
+ *
+ * \param ctx The context object.
+ * \param db The database to be recovered.
+ *
+ * \return \ref GRN_SUCCESS on success, the appropriate \ref grn_rc on error.
+ */
 GRN_API grn_rc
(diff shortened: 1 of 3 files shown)


include/groonga/db.h
```diff
@@ -42,6 +42,24 @@ GRN_API void
 grn_db_touch(grn_ctx *ctx, grn_obj *db);
 GRN_API grn_rc
 grn_db_recover(grn_ctx *ctx, grn_obj *db);
+/**
+ * \brief Unmaps all opened tables and columns in the passed database.
+ *        Resources used by these opened tables and columns are freed.
+ *        Normally, this API isn't useless. Because resources used by opened
```
kou, line 48:
```suggestion
 *        Normally, this API is needless. Because resources used by opened
```

include/groonga/db.h
```diff
@@ -42,6 +42,24 @@ GRN_API void
 grn_db_touch(grn_ctx *ctx, grn_obj *db);
 GRN_API grn_rc
 grn_db_recover(grn_ctx *ctx, grn_obj *db);
+/**
+ * \brief Unmaps all opened tables and columns in the passed database.
+ *        Resources used by these opened tables and columns are freed.
```
otegami, line 47: How about adding a new line here for readability?

```suggestion
 *
 *        Resources used by these opened tables and columns are freed.
```

include/groonga/db.h
```diff
@@ -42,6 +42,24 @@ GRN_API void
...
+ *        Normally, this API isn't useless. Because resources used by opened
+ *        tables and columns are managed by OS automatically.
+ *
+ * \note This is an experimental API.
+ *
+ * \attention This is a thread unsafe API. You can't touch the database while
+ *             this API is running.
```
kou, line 54:
```suggestion
 *            this API is running.
```

include/groonga/db.h
otegami: We used the double negative here so it was a little bit hard for me to understand it.
I guess we might want to express the negative meaning here judging from the following sentence.
```suggestion
 *        Normally, this API isn't necessary. Because resources used by opened
```

include/groonga/db.h
```diff
@@ -42,6 +42,24 @@ GRN_API void
 grn_db_touch(grn_ctx *ctx, grn_obj *db);
 GRN_API grn_rc
 grn_db_recover(grn_ctx *ctx, grn_obj *db);
+/**
+ * \brief Unmaps all opened tables and columns in the passed database.
+ *        Resources used by these opened tables and columns are freed.
+ *        Normally, this API is needless. Because resources used by opened
+ *        tables and columns are managed by OS automatically.
+ *
+ * \note This is an experimental API.
```
otegami, line 51: Is it still an experimental API? Maybe we can remove this one in this change.

include/groonga/db.h
abetomo: I don't have a line break because I think it is a set description with the previous sentence, but would it be easier to understand if there was a line break?

include/groonga/db.h
kou: We can mark this as a stable API.

include/groonga/db.h
otegami: Ahh I ses. I think it's good as it is,

PR #1987, opened 2024-10-07
ci macos: use macos-13 temporary

otegami:
This serves as a workaround while awaiting a fix for the rpath problem in Rroonga.
Once resolved, we should revert to using macos-latest.


kou:
Hmm. This doesn't solve anything...

Let's disable "Test: mruby" on macos-latest temporary.

PR #1990, opened 2024-10-08
docs: api/grn_db: move grn_db_(create|open) reference to the header file

abetomo:
GitHub: GH-1817


include/groonga/db.h
```diff
@@ -29,13 +29,31 @@ struct _grn_db_create_optarg {
   int n_builtin_type_names;
 };
 
+/**
+ * \brief Create the database.
```
otegami, line 33: How about using 'new' here?
```suggestion
 * \brief Create new database.
```

PR #1992, opened 2024-10-08
docs: api/grn_db: move grn_obj_db() reference to the header file

abetomo:
GitHub: GH-1817

### DIFF
--- a/include/groonga/groonga.h
+++ b/include/groonga/groonga.h
@@ -1354,8 +1354,8 @@ grn_obj_defrag(grn_ctx *ctx, grn_obj *obj, int threshold);
  * \param ctx The context object.
  * \param obj Target object.
  *
- * \return The database object to which `obj` belongs if `obj` is the database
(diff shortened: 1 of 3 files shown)


include/groonga/groonga.h
```diff
@@ -1348,6 +1348,15 @@ grn_obj_flush_only_opened(grn_ctx *ctx, grn_obj *obj);
...
 grn_obj_defrag(grn_ctx *ctx, grn_obj *obj, int threshold);
 
+/**
+ * \brief Return the database to which `obj` belongs.
+ *
+ * \param ctx The context object.
+ * \param obj Target object.
+ *
+ * \return The database object to which `obj` belongs if `obj` is the database
+ *         object, `NULL` otherwise.
```
kou, line 1358:
```suggestion
 * \return The database object to which `obj` belongs, `NULL` if `obj` doesn't belong to any database.
```

PR #1993, opened 2024-10-09
package: windows: Add `vcomp140.dll` to with-vcruntime package

abetomo:
Fix the missing DLL error.


abetomo:
Artifactが生成されたら動作確認をする。

abetomo:
I checked the startup with groonga-14.1.0-2e456be-x64-vs2019-with-vcruntime in https://github.com/groonga/groonga/actions/runs/11264935137.
There was no error about missing DLL.

PR #2038, opened 2024-10-25
request_cancel: Fix a bug that canceled request may return an invalid response

abetomo:
Sometimes the following response is received when canceling. Example of command version 3 (Incomplete JSON) (`...` is the omission of an element):

```
{
  "body": {
  "columns":[...],
  "records":[
    [...],
    "header":{
      "return_code":-77,
      ...
    }
  }
```

I found that the e5400b36db0820c8baf37bacafa7d70b67129bb9 approach did not fix it.
Bugs with invalid JSON occur during the streaming process.

In particular, `logical_range_filter` may raise an exception and not output all if `request_cancel`.
Fix `logical_range_filter` to output closing brackets in case of exception due to cancellation, so that it becomes normal JSON.

### DIFF
--- a/plugins/sharding/logical_range_filter.rb
+++ b/plugins/sharding/logical_range_filter.rb
@@ -61,6 +61,11 @@ module Groonga
             writer.close_result_set
           end
           query_logger.log(:size, ":", "output(#{n_elements})")
+        rescue Groonga::Cancel
+          unless is_first
+            writer.close_table_records
+            writer.close_result_set
+          end
         ensure
           context.close
         end
--- a/lib/ctx.cpp
+++ b/lib/ctx.cpp
@@ -2222,23 +2222,12 @@ grn_ctx_recv(grn_ctx *ctx, char **str, unsigned int *str_len, int *flags)
         ctx->errline = 0;
         ctx->errfile = NULL;
         ctx->errfunc = NULL;
-        if (ctx->rc == GRN_CANCEL) {
-          *str = NULL;
-          *str_len = 0;
-          *flags = GRN_CTX_QUIT;
-        }
       }
       goto exit;
     } else {
-      if (ctx->rc == GRN_CANCEL) {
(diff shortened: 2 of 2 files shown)


plugins/sharding/logical_range_filter.rb
```diff
@@ -61,6 +61,11 @@ def run_body(input)
             writer.close_result_set
           end
           query_logger.log(:size, ":", "output(#{n_elements})")
+        rescue Groonga::Cancel
+          unless is_first
+            writer.close_table_records
+            writer.close_result_set
+          end
```
abetomo, line 68:
```diff
               break if limit <= 0
             end
           end
-          unless is_first
-            writer.close_table_records
-            writer.close_result_set
-          end
           query_logger.log(:size, ":", "output(#{n_elements})")
-        rescue Groonga::Cancel
+        ensure
           unless is_first
             writer.close_table_records
             writer.close_result_set
           end
-        ensure
           context.close
         end
       end
```

Is it better this way?

kou: Like this?

```diff
diff --git a/plugins/sharding/logical_range_filter.rb b/plugins/sharding/logical_range_filter.rb
index b62ff1148..3508507b5 100644
--- a/plugins/sharding/logical_range_filter.rb
+++ b/plugins/sharding/logical_range_filter.rb
@@ -34,6 +34,7 @@ module Groonga
 
           n_elements = 0
           is_first = true
+          begin
           executor = Executor.new(context)
           executor.execute do |result_set|
             n_elements += result_set.size
@@ -56,11 +57,13 @@ module Groonga
               break if limit <= 0
             end
           end
+          ensure
           unless is_first
             writer.close_table_records
             writer.close_result_set
           end
           query_logger.log(:size, ":", "output(#{n_elements})")
+          end
         ensure
           context.close
         end
```

PR #2043, opened 2024-10-28
output: grn_output_table_records_v1: Separate open and content

abetomo:
GitHub: Related: GH-2031

Prepare to fix the output of `records`.
The output of the content was implemented in grn_output_table_records_content().

### DIFF
--- a/lib/output.c
+++ b/lib/output.c
@@ -2537,18 +2537,21 @@ grn_output_table_records_content(grn_ctx *ctx,
                                  grn_obj *table,
                                  grn_obj_format *format)
 {
-  grn_table_cursor *tc;
 
-  tc = grn_table_cursor_open(ctx,
-                             table,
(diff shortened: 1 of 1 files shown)


abetomo:
Should I add grn_output_result_set_content_v1() because it changes the behavior of grn_ctx_output_table_records_content()?             
https://github.com/groonga/groonga/blob/v14.0.9/lib/ctx.cpp#L2818-L2827

kou:
Does it mean that we should add `grn_output_result_set_content_v1()` instead of this change?

If so, the answer is no.

We should do this and add `grn_output_result_set_content_v1()` too. In general, `grn_output_table_*()` will not be used directly. Normal users will use `grn_output_result_set()` and advanced users will use `grn_output_result_set_{open,content,close}()`.

PR #2044, opened 2024-10-28
output: grn_output_table_records: Separate open and content

abetomo:
GitHub: Related: GH-2031, GH-2043

Prepare to fix the output of `records`.

Added grn_output_table_records_content_v1().
Also added grn_output_table_records_content_v3().

The closing process is still included in content() because it is still being separated, but we plan to separate the closing process later.
The process of executing content() in open() will also be separated.

### DIFF
--- a/lib/output.c
+++ b/lib/output.c
@@ -2537,7 +2537,6 @@ grn_output_table_records_content(grn_ctx *ctx,
                                  grn_obj *table,
                                  grn_obj_format *format)
 {
-
   if (format) {
     grn_table_cursor *tc = grn_table_cursor_open(ctx,
                                                  table,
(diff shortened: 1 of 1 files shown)


lib/output.c
```diff
@@ -2642,6 +2654,21 @@ grn_output_result_set_close_v1(grn_ctx *ctx,
...
+                                    grn_content_type output_type,
+                                    grn_obj *table,
+                                    grn_obj_format *format)
+{
+  grn_output_table_records_content(ctx, outbuf, output_type, table, format);
+  if (format) {
+    grn_output_table_records_close(ctx, outbuf, output_type);
+  } else {
+    grn_output_array_close(ctx, outbuf, output_type);
+  }
```
kou, line 2669: Do we need them?
We want to call them in `grn_output_table_records_close_v3()`.

kou: Ah, it will be done in the next PR?

abetomo: I added `grn_output_table_records_close_v*` because it can be used in the next PR.

PR #2064, opened 2024-11-05
grn_hook: Change `hld_size` to `data_size`

abetomo:
No logic changes, just renaming.

Change `hld_size` to `data_size` since it is not easy to understand.
There is no impact on compatibility since it is only used internally.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -9842,7 +9842,7 @@ grn_obj_get_hook(grn_ctx *ctx,
                  grn_obj *obj,
                  grn_hook_entry entry,
                  int offset,
-                 grn_obj *data_holder_buf)
+                 grn_obj *data_buf)
 {
   grn_obj *res = NULL;
   GRN_API_ENTER;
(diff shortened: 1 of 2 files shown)


lib/db.c
```diff
@@ -8133,7 +8133,7 @@ del_hook(grn_ctx *ctx, grn_obj *obj, grn_hook_entry entry, grn_id target_id)
     if (hook->proc) {
       continue;
     }
-    if (hook->hld_size != sizeof(grn_obj_default_set_value_hook_data)) {
+    if (hook->data_holder_size != sizeof(grn_obj_default_set_value_hook_data)) {
```
otegami, line 8136: How about simplifying `data` to something like `holder_size` or replacing it with `metadata` because I'm thinking `data` is a rather generic name here?

otegami (commented):
Thank you so much!
Do we need to change the documentation in the header ?

kou:
How about using just `data_size`?

abetomo:
名前は迷ったんですが、 `grn_obj_default_set_value_hook_data` のサイズだったりするので、シンプルに `data_size` にします。

abetomo:
> Do we need to change the documentation in the header ?

There should be no files that need to be updated at this time.

lib/db.c
```diff
@@ -9838,8 +9838,11 @@ grn_obj_get_nhooks(grn_ctx *ctx, grn_obj *obj, grn_hook_entry entry)
 }
 
 grn_obj *
-grn_obj_get_hook(
-  grn_ctx *ctx, grn_obj *obj, grn_hook_entry entry, int offset, grn_obj *hldbuf)
+grn_obj_get_hook(grn_ctx *ctx,
+                 grn_obj *obj,
+                 grn_hook_entry entry,
+                 int offset,
+                 grn_obj *data_holder_buf)
```
otegami, line 9845: How about using `data_buf` here?
```suggestion
                 grn_obj *data_buf)
```

otegami (approved):
LGTM 🍀

PR #2075, opened 2024-11-12
clang-format: add lib/grn_db.h

abetomo:


.pre-commit-config.yaml
```diff
@@ -79,8 +79,9 @@ repos:
           ?lib/grn_ctx\.hpp|
           ?lib/grn_ctx\.h|
           ?lib/grn_ctx_impl\.h|
-          ?lib/grn_distance\.h|
+          ?lib/grn_db\.h|
           ?lib/grn_distance\.hpp|
+          ?lib/grn_distance\.h|
```
abetomo, line 84: This is not related to the main topic, but I sorted around grn_db.h and found the difference, so I committed it.

PR #2098, opened 2024-11-25
docs: api/grn_ctx: move grn_ctx_set_finalizer() reference to the header file

abetomo:
GitHub: GH-1817


include/groonga/groonga.h
```diff
@@ -389,6 +389,14 @@ grn_ctx_open(int flags);
  */
 GRN_API grn_rc
 grn_ctx_close(grn_ctx *ctx);
+/**
+ * \brief Set the callback function to be called when finishing the context.
```
kou, line 393:
```suggestion
 * \brief Set the callback function to be called when the context is finished.
```

PR #2090, opened 2024-11-21
index_column_diff: increase timeout for too_many_token's tests on Windows

otegami:
In CI, `Test: HTTP: reference count: Apache Arrow: chunked` step for Windows failed as follows.
These tests are unstable using the current timeout time.

```
...
--- (expected)
+++ (actual)
@@ -6,6 +6,5 @@
 [[0,0.0,0.0],true]
 column_create Deltas index COLUMN_INDEX|WITH_POSITION PostingLists deltas
 [[0,0.0,0.0],true]
-#|w| [ii][update][one] too many postings: <Deltas.index>: <1>(1): record:<PostingLists>(1:1), n-postings:<131072>, n-discarded-postings:<1>
 index_column_diff Deltas index
 [[0,0.0,0.0],[]]

===============================================================================
[2] index_column_diff/too_many_tokens
  int64_vector                                                56.2620s [failed]
===============================================================================
===============================================================================
...
```

So we will try to extend the timeout from 60 seconds to 120 seconds for these too_many_token's tests.

PR #2124, opened 2024-12-02
index_column_diff: omit tests that use large data with HTTP

otegami:
Because HTTP chunked test is too slow with them although we extended the timeout from 60 to 120 seconds at GH-2090.


otegami:
I will check whether the CI can pass on my fork repository.

otegami:
We are still investigating the reason why Windows MSVC's CI is slow.
After investigating, we will handle this PR.

otegami:
We will close this PR because this change has already been added by this commit 983db83c0bc6ca7518a44d904490123deb6a42f3.

PR #2145, opened 2024-12-04
cmake: docs: Add `--quiet` option to sphinx-build

abetomo:
Because there is a lot of output and it is hard to find out what is causing the warning.

```
$ sphinx-build --help | grep quiet
  --quiet, -q           no output on stdout, just warnings on stderr
```


otegami (approved):
LGTM 👍🏾

PR #2159, opened 2024-12-18
test: Remove the tests with Cutter

abetomo:
Related: GitHub GH-2157

We remove tests using Cutter, because we do not maintain these tests recently and also because Cutter does not support CMake.


otegami (commented):
We can also delete the following in .gitignore.
Except for it, I think it looks nice to me! Thanks!!
```console
$ git grep test/unit
.gitignore:/test/unit/Gemfile.lock
.gitignore:/test/unit/lib/*-*.*.*/
.gitignore:/test/unit/run-test.sh.log
.gitignore:/test/unit/run-test.sh.trs
.gitignore:/test/unit/test-suite.log
.gitignore:/test/unit/tmp
```

otegami (approved):
LGTM 👍🏾 
We will delete documents later.

configure.ac
```diff
@@ -1396,21 +1376,6 @@ For OS X with Homebrew:
 fi
 
 # Check for misc.
```
kou, line 1378:
```suggestion
```

PR #2162, opened 2024-12-24
ci release package: group by distribution+version

kou:
This is for https://github.com/groonga/packages.groonga.org/blob/main/ansible/files/home/packages/tasks/repository-task.rb .


abetomo (approved):

otegami (approved):
LGTM 👍🏾

PR #2182, opened 2025-01-17
msgpack: pack: Set the empty string as it is instead of nil

abetomo:
In the case of empty string, we would like to handle it as an empty string as it is.


abetomo:
Leak with `version` option in error. #2181

lib/msgpack.c
```diff
@@ -76,9 +76,17 @@ grn_msgpack_pack_raw_internal(grn_ctx *ctx,
                               size_t value_size,
                               grn_id value_domain)
 {
-  if (value_size == 0) {
-    msgpack_pack_nil(packer);
-    return ctx->rc;
+  switch (value_domain) {
```
kou, line 79: `grn_type_id_is_text_family()`というのがあってな。。。

abetomo: ぉぉ、さっき見かけたやつだった…。

lib/msgpack.c
```diff
@@ -76,7 +76,9 @@ grn_msgpack_pack_raw_internal(grn_ctx *ctx,
                               size_t value_size,
                               grn_id value_domain)
 {
-  if (value_size == 0) {
+  /* If the length is 0, if text, set to an empty string.
```
kou, line 79:
```suggestion
  /* If the length is 0 and the given value is a text, set to an empty string.
```

PR #2210, opened 2025-02-04
docs: NormalizerNFKC: Convert from reStructuredText to Markdown

abetomo:
Added line breaks for easier viewing only in the executable example.
Others were just converted to Markdown, but the content was not changed.


abetomo:
https://github.com/groonga/groonga/actions/runs/13129301574/job/36631288078
失敗しているので確認する。

PR #2211, opened 2025-02-04
pat: add support for Float32 as key type

komainu8:
Currently, Groonga doesn't support Float32 as a key of patricia trie table.
So, Groonga can't encode Float32 value correctly when we register
Float32 value as a key of patricia trie table.

In this case, we can't get result correctly as below.

```
table_create Lexicon TABLE_PAT_KEY Float32

table_create Numbers TABLE_HASH_KEY UInt64
column_create Numbers id COLUMN_SCALAR Float32

load --table Numbers
[
["_key","id"],
[1,1.1],
[2,2.1],
[3,3.1]
]

column_create Lexicon index COLUMN_INDEX Numbers id

select Numbers --filter 'id >= 2.1' --output_pretty yes
[
  [
    0,
    1738641494.864537,
    0.002141714096069336
  ],
  [
    [
      [
        1
      ],
      [
        [
          "_id",
          "UInt32"
        ],
        [
          "_key",
          "UInt64"
        ],
        [
          "id",
          "Float32"
        ]
      ],
      [
        1,
        1,
        1.1
      ]
    ]
  ]
]
```

### DIFF
--- a/test/command/suite/select/filter/index/equal/float32.test
+++ b/test/command/suite/select/filter/index/equal/float32.test
@@ -1,3 +1,6 @@
+# Equal for floating point number may not be portable.
+# If this test is fragile on some environments, we'll remove this test.
+
 table_create Values TABLE_NO_KEY
 column_create Values number COLUMN_SCALAR Float32
 
--- a/test/command/suite/select/filter/index/equal/float32.test
+++ b/test/command/suite/select/filter/index/equal/float32.test
@@ -1,0 +1,16 @@
+table_create Values TABLE_NO_KEY
+column_create Values number COLUMN_SCALAR Float32
+
+table_create Numbers TABLE_PAT_KEY Float32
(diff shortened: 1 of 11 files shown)


test/command/suite/select/filter/index/equal/float32.test
kou, line 1: これってポータブルなのかな。。。

CIは通っているから失敗するようになったら消せばいいのかな。。。

test/command/suite/select/filter/index/equal/float32.test
```diff
@@ -0,0 +1,16 @@
+table_create Values TABLE_NO_KEY
```
kou, line 1:
```suggestion
# Equal for floating point number may not be portable.
# If this test is fragile on some environments, we'll remove this test.

table_create Values TABLE_NO_KEY
```

PR #2217, opened 2025-02-10
packages ubuntu: support Apache Arrow in packages.groonga.org

otegami:
Previously, we did not provide Groonga with Apache Arrow in the Launchpad because Launchpad could not reference Apache Arrow. By distributing from packages.groonga.org instead, we remove that restriction.

Enabling Apache Arrow as a dependency in Groonga unlocks additional features, including the recently added static parallel index building support. By adding Apache Arrow to Groonga dependencies, these features become available.

### DIFF
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -43,11 +43,9 @@ apt install -V -y \
   ${repositories_dir}/${distribution}/pool/${code_name}/${component}/*/*/*_{${architecture},all}.deb
 
 groonga --version
-if [ "${distribution}" != "ubuntu" ]; then
-  if ! groonga --version | grep -q apache-arrow; then
-    echo "Apache Arrow isn't enabled"
-    exit 1
-  fi
+if ! groonga --version | grep -q apache-arrow; then
+  echo "Apache Arrow isn't enabled"
+  exit 1
 fi
 
 # There are some problems for running arm64 tests:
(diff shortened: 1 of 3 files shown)


packages/apt/test.sh
```diff
@@ -20,7 +20,9 @@ esac
 architecture=$(dpkg --print-architecture)
 
 case "${distribution}-${code_name}" in
-  debian-bookworm|ubuntu-focal)
+  # TODO: We will enable this conditon about Ubuntu focal when we add the case
+  # which is built with mirrored Apache Arrow.
+  debian-bookworm)
```
komainu8, line 25: We have already mirrored Apache Arrow for Ubuntu focal.
See: https://packages.groonga.org/ubuntu/pool/focal/main/a/apache-arrow/

So, we should add `ubuntu-focal` into this condition.

otegami: Thanks. When we build Groonga with Apache Arrow in the Apache Arrow APT Repository, it will fail as follows.
So, if we want to add this condition, I thought we would have to add a pattern that we would build packages with mirrored Apache Arrow.

```
The following packages have unmet dependencies:
 libgroonga0 : Depends: libarrow1900 (>= 19.0.0) but it is not installable
E: Unable to correct problems, you have held broken packages.
```
ref: https://github.com/otegami/groonga/actions/runs/13235719431/job/36940316136#step:14:650

otegami: Thanks, I see. I will drop support for Ubuntu 20.04 and then restart to do this PR.

otegami: fix: https://github.com/groonga/groonga/pull/2218 I dropped the support for Ubuntu 20.04.

otegami:
Should we add the step to build Groonga without Apache Arrow for Lanchpad?
If so, I will add the step in next PR.

kou:
We don't need it.

PR #2218, opened 2025-02-12
packages ubuntu: drop support for Ubuntu 20.04 (Focal Fossa)

otegami:
Because Ubuntu 20.04 will reach EOL on April, 2025.
ref: https://ubuntu.com/about/release-cycle


packages/Rakefile
```diff
@@ -61,20 +61,8 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
...
-    case target
-    when /\Aubuntu-focal/
-      enable_h3 = "#"
-    else
-      enable_h3 = ""
-    end
-    control.gsub(/@ENABLE_H3@/, enable_h3)
-  end
-
   def apt_prepare_debian_control(control_in, target)
```
otegami, line 64: note: We should define `apt_prepare_debian_control` because it's inherited from `PackageTask`.
https://github.com/apache/arrow/blob/13940cdc7e463eff8001e05b86be62a7c0e4c189/dev/tasks/linux-packages/package-task.rb#L320-L323

kou: We don't need to define this by renaming `debian/control.in` to `debian/control`: https://github.com/apache/arrow/blob/13940cdc7e463eff8001e05b86be62a7c0e4c189/dev/tasks/linux-packages/package-task.rb#L309-L317

otegami: fix: [24a0b26](https://github.com/groonga/groonga/pull/2218/commits/24a0b26b6d76255713dcb35f8094b9d16104bd3f) Thanks you, I missed it.

abetomo (approved):

kou:
Could you update https://github.com/groonga/groonga/blob/bcde904d9c043455c7df6412700084671ef8b05f/packages/debian/libgroonga0.install#L2-L12 too?

Or should we work on it as a separated task?

otegami:
fix: [39a78c5](https://github.com/groonga/groonga/pull/2218/commits/39a78c5d0ca1803171e179cde578670cbca59d2b)
Sorry, I just missed these files. I update the required version of CMake and update them.

otegami:
Although the setup of AlmaLinux9 failed, there is no problem in my local.
It might pass after rerunning this CI.
```
incus launch images:almalinux/9 almalinux9
incus config device add almalinux9 host disk source=$PWD path=/host
incus exec almalinux9 -- /host/setup.sh
incus exec almalinux9 -- /host/build.sh /host /tmp/build /tmp/local
```

packages/debian/libgroonga0.install
```diff
@@ -1,16 +1,8 @@
...
-# usr/lib/*/groonga/scripts/*
-usr/lib/*/groonga/[gs]*/*
-# language_model.* doesn't exist with Ubuntu 20.04 becuase CMake is old.
-# We can split this when we drop support for Ubuntu 20.04.
-#
-# usr/lib/*/groonga/plugins/functions/index_column.*
-# usr/lib/*/groonga/plugins/functions/language_model.*
-usr/lib/*/groonga/plugins/functions/[il]*.*
+usr/lib/*/groonga/ggml/*
+usr/lib/*/groonga/scripts/*
```
kou, line 3: Could you keep this list in alphabetical order?

otegami: fix: [ffafd7f](https://github.com/groonga/groonga/pull/2218/commits/ffafd7f81ec54593133027d68c58db415d29989a) Sure.

PR #2219, opened 2025-02-12
rake: Automatically set version

abetomo:
We can get the next version with `String#succ`.

```
irb(main):001:0> "15.0.0".succ
=> "15.0.1"
irb(main):002:0> "15.0.9".succ
=> "15.1.0"
```

It is no longer necessary to specify in `NEW_VERSION`.


otegami (approved):
LGTM 👍🏾

PR #2235, opened 2025-02-28
doc reference regular expression: fix wrong description about index ready

komainu8:
Only "\A" and "\z" can be evaluated by index.


otegami (approved):
LGTM 👍🏾

kou:
In general, we should not use o/x for check marks...
They are natural in Japan but may not natural in other countries...

FYI: Apache Arrow's docs use: ✓ (`&check;`)

komainu8:
> In general, we should not use o/x for check marks... They are natural in Japan but may not natural in other countries...
> 
> FYI: Apache Arrow's docs use: ✓ (`&check;`)

Ah...I see.

I'll use `&check;`.

komainu8:
I used "&check;".

PR #2236, opened 2025-03-03
clang-format: add include/groonga/cache.h

abetomo:


otegami (approved):
LGTM 👍🏾

include/groonga/cache.h
```diff
@@ -42,7 +44,8 @@ GRN_API const char *grn_get_default_cache_base_path(void);
  *
  * Persistent Cache:
  * - The same cache is shared across multiple processes.
- * - Persistent cache is slightly slower than memory cache due to filesystem I/O.
+ * - Persistent cache is slightly slower than memory cache due to filesystem
+ * I/O.
```
kou, line 48: Is this rendered correctly?

Do we need to format this something like the following?

```c
 * - Persistent ...
 *   I/O.
```

abetomo: The document was rendered as expected.
But it was easier to see if the indentation was aligned, so it was aligned.

PR #2257, opened 2025-03-26
docs: api/plugin: move grn_plugin_mutex_open() reference to the header file

abetomo:
GitHub: GH-1817

### DIFF
--- a/include/groonga/plugin.h
+++ b/include/groonga/plugin.h
@@ -152,6 +152,16 @@ grn_plugin_logtrace(grn_ctx *ctx, grn_log_level level);
 
 typedef struct _grn_plugin_mutex grn_plugin_mutex;
 
+/**
+ * \brief Create a new object of \ref grn_plugin_mutex.
+ *
+ * Memory for the new object is obtained with \ref GRN_PLUGIN_MALLOC.
+ *
+ * \param ctx The context object.
+ *
+ * \return A newly created \ref grn_plugin_mutex object on success, `NULL` on
+ *         sufficient memory is not available.
+ */
 GRN_API grn_plugin_mutex *
 grn_plugin_mutex_open(grn_ctx *ctx);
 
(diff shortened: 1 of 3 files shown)


include/groonga/plugin.h
```diff
@@ -152,6 +152,16 @@ grn_plugin_logtrace(grn_ctx *ctx, grn_log_level level);
...
 
+/**
+ * \brief Create a new object of \ref grn_plugin_mutex.
+ *
+ * Memory for the new object is obtained with \ref GRN_PLUGIN_MALLOC.
+ *
+ * \param ctx The context object.
+ *
+ * \return A newly created \ref grn_plugin_mutex object on success, `NULL` on
+ *         sufficient memory is not available.
```
otegami, line 163:
```suggestion
 * \return A newly created \ref grn_plugin_mutex object on success, `NULL` if
 *         sufficient memory is not available.
```

otegami (approved):
LGTM 👍🏾

include/groonga/plugin.h
```diff
@@ -152,6 +152,16 @@ grn_plugin_logtrace(grn_ctx *ctx, grn_log_level level);
...
 
+/**
+ * \brief Create a new object of \ref grn_plugin_mutex.
+ *
+ * Memory for the new object is obtained with \ref GRN_PLUGIN_MALLOC.
+ *
+ * \param ctx The context object.
+ *
+ * \return A newly created \ref grn_plugin_mutex object on success, `NULL` if
+ *         sufficient memory is not available.
```
kou, line 163: Can we add something like "see ctx->rc for error details"?

komainu8, line 163: How about add such as the following description?
"THe new created grn_plugin_mutex object should be freed after use by calling `grn_plugin_mutex_close()`".

abetomo: https://github.com/groonga/groonga/blob/d3e0120790de90441afd33da63e7187ab0c06b26/include/groonga/cache.h#L83-L84

Added with the same wording as existing.

PR #2258, opened 2025-03-27
expr_executor: operation slash: Separate cases by uintN_t

abetomo:
Related Commit:
https://github.com/groonga/groonga/commit/138a586b597e136df33fffac30c031288c8bd949

`UInt8 / -1` and `UInt16 / -1` should also be negative.

### DIFF
--- a/lib/expr_executor.cpp
+++ b/lib/expr_executor.cpp
@@ -252,16 +252,17 @@ namespace {
     }
   }
 
-  /* To avoid the following warning, which occur in uint32_t and uint64_t.
+  /* To avoid the following warning, which occur in uintN_t.
    *
    * warning C4146: unary minus operator applied to unsigned type, result still
    * unsigned
    */
   template <typename RESULT_TYPE, typename X, typename Y>
-  std::enable_if_t<!(std::is_same_v<X, uint32_t> ||
-                     std::is_same_v<X, uint64_t>) &&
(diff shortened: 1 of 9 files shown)


lib/expr_executor.cpp
```diff
@@ -252,16 +252,17 @@ namespace {
     }
   }
 
-  /* To avoid the following warning, which occur in uint32_t and uint64_t.
+  /* To avoid the following warning, which occur in uintN_t.
    *
    * warning C4146: unary minus operator applied to unsigned type, result still
    * unsigned
```
kou, line 258: Could you update why this is needed?

The current description is the original motivation but it's not different now. We do this for `X / -1 == -X`.

lib/expr_executor.cpp
```diff
@@ -276,7 +277,8 @@ namespace {
   }
 
   template <typename RESULT_TYPE, typename X, typename Y>
-  std::enable_if_t<(std::is_same_v<X, uint32_t> ||
+  std::enable_if_t<(std::is_same_v<X, uint8_t> || std::is_same_v<X, uint16_t> ||
```
kou, line 280: Can we use `Int16` for `uint8_t` and `Int32` for `uint16_t`?

otegami (approved):
LGTM 👍🏾

PR #2266, opened 2025-04-02
cmake: Upgrade bundled Zstandard

abetomo:
Related: https://github.com/pgroonga/pgroonga/pull/726

1.5.5 is an error with CMake 4.0.0.
Upgraded to 1.5.7 which can be built with CMake 4.0.0.


kou:
The next Apache Arrow that will be released in this month will fix the CMake problems.
So we can use `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` for Apache Arrow for now.

PR #2270, opened 2025-04-08
Use `offsetof()` for Groonga record and hash entry size calculations

otegami:
GitHub: ref https://github.com/MariaDB/server/pull/3932
GitHub: GH-2273

This changes replace unsafe pointer arithmetic with `offsetof()` for calculating structure field offsets.
And they also fix the following runtime errors.

- runtime error: member access within null pointer of type 'grn_plain_hash_entry'
- runtime error: member access within null pointer of type 'grn_tiny_hash_entry'
- runtime error: member access within null pointer of type 'grn_io_hash_entry_normal'
- runtime error: member access within null pointer of type 'grn_io_hash_entry_large'
- runtime error: member access within null pointer of type 'grn_rich_hash_entry'
- runtime error: member access within null pointer of type 'grn_ii_builder_term'
- runtime error: member access within null pointer of type 'grn_rset_recinfo'

### DIFF
--- a/lib/ii.cpp
+++ b/lib/ii.cpp
@@ -15480,7 +15480,7 @@ grn_ii_builder_options_fix(grn_ii_builder_options *options)
 }
 
 #define GRN_II_BUILDER_TERM_INPLACE_SIZE                                       \
-  (sizeof(grn_ii_builder_term) - (uintptr_t)&((grn_ii_builder_term *)0)->dummy)
+  (sizeof(grn_ii_builder_term) - offsetof(grn_ii_builder_term, dummy))
 
 typedef struct {
   grn_id rid;   /* Last record ID */
--- a/lib/db.c
+++ b/lib/db.c
@@ -1506,7 +1506,7 @@ calc_rec_size(grn_table_flags flags,
       break;
     }
     *value_size =
-      (uintptr_t)GRN_RSET_SUBRECS_NTH((((grn_rset_recinfo *)0)->subrecs),
+      (uintptr_t)GRN_RSET_SUBRECS_NTH(offsetof(grn_rset_recinfo, subrecs),
                                       *subrec_size,
                                       max_n_subrecs);
   } else {
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -1775,16 +1775,16 @@ grn_io_hash_calculate_entry_size(uint32_t key_size, uint32_t value_size,
 {
   if (flags & GRN_OBJ_KEY_VAR_SIZE) {
     if (flags & GRN_OBJ_KEY_LARGE) {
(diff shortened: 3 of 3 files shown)


kou:
Could you add `Co-authored-by: Daniel Black <daniel@mariadb.org>` trailer?

Can we add a job that enables UBSAN?

otegami:
> Could you add Co-authored-by: Daniel Black <daniel@mariadb.org> trailer?

Sure. I added them.

> Can we add a job that enables UBSAN?

Sure, I will add the job before opening this PR in another PR.

abetomo (approved):

lib/ii.cpp
```diff
@@ -15480,7 +15480,7 @@ grn_ii_builder_options_fix(grn_ii_builder_options *options)
 }
 
 #define GRN_II_BUILDER_TERM_INPLACE_SIZE                                       \
-  (sizeof(grn_ii_builder_term) - (uintptr_t)&((grn_ii_builder_term *)0)->dummy)
+  (sizeof(grn_ii_builder_term) - offsetof(grn_ii_builder_term, dummy))
```
kou, line 15483: Have you checked that this doesn't change the current behavior? (`&` is used here.)

otegami: I just make sure the current test has passed. I haven’t actually checked what happens when that code is executed at runtime.

otegami: I made the small program to check it and then it works as we expected.

```c
#include <stdio.h>
#include <stddef.h>
#include <stdint.h>

typedef uint64_t grn_id;

typedef struct {
    grn_id   rid;         /* Last record ID */
    uint32_t sid;         /* Last section ID */
    /* Last position (GRN_OBJ_WITH_POSITION) or frequency. */
    uint32_t pos_or_freq;
    uint32_t offset;      /* Buffer write offset */
    uint32_t size;        /* Buffer size */
    uint32_t dummy;       /* Padding */
    uint8_t  *buf;        /* Buffer (to be freed) */
} grn_ii_builder_term;

#define ORIG_INPLACE_SIZE \
    (sizeof(grn_ii_builder_term) - (uintptr_t)&((grn_ii_builder_term *)0)->dummy)

#define NEW_INPLACE_SIZE  \
    (sizeof(grn_ii_builder_term) - offsetof(grn_ii_builder_term, dummy))

int main(void) {
    size_t orig = ORIG_INPLACE_SIZE;
    size_t neu  = NEW_INPLACE_SIZE;

    printf("original formula: %zu\n", orig);
    printf("offsetof formula: %zu\n", neu);
    return 0;
}
```
```console
$ gcc test_sample.c -o test_sample
$ ./test_sample
original formula: 16
offsetof formula: 16
```

PR #2274, opened 2025-04-10
Fix Apache Arrow output for slice and drilldown

abetomo:
(drilldown or slice) + output_type=apache-arrow + command_version=3 combination will cause a crash.

The reason is that the label is set.
So, for output_type=apache-arrow, do not set the label.

Reference:
https://github.com/groonga/groonga/blob/a112f73d9bdd1151796d3923f7524fa1dc56c609/lib/output.c#L2504-L2518

### DIFF
--- a/test/command/suite/select/command_version/3/apache_arrow/slice.expected
+++ b/test/command/suite/select/command_version/3/apache_arrow/slice.expected
@@ -1,0 +1,65 @@
+plugin_register functions/time
+[[0,0.0,0.0],true]
+table_create Tags TABLE_PAT_KEY ShortText
+[[0,0.0,0.0],true]
+table_create Memos TABLE_HASH_KEY ShortText
+[[0,0.0,0.0],true]
+column_create Memos date COLUMN_SCALAR Time
+[[0,0.0,0.0],true]
+column_create Memos tag COLUMN_SCALAR Tags
+[[0,0.0,0.0],true]
+load --table Memos
+[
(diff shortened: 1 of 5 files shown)


kou:
Could you add tests that use `output_type=apache-arrow` with `slices`/`drilldowns`?

test/command/suite/select/command_version/3/apache_arrow/slice.expected
```diff
@@ -0,0 +1,65 @@
...
+GROONGA:n_hits: 2
+	  _key	                     date
+	(utf8)	              (timestamp)
+0	Groonga is fast!	2016-05-19T12:00:00+09:00
+1	Groonga sticker!	2016-05-20T12:00:02+09:00
+========================================
+_key: timestamp[ns]
+_nsubrecs: int32
+-- metadata --
+GROONGA:n_hits: 2
```
kou, line 61: Hmm. If we just ignore labels for Apache Arrow output, users can't distinguish which record batch is for a labeled slice/drilldown...

How about adding a label as metadata? (`GROONGA:data_type = slice` + `GROONGA:label: #{label}`?)

abetomo: Current:
```
-- metadata --
GROONGA:n_hits: 2
```

After:
```
-- metadata --
GROONGA:n_hits: 2
GROONGA:data_type: slice
GROONGA:label: groonga
```

otegami:
We will fix this issue at https://github.com/groonga/groonga/pull/2399 instead of this PR.

PR #2276, opened 2025-04-14
docs: api/grn_expr move grn_expr_syntax_escape_query reference to the header file

otegami:
GitHub: GH-1817


kou (commented):
Could you remove `\see` from `grn_expr_syntax_escape()` and add a note that users should use `grn_expr_syntax_escape_query()` for query syntax in general to  `grn_expr_syntax_escape()` ?

https://github.com/groonga/groonga/blob/70d16b63c16049d51bbace56bd394e972f40bc34/include/groonga/expr.h#L154-L156

otegami:

Issue #2273, opened 2025-04-10
ci cmake: support UBSAN


otegami:
### What is your problem?

We want to run UBSAN checks in the Groonga. Currently, MariaDB bundles Mroonga, and Mroonga uses Groonga internally. When MariaDB is built with UBSAN enabled, it encounters undefined behavior issues originating from the Groonga layer, causing build errors. To prevent such issues from reaching the MariaDB distribution, it would be better for Groonga to detect undefined behavior by enabling UBSAN checks in its own CI. This will allow us to identify and fix potential issues early, ensuring higher code quality and improved stability.

otegami:
In MariaDB, they use the following options for UBSAN.
Will `no-sanitize=alignment` option suppress the `requires 8 byte alignment` errors?

https://github.com/MariaDB/server/blob/22efc2c784e1b7199fb5804e6330168277ea7dce/CMakeLists.txt#L237-L240

otegami:
There is no `no-sanitize=alignment` option in GCC.
- https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fno-sanitize_003dall

In clang, there is this option.
- https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#usage

otegami:
I tried to build Groonga using clang with these options.
I faced the follwoing build errors.

```bash
rm -rf ../groonga.ubsan && \
cmake \
  -S . \
  -B ../groonga.ubsan \
  --preset=debug-default \
  -DCMAKE_INSTALL_PREFIX=/tmp/local \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_CXX_COMPILER=clang++ \
  -DCMAKE_C_FLAGS="-fsanitize=undefined -fno-sanitize=alignment" \
  -DCMAKE_CXX_FLAGS="-fsanitize=undefined" &&\
cmake --build ../groonga.ubsan
[397/398] Linking C executable src/suggest/groonga-suggest-create-dataset
FAILED: src/suggest/groonga-suggest-create-dataset 
[...]

PR #2277, opened 2025-04-14
ci make: support UBSAN

otegami:
GitHub: GH-2273
GitHub: ref https://github.com/groonga/groonga/pull/2270#issuecomment-2785257730

MariaDB bundles Groonga. With the recent UBSAN improvements in MariaDB,
UBSAN errors are detected in the Groonga layer. To catch undefined behavior
issues and ensure robust integration, this change adds a new UBSAN build configuration.

## For reviewers

This UBSAN support is implemented only with Clang because GCC doesn't support the `-fno-sanitize=alignment` option,
leading to numerous misalignment errors. And also, MariaDB is using this option too.

ref: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#usage
ref: https://github.com/MariaDB/server/blob/22efc2c784e1b7199fb5804e6330168277ea7dce/CMakeLists.txt#L239

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -194,7 +194,7 @@ jobs:
           key: red-datasets-ubuntu
       - name: "Test: command line"
         if: env.GRN_TEST_RUN_COMMAND_LINE == 'yes'
-        continue-on-error: true
+        continue-on-error: ${{ matrix.ubsan == 'ON' }}
         run: |
           test/command_line/run-test.rb --groonga-install-prefix=$PWD/install
       - name: "Test: mruby"
@@ -202,7 +202,7 @@ jobs:
         run: |
           USE_SYSTEM=yes test/mruby/run-test.rb
       - name: "Test: stdio"
-        continue-on-error: true
+        continue-on-error: ${{ matrix.ubsan == 'ON' }}
         run: |
           grntest \
             --base-dir test/command \
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -168,11 +168,12 @@ jobs:
       - name: Set environment variables
         run: |
           echo "COLUMNS=79" >> ${GITHUB_ENV}
-          echo "LD_LIBRARY_PATH=$PWD/install/lib" >> ${GITHUB_ENV}
-          if [ "${{ matrix.ubsan }}" == "ON" ] ; then
+          if [ "${{ matrix.ubsan }}" == "ON" ]; then
             CLANG_VER=$(clang -dumpversion | cut -d. -f1)
             CLANG_LIBRARY_PATH="/usr/lib/clang/${CLANG_VER}/lib/linux"
-            echo "LD_LIBRARY_PATH=${CLANG_LIBRARY_PATH}:${env.LD_LIBRARY_PATH}" >> ${GITHUB_ENV}
+            echo "LD_LIBRARY_PATH=$PWD/install/lib:${CLANG_LIBRARY_PATH}" >> ${GITHUB_ENV}
+          else
+            echo "LD_LIBRARY_PATH=$PWD/install/lib" >> ${GITHUB_ENV}
           fi
           echo "PKG_CONFIG_PATH=$PWD/lib/pkgconfig" >> ${GITHUB_ENV}
           echo "TZ=Asia/Tokyo" >> ${GITHUB_ENV}
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -77,9 +77,12 @@ jobs:
             mruby: "OFF"
           - cc: clang
             cxx: clang++
-          - cc: clang
-            cxx: clang++
-            ubsan: "ON"
+          # TODO: Temporarily disable UBSAN checks in our CI matrix due to
+          # current error noise. Once all UBSAN failures have been addressed,
+          # we will enable UBSAN checks.
+          # - cc: clang
+          #   cxx: clang++
+          #   ubsan: "ON"
     runs-on: ubuntu-latest
     timeout-minutes: 90
     env:
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -168,13 +168,7 @@ jobs:
       - name: Set environment variables
         run: |
           echo "COLUMNS=79" >> ${GITHUB_ENV}
-          if [ "${{ matrix.ubsan }}" == "ON" ]; then
-            CLANG_VER=$(clang -dumpversion | cut -d. -f1)
-            CLANG_LIBRARY_PATH="/usr/lib/clang/${CLANG_VER}/lib/linux"
-            echo "LD_LIBRARY_PATH=$PWD/install/lib:${CLANG_LIBRARY_PATH}" >> ${GITHUB_ENV}
-          else
-            echo "LD_LIBRARY_PATH=$PWD/install/lib" >> ${GITHUB_ENV}
-          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -153,6 +157,27 @@ jobs:
...
+            -GNinja \
+            -DCMAKE_INSTALL_PREFIX=$PWD/install \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DGRN_ALLOW_WARNING=OFF \
+            -DGRN_WITH_APACHE_ARROW=ON \
+            -DGRN_WITH_BLOSC=auto \
+            -DGRN_WITH_MRUBY=${GRN_WITH_MRUBY:-ON} \
+            -DCMAKE_C_FLAGS="-fsanitize=undefined -fno-sanitize=alignment" \
+            -DCMAKE_CXX_FLAGS="-fsanitize=undefined -fno-sanitize=alignment" \
+            -DCMAKE_EXE_LINKER_FLAGS="-L${CLANG_LIBRARY_PATH} -lclang_rt.ubsan_standalone-x86_64"
```
otegami, line 180: Should we enable UBSAN using CMake options like `-DWITH_UBSAN=ON`?

kou: Yes. We should not use raw `CMAKE_{C,CXX}_FLAGS`  and so on for it.

otegami: Sure. I will address it.

otegami: fix: [64d18ac](https://github.com/groonga/groonga/pull/2277/commits/64d18acb042510b543bf627cc98e7397120f97fc) Thanks. I've added `GRN_WITH_UBSAN` option here.

otegami:

.github/workflows/cmake.yml
```diff
@@ -163,7 +168,13 @@ jobs:
       - name: Set environment variables
         run: |
           echo "COLUMNS=79" >> ${GITHUB_ENV}
-          echo "LD_LIBRARY_PATH=$PWD/install/lib" >> ${GITHUB_ENV}
+          if [ "${{ matrix.ubsan }}" == "ON" ]; then
```
kou, line 171:
```suggestion
          if [ "${GRN_WITH_UBSAN}" = "ON" ]; then
```

* In general, environment variable is better than `${{ ... }}` because we can unify evaluation phase. (`${{ ... }}` is expanded BEFORE `run` is evaluated.)
* In general, `=` is better than `==` for `[` because `=` is POSIX compatible.

.github/workflows/cmake.yml
```diff
@@ -163,7 +168,13 @@ jobs:
...
+            CLANG_VER=$(clang -dumpversion | cut -d. -f1)
+            CLANG_LIBRARY_PATH="/usr/lib/clang/${CLANG_VER}/lib/linux"
+            echo "LD_LIBRARY_PATH=$PWD/install/lib:${CLANG_LIBRARY_PATH}" >> ${GITHUB_ENV}
```
kou, line 174: Do we really need this?

CMakeLists.txt
```diff
@@ -521,6 +521,30 @@ option(GRN_WITH_NFKC "use NFKC based UTF8 normalization." ON)
...
+    string(APPEND CMAKE_CXX_FLAGS " ${GRN_UBSAN_FLAGS}")
+    message(STATUS "UBSAN flags: ${GRN_UBSAN_FLAGS}")
+    execute_process(
+      COMMAND ${CMAKE_C_COMPILER}
+              -print-file-name=libclang_rt.ubsan_standalone-x86_64.so
+      OUTPUT_VARIABLE GRN_UBSAN_LIB_PATH
+      OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if(GRN_UBSAN_LIB_PATH)
+      string(APPEND CMAKE_EXE_LINKER_FLAGS " ${GRN_UBSAN_LIB_PATH}")
+      message(STATUS "Found UBSAN runtime library: ${GRN_UBSAN_LIB_PATH}")
```
kou, line 539: Why do we need this?
Can we remove this by `-fno-sanitize=alignment,vptr`?

.github/workflows/cmake.yml
otegami: Thank you so much!

otegami:

kou:
Can we ignore UBSAN CI failure until we fix UBSAN failures?

otegami:
> Can we ignore UBSAN CI failure until we fix UBSAN failures?

fix: [ec013d8](https://github.com/groonga/groonga/pull/2277/commits/ec013d863b4074bc57c5c590c207492616681d7f) Yes, I've disabled the UBSAN checks because these errors would be noise.

kou:
We want to enable the job but "ignore" the failure result.
Because we want to check whether our fix really fix the failure in CI.

We can use `continue-on-error` or something for it:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

otegami:
fix: [fca6d67](https://github.com/groonga/groonga/pull/2277/commits/fca6d6744bd3216dac6ecf2aa02320666d8f5d0e) I enabled the `continue-on-error` option at the steps which are run with UBSAN option.

.github/workflows/cmake.yml
```diff
@@ -189,13 +194,15 @@ jobs:
           key: red-datasets-ubuntu
       - name: "Test: command line"
         if: env.GRN_TEST_RUN_COMMAND_LINE == 'yes'
+        continue-on-error: true
```
kou, line 197: Can we use `true` only when we use UBSAN?

```suggestion
        continue-on-error: ${{ matrix.ubsan == 'ON' }}
```

otegami: fix: [6c124da](https://github.com/groonga/groonga/pull/2277/commits/6c124da31e2e807ecd5c902f0d8911f99a90a47d) Ahh I understood it. Also, I will address the same approach on Mroonga with MariaDB main CI.
ref: https://github.com/mroonga/mroonga/pull/912#issuecomment-2817614852

otegami:

CMakeLists.txt
```diff
@@ -521,6 +521,19 @@ option(GRN_WITH_NFKC "use NFKC based UTF8 normalization." ON)
 
 option(GRN_WITH_MEMORY_DEBUG "use debug memory management" OFF)
 
+option(GRN_WITH_UBSAN "Enable Undefined Behavior Sanitizer (UBSAN) checking"
+       OFF)
+if(GRN_WITH_UBSAN)
+  if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
```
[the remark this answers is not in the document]
otegami: fix: [6960b61](https://github.com/groonga/groonga/pull/2277/commits/6960b6163d1a7f9efbe678cbd14b2e695d8e05ed) I added the comment why we cannot use GCC now.

> When can we enable GCC here too?

I couldn't find GCC will support for `-fno-sanitize` options at this moment.
ref: https://gcc.gnu.org/cgi-bin/search.cgi?q=fno-sanitize&cmd=Search&m=all&s=DRP

otegami:

CMakeLists.txt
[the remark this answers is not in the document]
kou: https://github.com/apache/arrow/blob/c54b039c77dd0bfa822bc0a54c7f4ca1189e0d57/cpp/cmake_modules/san-config.cmake#L47-L51
It seems that Apache Arrow uses it.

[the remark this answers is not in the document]
otegami: Thank you so much. I will check it in the following PRs.

PR #2278, opened 2025-04-15
output: apache-arrow: Support for multiple metadata

abetomo:
Currently only one metadata can be set.
Change it so that multiple metadata can be set.

Related: GitHub GH-2274

### DIFF
--- a/lib/arrow.cpp
+++ b/lib/arrow.cpp
@@ -2012,6 +2012,7 @@ namespace grnarrow {
     StreamWriter(grn_ctx *ctx, grn_obj *bulk)
       : ctx_(ctx),
         output_(ctx, bulk),
+        metadata_(),
         schema_builder_(),
         schema_(),
         writer_(),
@@ -2036,9 +2037,8 @@ namespace grnarrow {
     void
     add_metadata(const char *key, const char *value)
     {
-      arrow::KeyValueMetadata metadata;
-      metadata.Append(key, value);
-      auto status = schema_builder_.AddMetadata(metadata);
+      metadata_.Append(key, value);
+      auto status = schema_builder_.AddMetadata(metadata_);
       if (!status.ok()) {
         std::stringstream context;
         check(ctx_,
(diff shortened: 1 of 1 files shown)


lib/arrow.cpp
```diff
@@ -2036,9 +2037,8 @@ namespace grnarrow {
     void
     add_metadata(const char *key, const char *value)
```
abetomo, line 2038: Would it be better to pass a hash of metadata?

kou: We can keep this signature. If we add the signature, caller needs to build a hash table something like `grn_hash`. It's not convenient and cheep.

lib/arrow.cpp
```diff
@@ -2036,9 +2037,8 @@ namespace grnarrow {
...
     {
-      arrow::KeyValueMetadata metadata;
-      metadata.Append(key, value);
-      auto status = schema_builder_.AddMetadata(metadata);
+      metadata_.Append(key, value);
+      auto status = schema_builder_.AddMetadata(metadata_);
```
kou, line 2041: If we call this multiple times, are the same metadata added multiple times?

```cpp
writer.add_metadata("x", "1"); // add x=1
writer.add_metadata("y", "2"); // add x=1 and y=2?
```

abetomo: `AddMetadata(metadata_)` is `Set` behavior rather than `Add`.

```c++
writer.add_metadata("x", "1");
writer.add_metadata("y", "2");
```

In that example, x=1 and y=2 are set in the metadata.
There is no such thing as two x=1's being set.

kou: Oh.

https://github.com/apache/arrow/issues/46146

kou: Could you add a comment that this is "set" not "add"?

otegami (approved):
LGTM 👍🏾

lib/arrow.cpp
```diff
@@ -2036,9 +2037,9 @@ namespace grnarrow {
     void
     add_metadata(const char *key, const char *value)
     {
-      arrow::KeyValueMetadata metadata;
-      metadata.Append(key, value);
-      auto status = schema_builder_.AddMetadata(metadata);
+      metadata_.Append(key, value);
+      // It is named AddMetadata(), but it "set" metadata, not "add" it.
```
kou, line 2041:
```suggestion
      // It is named AddMetadata(), but it "set" metadata, not "add" it.
      // This behavior may be changed.
      // See also: https://github.com/apache/arrow/issues/46146
```

PR #2284, opened 2025-04-23
tokenize: add a new option to output only tokens

komainu8:
Currently, the output of `tokenize` command show tokenized token and various attributes.
This output is useful for debugging. However, this output is hard to read.

So, we add a new option to output only tokenized tokens.
By this, the result of `tokenize` command is easy to read.


test/command/suite/tokenize/without_attributes.expected
```diff
@@ -0,0 +1,24 @@
...
+    "value": "の",
+    "value": "カラム",
+    "value": "ストア",
+    "value": "機能",
+    "value": "付き",
+    "value": "全文",
+    "value": "検索",
+    "value": "エンジン",
+    "value": "です",
+    "value": "。"
```
abetomo, line 22: このJSONだと不正なフォーマットなので、

```json
[
  "Groonga",
  "は",
  ...
  "。"
]
```

か

```json
[
  { "value": "Groonga" },
  { "value": "は" },
  ...
  { "value": "。" }
]
```

のようにする必要があります。

test/command/suite/tokenize/without_attributes.test
```diff
@@ -0,0 +1,6 @@
+#@require-feature mecab
+
+tokenize \
+  --tokenizer TokenMecab \
+  --string 'Groongaはオープンソースのカラムストア機能付き全文検索エンジンです。' \
+  --without_attributes
```
kou, line 6: In general, we should not use negative word ("without" in this case) for option name. It's difficult to understand.

See also: https://www.clear-code.com/blog/2014/2/6.html

lib/proc/proc_tokenize.c
```diff
@@ -54,6 +54,35 @@ fin_tokens(grn_ctx *ctx, grn_obj *tokens)
   GRN_OBJ_FIN(ctx, tokens);
 }
 
+static void
+output_only_tokens(grn_ctx *ctx, grn_obj *tokens, grn_obj *lexicon)
+{
+  size_t i, n_tokens, n_elements;
+  n_tokens = GRN_BULK_VSIZE(tokens) / sizeof(tokenize_token);
+  n_elements = 1;
```
abetomo, line 62:
```suggestion
  size_t i, n_tokens;
  n_tokens = GRN_BULK_VSIZE(tokens) / sizeof(tokenize_token);
```

This variable is not used.

lib/proc/proc_tokenize.c
```diff
@@ -54,6 +54,35 @@ fin_tokens(grn_ctx *ctx, grn_obj *tokens)
...
+
+  grn_ctx_output_array_open(ctx, "TOKENS", (int)n_tokens);
+  for (i = 0; i < n_tokens; i++) {
+    tokenize_token *token;
+    token = ((tokenize_token *)(GRN_BULK_HEAD(tokens))) + i;
```
abetomo, line 67:
```suggestion
    tokenize_token *token = ((tokenize_token *)(GRN_BULK_HEAD(tokens))) + i;
```

How about declaring and assigning at the same time?

lib/proc/proc_tokenize.c
```diff
@@ -54,6 +54,35 @@ fin_tokens(grn_ctx *ctx, grn_obj *tokens)
...
+
+    char value[GRN_TABLE_MAX_KEY_SIZE];
+    int value_size;
+    value_size =
+      grn_table_get_key(ctx, lexicon, token->id, value, GRN_TABLE_MAX_KEY_SIZE);
```
abetomo, line 72:
```suggestion
    int value_size =
      grn_table_get_key(ctx, lexicon, token->id, value, GRN_TABLE_MAX_KEY_SIZE);
```

kou:
Close this in favor of #2302.

komainu8:
This feature already merged by #2302.

PR #2288, opened 2025-04-28
test: update expected values

abetomo:
json gem 2.11.0 changed the format of the float output:

* Before: `1.0e-06`
* After:  `0.000001`

Update the expected file following the new format.


abetomo:
There are some tests that fail due to Apache Arrow dependencies, so we will fix that issue first.
#2292

abetomo:
The test has failed in Windows.
https://github.com/groonga/groonga/actions/runs/14747625656

It is this same issue.
https://github.com/groonga/grntest/pull/22

All other tests were passed.

PR #2293, opened 2025-04-30
test: fix expected float value

otegami:
GitHub: https://github.com/groonga/grntest/pull/21

The format of float has been changed since 2.11.0.
And then the grntest's output is changed so we should update the expected values.


otegami:
This is duplicated PR.
ref: https://github.com/groonga/groonga/pull/2288

Issue #2286, opened 2025-04-25
release: automate social media release announcements


otegami:
### What is your problem?

Whenever we release it, we still have to manually post the same summary to
social media. This slows down our release process. We automatically want to do
that on GitHub Actions workflow.

otegami:
I will ask the Groonga team to configure X tokens. After configuring it, I will close this issue.

otegami:
We got the following error when using release announcements.
```
Run VERSION=${GITHUB_REF_NAME#v} bundle exec rake release:announce
git ls-files
rake aborted!
X::Forbidden: Forbidden: Your client app is not configured with the appropriate oauth1 app permissions for this endpoint. (X::Forbidden)
/home/runner/work/groonga/groonga/vendor/bundle/ruby/3.4.0/gems/x-0.16.0/lib/x/response_parser.rb:39:in 'X::ResponseParser#parse'
/home/runner/work/groonga/groonga/vendor/bundle/ruby/3.4.0/gems/x-0.16.0/lib/x/client.rb:[11](https://github.com/groonga/groonga/actions/runs/17315337537/job/49160702322#step:10:12)9:in 'X::Client#execute_request'
/home/runner/work/groonga/groonga/vendor/bundle/ruby/3.4.0/gems/x-0.16.0/lib/x/client.rb:51:in 'X::Client#post'
/home/runner/work/groonga/groonga/Rakefile:266:in 'block (2 levels) in <top (required)>'
/home/runner/work/groonga/groonga/vendor/bundle/ruby/3.4.0/gems/rake-[13](https://github.com/groonga/groonga/actions/runs/17315337537/job/49160702322#step:10:14).3.0/exe/rake:27:in '<top (required)>'
/opt/hostedtoolcache/Ruby/3.4.5/x64/bin/bundle:25:in 'Kernel#load'
/opt/hostedtoolcache/Ruby/3.4.5/x64/bin/bundle:25:in '<main>'
Tasks: TOP => release:announce
(See full trace by running task with --trace)
Error: Process completed with exit code 1.
```
ref: https://github.com/groonga/groonga/actions/runs/17315337537/job/49160702322

PR #2294, opened 2025-04-30
release: post announce to X

otegami:
GitHub: GH-2286

This change will post the release summary from News (Release notes) to X.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,3 +96,15 @@ jobs:
           echo "${LAUNCHPAD_DEPLOY_KEY}" | gpg --import
           echo "trusted-key ${LAUNCHPAD_UPLOADER_PGP_KEY}" > ~/.gnupg/gpg.conf
           rake -C packages ubuntu
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ruby
+      - name: Post Announcement
+        run: |
+          bundle install
+          VERSION=${GITHUB_REF_NAME#v} rake release:announce
+        env:
+          X_API_KEY: ${{ secrets.X_API_KEY }}
+          X_API_KEY_SECRET: ${{ secrets.X_API_KEY_SECRET }}
+          X_ACCESS_TOKEN: ${{ secrets.X_ACCESS_TOKEN }}
+          X_ACCESS_TOKEN_SECRET: ${{ secrets.X_ACCESS_TOKEN_SECRET }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -102,7 +102,6 @@ jobs:
           bundler-cache: true
       - name: Post Announcement
         run: |
-          bundle install
           VERSION=${GITHUB_REF_NAME#v} rake release:announce
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -96,3 +96,15 @@ jobs:
           echo "${LAUNCHPAD_DEPLOY_KEY}" | gpg --import
           echo "trusted-key ${LAUNCHPAD_UPLOADER_PGP_KEY}" > ~/.gnupg/gpg.conf
           rake -C packages ubuntu
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ruby
+      - name: Post Announcement
+        run: |
+          bundle install
```
kou, line 104: Can we use `bundler-cache: true` with `setup-ruby`?

Rakefile
```diff
@@ -19,6 +19,7 @@
 require "date"
 require "open-uri"
 require "tmpdir"
+require "x"
```
kou, line 22: Can we make this optional?
This isn't needed on local, right?

Rakefile
```diff
@@ -244,6 +245,16 @@ Groonga #{version} (#{latest_release_date}) has been released!
...
     ANNOUNCE
+
+    x_credentials = {
+      api_key: ENV["X_API_KEY"],
+      api_key_secret: ENV["X_API_KEY_SECRET"],
+      access_token: ENV["X_ACCESS_TOKEN"],
+      access_token_secret: ENV["X_ACCESS_TOKEN_SECRET"]
+    }
+    x_client = X::Client.new(**x_credentials)
+    tweet_body = { test: latest_release_announce }
```
kou, line 256: `text`?

```suggestion
    tweet_body = { text: latest_release_announce }
```

Rakefile
otegami: fix: [d250ee3](https://github.com/groonga/groonga/pull/2294/commits/d250ee38ae3c92d8f40e5ecb80a2c19ecd7e5c4e) Yes, we just want to use it on the release workflow

Rakefile
otegami: fix: [c8b60d2](https://github.com/groonga/groonga/pull/2294/commits/c8b60d2d0944f18dc7416e3c9634fc419b56f91c)

.github/workflows/release.yml
otegami: fix: [7f4be7b](https://github.com/groonga/groonga/pull/2294/commits/7f4be7b4a73188275f1589c3bec42e9c14959546) Yes, we can use it!

otegami:

.github/workflows/release.yml
kou: Do we still need `bundle install` with `bundler-cache: true`?

otegami: fix: [0fc5908](https://github.com/groonga/groonga/pull/2294/commits/0fc590843ca063528e3815ec66fb4cfeaaad93a9) We don't need. I misunderstood the `bundler-cache` option.
- https://github.com/ruby/setup-ruby?tab=readme-ov-file#caching-bundle-install-automatically

otegami:

PR #2295, opened 2025-04-30
Fix miscellaneous typos

qobood:
Greetings!
I've made a patch that fixes some typos, could someone look through it?
There are some Windows-related checks failing, but as far as I understand it's been like that even before my patch.


abetomo:
Thank you for your contribution!

`data/html/admin/assets/flot-4.2.1` is a copy of https://github.com/flot/flot.
Could you report this to flot and exclude it from this PR?

Also, since there are a lot of changes and it is hard to review, could you separate the PR by directory or other separable units?

qobood:
@abetomo I'm on it!

otegami:
@qobood 
> data/html/admin/assets/flot-4.2.1 is a copy of https://github.com/flot/flot.
Could you report this to flot and exclude it from this PR?

Thank you for handling it. Like the above, the file `data/html/admin/assets/jquery-ui-1.12.1/jquery-ui.js` appears to be copied from https://github.com/jquery/jquery-ui. Could you report this to the jQuery UI project and remove it from this PR?

qobood:
Greetings, @otegami!
Sure, will do everything mentioned earlier too.

packages/debian/watch
```diff
@@ -1,3 +1,3 @@
 version=3
 
-http://packages.groonga.org/source/groonga/groonga-(.*).tar.gz debian uupdate
+http://packages.groonga.org/source/groonga/groonga-(.*).tar.gz debian update
```
abetomo: Sorry for the delay in response.

https://www.debian.org/doc/manuals/maint-guide/dother.en.html#watch
It is not a typo. `uupdate` is correct.

qobood: Thank you @abetomo, my mistake.
I've reverted the change.

lib/grn_hash.h
```diff
@@ -246,7 +246,7 @@ struct _grn_hash {
   // unsigned int subrec_size;
   // grn_rec_unit record_unit;
   // grn_rec_unit subrec_unit;
-  // uint8_t arrayp;
+  // uint8_t array;
```
kou, line 249: I think that this is not a typo. I think that this means "array predicate" (is array or not).

Anyway, we can remove these comment outed codes. We will never use them. We can remove them before or after this PR.

PR #2301, opened 2025-05-01
ci msvc: remove needless configuration

kou:
This also fixes GCC 15 related errors because we don't upgrade GCC to 15 by removed configuration.

See also: https://bugs.ruby-lang.org/issues/21286


otegami (approved):
LGTM 👍🏾

PR #2302, opened 2025-05-02
tokenize: add `--output_style` option

otegami:
This change introduces a `output_style` option to the `tokenize` command.
It will get easier for users to know only tokenized tokens when they want to focus on them.

- **full** (default): preserves the existing behavior.  
- **simple**: shows only tokens like `["We","are", …]` for readability.


lib/proc/proc_tokenize.c
```diff
@@ -54,6 +54,29 @@ fin_tokens(grn_ctx *ctx, grn_obj *tokens)
   GRN_OBJ_FIN(ctx, tokens);
 }
 
+static void
+output_only_token_values(grn_ctx *ctx, grn_obj *tokens, grn_obj *lexicon)
```
abetomo, line 58: How about the function name `output_simple_tokens`?

kou: `output_tokens_simple()`?

kou:
We need update the PR title.

kou:
BTW, do we need "new" in the PR title?
(Any added option is always "new" option, right?)

otegami:
> We need update the PR title.
> BTW, do we need "new" in the PR title?
(Any added option is always "new" option, right?)

Thank you so much. I've updated the PR title.

lib/proc/proc_tokenize.c
otegami: fix: [06a08ad](https://github.com/groonga/groonga/pull/2302/commits/06a08adda05355e58ec9030905e2406d63e94783)
Thank you so much.
I follow Groonga’s post-modifier naming convention like `output_columns_xx`.

kou:
> This output is useful for debugging. However, this output is hard to read when users want to know the tokenized token values.

Really? The current output is useful for normal use and simplified output is useful for debugging, isn't it?

otegami:
I will update the description now on.

lib/proc/proc_tokenize.c
```diff
@@ -54,6 +54,29 @@ fin_tokens(grn_ctx *ctx, grn_obj *tokens)
   GRN_OBJ_FIN(ctx, tokens);
 }
 
+static void
+output_tokens_simple(grn_ctx *ctx, grn_obj *tokens, grn_obj *lexicon)
+{
+  size_t i;
+  size_t n_tokens = GRN_BULK_VSIZE(tokens) / sizeof(tokenize_token);
+
+  grn_ctx_output_array_open(ctx, "TOKEN_VALUES", (int)n_tokens);
```
kou, line 63: Why do you use `TOKEN_VALUES` not `TOKENS`?

lib/proc/proc_tokenize.c
```diff
@@ -54,6 +54,29 @@ fin_tokens(grn_ctx *ctx, grn_obj *tokens)
...
+  grn_obj key;
+  GRN_OBJ_INIT(&key, GRN_BULK, GRN_OBJ_DO_SHALLOW_COPY, lexicon->header.domain);
+  for (i = 0; i < n_tokens; i++) {
+    tokenize_token *token = ((tokenize_token *)(GRN_BULK_HEAD(tokens))) + i;
```
kou, line 67: How about defining `tokenize_token *tokenize_tokens = (tokenize_token *)GRN_BULK_HEAD(tokens);` out of `for` and use it in `for` like `tokenize_token *token = tokenize_tokens + i`?

lib/proc/proc_tokenize.c
otegami: I chose `TOKEN_VALUES` instead of `TOKENS` to make it clear that we’re outputting only the value field of each token. A full token object includes properties like `position` and `force_prefix`, but TOKEN_VALUES emphasizes that only the raw strings are shown.

```
tokenize   --tokenizer TokenMecab   --string 'Groongaはオープンソースのカラムストア機能付き全文検索エンジンです。'   --output_style full
[
  [
    0,
    0.0,
    0.0
  ],
  [
    {
      "value": "Groonga",
      "position": 0,
      "force_prefix": false,
      "force_prefix_search": false
    },
    ...
  ]
]
```

otegami: I'm rethinking of it. We don't use the `TOKEN_VALUES` word in Groonga code bases. So I think we should use the `TOKEN` and it's easy for us to understand it because `values` is `TOKEN` too.

kou: Do we need to emphasize it here? Which case is this helpful?
FYI: Array name (`TOKEN_VALUES`) is used for XML output.

I feel that it's better that we choose shorter array name (`TOKENS`) rather than longer array name (`TOKEN_VALUES`) for "simple" style.

lib/proc/proc_tokenize.c
otegami: fix: [cbbb099](https://github.com/groonga/groonga/pull/2302/commits/cbbb099920e12ac13ec7ac02fedf57c952501b37) Thanks, it looks better.

lib/proc/proc_tokenize.c
otegami: > I feel that it's better that we choose shorter array name (TOKENS) rather than longer array name (TOKEN_VALUES) for "simple" style.

Thank you so much, I agreed with your opinion.

otegami:
> Really? The current output is useful for normal use and simplified output is useful for debugging, isn't it?

I've tried to update the description now.

kou:
Do we need the same option for `table_tokenize` https://groonga.org/docs/reference/commands/table_tokenize.html ?

PR #2323, opened 2025-05-08
ci release: fix contributors generation

kou:
https://github.com/groonga/groonga/actions/runs/14903554778/job/41867597413#step:4:111

```text
$ git shortlog -sn v15.0.4..
fatal: ambiguous argument 'v15.0.4..': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          set -opipefail
+          set -o pipefail
 
           latest_news_md=$(cd doc/source/news && \
                              ls *.md | sort --human-numeric-sort | tail -n1)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -52,6 +54,8 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          set -opipefail
```
otegami, line 57:
```suggestion
          set -o pipefail
```

kou: Oh, I didn't know that we need a space here.

.github/workflows/release.yml
```diff
@@ -17,6 +17,8 @@ jobs:
     environment: release
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
otegami, line 21: 📝 Set fetch-depth: 0 to fetch all history for all branches and tags.
ref: https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4

otegami (approved):
LGTM 👍🏾

Issue #2315, opened 2025-05-07
Groonga 15.0.9


otegami:
We plan to release 15.0.9 in a few days.

guide: http://groonga.org/ja/docs/contribution/development/release.html (Japanese)

- [x] Add a release note
  - [ ] Update `doc/text/news/XX.md`
  - [ ] Translate news
- [x] Prepare announcement text
  - [ ] Announce
    - [ ] X (Japanese/English)
    - [ ] Facebook (Japanese/English)
- [ ] Update documentation
  - [ ] Update .rst
  - [ ] Translate (Update .po)
- [x] Check CI https://github.com/groonga/groonga/actions
- [x] Check Launchpad Nightly https://launchpad.net/~groonga/+archive/ubuntu/nightly/+packages
- [x] Release: `rake release NEW_RELEASE_DATE=$(date +%Y-%m-%d)`
- [x] Upload packages
  - [ ] Ubuntu (launchpad)
- [x] Tagging on groonga/groonga.org: `rake release:version:update`
- [x] Announce
  - [x] X (Japanese/English)
  - [x] Facebook (Japanese/English)
- [x] Update MSYS2 package: `packages/post-msys2.sh your/MINGW-packages/fork`
- [x] Update Docker image on groonga/docker: `./update.sh`

askdkc:
現行が15.0.4なのに急にマイナーバージョンが4から7にジャンプする件について

otegami:
GroongaとMroongaのリリースバージョンを合わせるようにしているのですが、
今回は、Mroongaが先行して何回かリリースをしていたので、それに合わせるかたちでジャンプしています。

askdkc:
なるほど
新たなトリビアを知ってしまった

PGroongaもいるけど、MroongaとGroongaはバージョン揃えてる感じなんですね〜

otegami:
Release announce is failed.
```
Run VERSION=${GITHUB_REF_NAME#v} rake release:announce
  VERSION=${GITHUB_REF_NAME#v} rake release:announce
  shell: /usr/bin/bash -e {0}
  env:
    X_API_KEY: ***
    X_API_KEY_SECRET: ***
    X_ACCESS_TOKEN: ***
    X_ACCESS_TOKEN_SECRET: ***
git ls-files
rake aborted!
LoadError: cannot load such file -- x (LoadError)
[...]

PR #2324, opened 2025-05-08
ci release: fix gem load error

otegami:
GitHub: GH-2315

## Problem

In release workflow, the following error happened
because gem `x` isn't loaded.

```
Run VERSION=${GITHUB_REF_NAME#v} rake release:announce
  VERSION=${GITHUB_REF_NAME#v} rake release:announce
  shell: /usr/bin/bash -e {0}
  env:
    X_API_KEY: ***
    X_API_KEY_SECRET: ***
    X_ACCESS_TOKEN: ***
    X_ACCESS_TOKEN_SECRET: ***
git ls-files
rake aborted!
LoadError: cannot load such file -- x (LoadError)
<internal:/opt/hostedtoolcache/Ruby/3.4.3/x64/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
<internal:/opt/hostedtoolcache/Ruby/3.4.3/x64/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
/home/runner/work/groonga/groonga/Rakefile:223:in 'block (2 levels) in <top (required)>'
Tasks: TOP => release:announce
(See full trace by running task with --trace)
```

## Solution

Using `bundle exec` loads gems including gem `x` in Gemfile.
And also we added the dry-run mode to check whether the announcement works or not.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,15 +139,12 @@ jobs:
           echo "trusted-key ${LAUNCHPAD_UPLOADER_PGP_KEY}" > ~/.gnupg/gpg.conf
           rake -C packages ubuntu
       - uses: ruby/setup-ruby@v1
-        if: |
-          github.ref_type == 'tag'
         with:
           ruby-version: ruby
           bundler-cache: true
       - name: Post Announcement
-        if: |
-          github.ref_type == 'tag'
         env:
+          DRY_RUN: ${{ github.ref_type != 'tag' }}
           X_API_KEY: ${{ secrets.X_API_KEY }}
           X_API_KEY_SECRET: ${{ secrets.X_API_KEY_SECRET }}
           X_ACCESS_TOKEN: ${{ secrets.X_ACCESS_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -153,4 +153,4 @@ jobs:
           X_ACCESS_TOKEN: ${{ secrets.X_ACCESS_TOKEN }}
           X_ACCESS_TOKEN_SECRET: ${{ secrets.X_ACCESS_TOKEN_SECRET }}
         run: |
-          VERSION=${GITHUB_REF_NAME#v} rake release:announce
+          VERSION=${GITHUB_REF_NAME#v} bundle exec rake release:announce
--- a/Rakefile
+++ b/Rakefile
@@ -246,17 +246,18 @@ Groonga #{version} (#{latest_release_date}) has been released!
 #{latest_release_note_summary}
 See: #{latest_release_url}#{latest_release_anchor}
     ANNOUNCE
-    puts latest_release_announce
 
-    unless ENV["DRY_RUN"]
-      x_credentials = {
-        api_key: ENV["X_API_KEY"],
-        api_key_secret: ENV["X_API_KEY_SECRET"],
-        access_token: ENV["X_ACCESS_TOKEN"],
-        access_token_secret: ENV["X_ACCESS_TOKEN_SECRET"]
-      }
-      x_client = X::Client.new(**x_credentials)
-      tweet_body = { text: latest_release_announce }
+    x_credentials = {
+      api_key: ENV["X_API_KEY"],
+      api_key_secret: ENV["X_API_KEY_SECRET"],
+      access_token: ENV["X_ACCESS_TOKEN"],
+      access_token_secret: ENV["X_ACCESS_TOKEN_SECRET"]
+    }
+    x_client = X::Client.new(**x_credentials)
+    tweet_body = { text: latest_release_announce }
+    if ENV["DRY_RUN"]
+      puts tweet_body
+    else
       x_client.post("tweets", tweet_body.to_json)
     end
   end
--- a/Rakefile
+++ b/Rakefile
@@ -233,8 +233,8 @@ namespace :release do
     latest_release_note_title = latest_release_note.lines.first
     latest_release_note_content = latest_release_note.lines[1..-1].join
     latest_release_note_version = latest_release_note_title[/[\d.]+/]
-    if !dry_run? && (latest_release_note_version != version)
-      raise "release note isn't written"
+    if latest_release_note_version != version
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -153,4 +153,4 @@ jobs:
           X_ACCESS_TOKEN: ${{ secrets.X_ACCESS_TOKEN }}
           X_ACCESS_TOKEN_SECRET: ${{ secrets.X_ACCESS_TOKEN_SECRET }}
         run: |
-          VERSION=${GITHUB_REF_NAME#v} rake release:announce
+          VERSION=${GITHUB_REF_NAME#v} bundle exec rake release:announce
```
kou, line 156: Can we run this step when non tag push event too to detect such problems before we release a new version.
For example, generating release note step uses this approach.

We may be able to implement a "dry run" mode and use it for non tag push event.

otegami: It sounds nice. I will implement it in this PR because we want to check whether it works or not.

otegami: fix: [d890398](https://github.com/groonga/groonga/pull/2324/commits/d8903984cf5f1ed709b4bb3aa30e1a28b3deaa6c) I added the "dry run" mode and check the generated announcement on CI.

```
Run VERSION=${GITHUB_REF_NAME#v} bundle exec rake release:announce
  VERSION=${GITHUB_REF_NAME#v} bundle exec rake release:announce
  shell: /usr/bin/bash -e {0}
  env:
    DRY_RUN: true
    X_API_KEY: 
    X_API_KEY_SECRET: 
    X_ACCESS_TOKEN: 
    X_ACCESS_TOKEN_SECRET: 
git ls-files
Groonga 2324/merge (2025-05-08) has been released! This release adds the tokenizer's option to make token inspection simpler and improves negative-division semantics for unsigned integer. See: https://groonga.org/docs/news/15.html#release-2324/merge
```

Rakefile
```diff
@@ -246,16 +246,19 @@ Groonga #{version} (#{latest_release_date}) has been released!
 #{latest_release_note_summary}
 See: #{latest_release_url}#{latest_release_anchor}
     ANNOUNCE
+    puts latest_release_announce
```
kou, line 249: Do we need this for non dry run mode, too?

Rakefile
```diff
@@ -246,16 +246,19 @@ Groonga #{version} (#{latest_release_date}) has been released!
...
-    x_client.post("tweets", tweet_body.to_json)
+    unless ENV["DRY_RUN"]
+      x_credentials = {
+        api_key: ENV["X_API_KEY"],
+        api_key_secret: ENV["X_API_KEY_SECRET"],
+        access_token: ENV["X_ACCESS_TOKEN"],
+        access_token_secret: ENV["X_ACCESS_TOKEN_SECRET"]
+      }
+      x_client = X::Client.new(**x_credentials)
+      tweet_body = { text: latest_release_announce }
```
kou, line 259: Can we extract this out of `unless ENV["DRY_RUN"]`?
We should execute our codes as much as possible to detect problems before we release a new version.

otegami: fix: [cd07398](https://github.com/groonga/groonga/pull/2324/commits/cd0739858247ef61a2d7c63e42d22a200a67fd4e) Sure. I run the codes as much as we can for testing.

Rakefile
```diff
@@ -229,7 +229,7 @@ namespace :release do
     latest_release_note_title = latest_release_note.lines.first
     latest_release_note_content = latest_release_note.lines[1..-1].join
     latest_release_note_version = latest_release_note_title[/[\d.]+/]
-    if latest_release_note_version != version
+    if !ENV["DRY_RUN"] && (latest_release_note_version != version)
```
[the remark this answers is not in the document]
otegami: fix: [e5b4d74](https://github.com/groonga/groonga/pull/2324/commits/e5b4d743bd385985ac34b95978272556169c6a7e) Although I expected `DRY_RUN: false` was given here, it was `"false" `. So I convert the given values to boolean value.

otegami:

Rakefile
```diff
@@ -61,6 +61,10 @@ def env_var(name, default=nil)
   value
 end
 
+def dry_run?
+  env_var("DRY_RUN", default="true") == "true"
```
kou, line 65: Why do we need `default=` here? It defines and assigns a local variable `default`.

```suggestion
  env_var("DRY_RUN", "true") == "true"
```

Rakefile
```diff
@@ -229,7 +233,7 @@ namespace :release do
     latest_release_note_title = latest_release_note.lines.first
     latest_release_note_content = latest_release_note.lines[1..-1].join
     latest_release_note_version = latest_release_note_title[/[\d.]+/]
-    if latest_release_note_version != version
+    if !dry_run? && (latest_release_note_version != version)
       raise "release note isn't written"
```
kou, line 237: If we put `!dry_run?` here, `latest_release_note_version != version` isn't evaluated without tag push.

```suggestion
    if latest_release_note_version != version
      raise "release note isn't written" unless dry_run
```

Rakefile
kou: BTW, is it expected that we use dry run mode by default. In general, dry run mode is disabled by default. For example, `git`, `rsync`, ...

Rakefile
otegami: fix: [6ce9642](https://github.com/groonga/groonga/pull/2324/commits/6ce9642b38f9036c343e22d27852732865b2a415) Thank you so much.

Rakefile
otegami: fix: [f74a654](https://github.com/groonga/groonga/pull/2324/commits/f74a654f72d6b52469cca1d697813066678464b0) I understood it. (I was afraid that I accidentally post announces on X during implementations. )

otegami:

abetomo (approved):

komainu8 (approved):

PR #2330, opened 2025-05-14
TokenMecab: add `internal_delimiter` option

otegami:
## Problem

User-defined dictionary entries with embedded spaces (e.g. “search engine”) are correctly recognized by MeCab itself. But when using TokenMecab in `-Owakati` mode we split on every space. As a result,
“search engine” ends up as two tokens (“search” and “engine”) instead of one.

## Cause

TokenMecab passes the `-Owakati` flag to MeCab and then treats the output as space-delimited. Any
internal spaces in dictionary entries are therefore mis-interpreted as token boundaries.

## Solution

Introduce a new `internal_delimiter` option. When set to `"tab"`, TokenMecab invokes MeCab with `-F%m\t` to emit tab-delimited tokens, and uses a tab to split the tokens. This preserves entries containing spaces as single tokens while retaining backward compatibility with the default space-delimited behavior.

### DIFF
--- a/lib/grn_str.h
+++ b/lib/grn_str.h
@@ -53,6 +53,8 @@ grn_str_len(grn_ctx *ctx,
 #define GRN_STR_CTYPE(c)   (c & 0x7f)
 
 GRN_API int
+grn_istab(const char *str);
+GRN_API int
 grn_isspace(const char *s, grn_encoding encoding);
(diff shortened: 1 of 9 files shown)


otegami:
We close this PR because this issue will be fixed in GH-2331.

PR #2331, opened 2025-05-14
TokenMecab: fix unintended splitting of user-defined words with embedded spaces

otegami:
## Problem

User-defined dictionary entries with embedded spaces (e.g. “search engine”) are correctly recognized by MeCab itself. However, in `-Owakati` mode, we split on every space. As a result, `"search engine"` ends up as two tokens (`"search"` and `"engine"`) instead of one.

## Cause

TokenMecab passes the `-Owakati` flag to MeCab and then treats the output as space-delimited. Any
internal spaces in dictionary entries are therefore mis-interpreted as token boundaries.

## Solution

Using  MeCab with `-F%m\n` to emit newline-delimited
tokens, and uses a tab to split the tokens instead of spaces.


plugins/tokenizers/mecab.c
```diff
@@ -1063,6 +1064,34 @@ mecab_next_default_format(grn_ctx *ctx,
...
+  case 0xA1:
+    if (encoding == GRN_ENC_EUC_JP && s[1] == 0xA1) {
+      return 2;
+    }
+    break;
+  case 0xE3:
+    if (encoding == GRN_ENC_UTF8 && s[1] == 0x80 && s[2] == 0x80) {
+      return 3;
+    }
+    break;
```
abetomo, line 1088:
```suggestion
```

This code is for full-width a space, so it should not be necessary.

otegami: We need this code to support the following case although the function name might be misleading.
- https://github.com/groonga/groonga/blob/main/test/command/suite/tokenizers/mecab/full_width_space/last.test

plugins/tokenizers/mecab.c
```diff
@@ -520,7 +520,8 @@ mecab_model_create(grn_ctx *ctx, grn_mecab_tokenizer_options *options)
 
   argv[argc++] = "Groonga";
   if (!need_default_output) {
-    argv[argc++] = "-Owakati";
+    argv[argc++] = "-F%m\t";
```
kou, line 523: Can we register a word that includes a `\t` to MeCab?

plugins/tokenizers/mecab.c
```diff
@@ -1063,6 +1064,34 @@ mecab_next_default_format(grn_ctx *ctx,
   }
 }
 
+int
```
kou, line 1067: We must use `static` for internal function:

```suggestion
static int
```

plugins/tokenizers/mecab.c
otegami: We could register it. Should we use the other word as a delimiter?
```
$ cat word.csv
search engine,1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
search	engine,1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
$ mecab
search engine
search engine	名詞,一般,*,*,*,*,ペン,ペン,ペン,,
EOS
search	engine
search	engine	名詞,一般,*,*,*,*,ペン,ペン,ペン,,
EOS
```

otegami: I will recheck the word we cannot register in MeCab's dictironary.

abetomo: The same problem should occur with full-width spaces in https://github.com/groonga/groonga/pull/2331#discussion_r2088087587 .

abetomo: How about `\n`?

plugins/tokenizers/mecab.c
kou: It seems that we just need to remove leading/trailing spaces for each tokens.

plugins/tokenizers/mecab.c
otegami: Thanks it sounds nice! We could use `\n` in CSV file but we cannot use it as a registered word as follows.
```console
$ cat word.csv
search engine,1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
search	engine,1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
"search\nengine",1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
$ mecab
search\nengine
search\nengine	名詞,一般,*,*,*,*,ペン,ペン,ペン,,
```
```console
$ cat word.csv
search engine,1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
search	engine,1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
"search
engine",1358,1358,-1,名詞,一般,*,*,*,*,ペン,ペン,ペン,,
$ nkf -e word.csv > word.euc.csv
$ /usr/lib/mecab/mecab-dict-index -d/usr/share/mecab/dic/naist-jdic-eucjp -u user-custom.dic -f euc-jp -t utf-8 word.euc.csv
reading word.euc.csv ... dictionary.cpp(316) [n == 5] format error: "search
```

plugins/tokenizers/mecab.c
```diff
@@ -1091,12 +1099,21 @@ mecab_next_wakati_format(grn_ctx *ctx,
...
         q += space_len;
       }
-      tokenizer->next = q;
-      break;
+      if (is_leading_space) {
+        tokenizer->next = q;
+        break;
+      }
     }
+    is_leading_space = false;
```
abetomo, line 1116: Space should not have to be special handled.
Wouldn't the following code work?

```diff
 
-    if (space_len > 0) {
-      const char *q = r + space_len;
-      if (mecab_wakati_delimiter(r)) {
-        tokenizer->next = q;
-        is_leading_space = true;
-        break;
-      }
-      while (q < e && !mecab_wakati_delimiter(q) &&
-             (space_len = grn_isspace(q, encoding))) {
-        q += space_len;
-      }
-      if (is_leading_space) {
-        tokenizer->next = q;
-        break;
-      }
+    if (mecab_wakati_delimiter(r)) {
+      tokenizer->next = r + 1;
+      break;
     }
-    is_leading_space = false;
   }
```

plugins/tokenizers/mecab.c
```diff
@@ -1063,13 +1064,20 @@ mecab_next_default_format(grn_ctx *ctx,
   }
 }
 
+static int
+mecab_wakati_delimiter(const char *str)
```
kou, line 1068: If we don't use a space for delimiter, it's not called "wakati gaki". So we should not use "wakati" in our code.

plugins/tokenizers/mecab.c
otegami: This 57625b2...4ea14e3 code could remove leading/trailing spaces in user-defined token like the ` search engine` or `search engine `.

otegami:
In macOS and AlmaLinux8/9, the dictionary might not include the `'聞き 糺せん'`. I will check it.

plugins/tokenizers/mecab.c
kou: Absolutely.

kou:
> In macOS and AlmaLinux8/9, the dictionary might not include the `'聞き 糺せん'`. I will check it.

It's included in naist-jdic not ipadic.

lib/str.c
```diff
@@ -1514,6 +1514,12 @@ grn_str_len(grn_ctx *ctx,
   return len;
 }
 
+int
```
komainu8, line 1517:
```suggestion
bool
```

Functions named `isXXX()` should return `true` or `false`.

plugins/tokenizers/mecab.c
otegami: fix: [5c9f82b](https://github.com/groonga/groonga/pull/2331/commits/5c9f82be569bb548b3360f3e4709e8f56e321850) I renamed it from `'wakati'` to `'newline'`.

lib/str.c
otegami: fix: [95ffcdb](https://github.com/groonga/groonga/pull/2331/commits/95ffcdba72041e7e6b3b71c72888f70d9e3d15d2) Sure!

.github/workflows/autotools.yml
```diff
@@ -159,6 +159,7 @@ jobs:
           test/command_line/run-test.rb --groonga-install-prefix=$PWD/install
       - name: "Test: stdio"
         run: |
+          MEACAB_DIC=naist-jdic \
```
kou, line 162: Hmm. If we add a test that needs ipadic, do we need to rewrite this approach?

plugins/tokenizers/mecab.c
```diff
@@ -1089,12 +1096,8 @@ mecab_next_wakati_format(grn_ctx *ctx,
...
     }
 
-    if (space_len > 0) {
-      const char *q = r + space_len;
-      while (q < e && (space_len = grn_isspace(q, encoding))) {
-        q += space_len;
-      }
-      tokenizer->next = q;
+    if ((mecab_next_newline_delimiter(r))) {
+      tokenizer->next = r + space_len;
```
kou, line 1100: Do we need to define `mecab_next_newline_delimiter()`?

```suggestion
    if (cl == 1 && r[0] == '\n') {
      tokenizer->next = r + cl;
```

.github/workflows/autotools.yml
otegami: I should think of another approach about this.
What do you think of passing used dictionaries using grntest's option?
```
grntest --mecab-dictionaries=ipadic,naist-jdic
```

kou: In general, we can't use multiple dictionaries with `TokenMecab` because it always uses the default dictionary.

So `--mecab-dictionaries=ipadic,naist-jdic` will not work.

otegami: Ah, I misunderstood. Our code only checks whether `MECAB_DIC` is set, not what its value actually is. Since we can’t know which dictionary it points to, we have no way to know which dictionary-specific tests to skip. That means if we want to add tests for `ipadic`, we still can’t selectively disable the tests for other dictionaries.

So we should pass one dictionary information. I will add a option to `grntest`.
```
grntest --mecab-dictionaries=ipadic
```

kou: If we want to use the approach, we can use `--mecab-dictionary=ipadic` because we can't use multiple dictionaries at once.
If we want to use the environment variable approach, we need to use `MECAB_DICTIONARY_NAIST_JDIC` or something.

plugins/tokenizers/mecab.c
otegami: fix: [5f3933c](https://github.com/groonga/groonga/pull/2331/commits/5f3933c66598718a8ef52864a8602b005237513f) Thanks! We don't need it because we just use it here in this moment.

.github/workflows/autotools.yml
otegami: fix: [ba714ab](https://github.com/groonga/groonga/pull/2331/commits/ba714ab156c71dcb50b399cd8beb33a41d831aad) I took the environment variable approach because adding MeCab's option to grntest is

I chose the environment‐variable approach because adding a MeCab‐specific flag to grntest felt out of place. I though the grntest options shouldn't be used for handling individual test cases.

otegami:

abetomo:
Could you also add a test that includes `\n`, because it is separated by `\n`?

otegami:
> Could you also add a test that includes \n, because it is separated by \n?

@abetomo 
fix [de039a4](https://github.com/groonga/groonga/pull/2331/commits/de039a4237a917cfbab4dc9f3a47da0e3251ef7a): Thanks, I've just added the test cases about `'\n'`.

komainu8 (approved):

test/command/suite/tokenizers/mecab/one_token_with_middle_space.test
```diff
@@ -0,0 +1,6 @@
+#@require-env MECAB_DICTIONARY_NAIST_JDIC
```
abetomo, line 1: I'm feeling strange that this test is executed whether `MECAB_DICTIONARY_NAIST_JDIC` is `YES` or `NO`.
What do you think?

otegami: I understood the point. If we want to control the behavior using the specified value, I think we couldn't use this option.
I think it's no too strange because  `require-env` option is checking whether the specified environment variable is defined or not. Should we change the value not like `YES`?

abetomo: In CI, where `MECAB_DICTIONARY_NAIST_JDIC=YES` is set, it may be enough to add a comment such as “The value is meaningless and will be executed if this variable is defined”.

Anyway, I'm worried that if I see `MECAB_DICTIONARY_NAIST_JDIC=YES` it looks like I can disable it with `MECAB_DICTIONARY_NAIST_JDIC=NO`.

otegami: @abetomo 
fix [9ef2d49](https://github.com/groonga/groonga/pull/2331/commits/9ef2d494c3ad58dd603b5eff3e5b66357da129e3): Thanks for the suggestion! At now, I’ve updated the CI step to include the comment clarifying that only the presence of `MECAB_DICTIONARY_NAIST_JDIC` is checked and its value is ignored. Next, I will extend grntest with a new option (e.g. --mecab-dictionary) in the following PR.

otegami:

PR #2332, opened 2025-05-15
Fix typos with first letters A, C, D inside doc/ 

qobood:


otegami (approved):
Thank you so much.

abetomo:
Thanks for your contribution!

Could you change the title of the pull request to something that makes sense?
It is not optimal because `part 1` has no meaning.

We believe that committed messages are important.
It is more helpful to be able to guess what the changes will be from the commit message.

The title and description of the pull request will be the commit message.

qobood:
Greetings, @abetomo!
Since "part X" doesn't provide any additional information, I can just get rid of it, would this be better?
I've tried to stick to alphabetical order, so I could include initial letters in PR's title (e.g. "A-C" instead of "part X"), though it will not be fitting all PRs, as latter commits include couple of changes put there to make size of commits more equal.

otegami:
@qobood 

Thanks! Small commit sizes aren’t a problem, so we don’t need to balance them perfectly. How about we stick to strict alphabetical order for the PR titles and include ranges like “A–C”? That way it stays consistent and clear. What do you think?

qobood:
Hi, @abetomo @otegami !
I rearranged some words so commits would stick to alphabetical order, although I didn't touch this PR (aside from renaming it), as it's already approved. So now there is two PRs covering A-D range, is it alright?
Also, does this naming suit better?

otegami (approved):
LGTM 👍🏾

komainu8 (approved):

PR #2333, opened 2025-05-15
Fix typos with first letters A, C, D inside doc/ 

qobood:


komainu8:
It seems that `doc/locale/ja/LC_MESSAGES/news/12.po` is not exist.

otegami (approved):

PR #2338, opened 2025-05-15
Fix typos with first letters S-V inside doc/

qobood:


otegami (commented):
The `doc/locale/ja/LC_MESSAGES/server/http/comparison.po` file is missing in these changes.
Could you check it?

abetomo:
Could you also add  `doc/locale/ja/LC_MESSAGES/reference/executables/groonga-suggest-create-dataset.po` ?

abetomo (approved):

PR #2339, opened 2025-05-19
db: handle empty names and null pointers

otegami:
## Problem

Using UBSAN raised the following error.

```
/home/runner/work/groonga/groonga/lib/db.c:1274:31: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/runner/work/groonga/groonga/lib/db.c:1274:31
```

## Cause

We tried to calculate the offset by adding name_size to null pointer in `name = NULL` case.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -1271,10 +1271,13 @@ grn_rc
 grn_db_check_name(grn_ctx *ctx, const char *name, unsigned int name_size)
 {
   int len;
-  const char *name_end = name + name_size;
-  if (name_size > 0 && *name == GRN_DB_PSEUDO_COLUMN_PREFIX) {
+  if (name_size == 0) {
+    return GRN_SUCCESS;
+  }
+  if (name == NULL || (name_size > 0 && *name == GRN_DB_PSEUDO_COLUMN_PREFIX)) {
     return GRN_INVALID_ARGUMENT;
   }
+  const char *name_end = name + name_size;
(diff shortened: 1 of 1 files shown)


lib/db.c
```diff
@@ -1271,10 +1271,13 @@ grn_rc
 grn_db_check_name(grn_ctx *ctx, const char *name, unsigned int name_size)
 {
   int len;
-  const char *name_end = name + name_size;
-  if (name_size > 0 && *name == GRN_DB_PSEUDO_COLUMN_PREFIX) {
+  if (name_size == 0) {
+    return GRN_SUCCESS;
+  }
+  if (name == NULL || (name_size > 0 && *name == GRN_DB_PSEUDO_COLUMN_PREFIX)) {
```
kou, line 1277: * We don't use `== NULL`/`!= NULL`.
* `name_size > 0` is needless because we have `name_size == 0`.

```suggestion
  if (!name || *name == GRN_DB_PSEUDO_COLUMN_PREFIX) {
```

otegami: fix: [83d4fea](https://github.com/groonga/groonga/pull/2339/commits/83d4feaaf359cd8794d4f038b636e7b70ae62210) Thank you so much.

otegami:

komainu8 (approved):

Issue #1896, opened 2024-08-23
Exact match searches not working with `TABLE_HASH_KEY`


otegami:
### What happened?

When you create an index for a hash table's `_key` using a plain text column, exact‐match queries (`--match_columns "_key" --query fr1`) correctly return only the expected result (`"fr1"` row) as follows. However, if you create an index for it using **reference** column (pointing to source table’s `_key`), the same exact‐match query returns unexpected results (**both** `"fr1"` and `"fr10"`) as follows.

#### Using a plain text column as an index (works as expected)

```
table_create CouponNames TABLE_HASH_KEY ShortText
table_create Coupons TABLE_HASH_KEY ShortText

load --table Coupons
[
  {"_key": "fr1"},
  {"_key": "fr10"}
]

column_create CouponNames coupon_key COLUMN_INDEX Coupons _key

select Coupons --match_columns "_key" --query fr1
```
```
[
  [
    0,
    1748216803.201206,
    0.0004985332489013672
  ],
  [
    [
      [
        1
      ],
      [
        [
          "_id",
          "UInt32"
        ],
        [
          "_key",
          "ShortText"
        ]
      ],
      [
        1,
        "fr1"
      ]
    ]
  ]
]
```

#### Using a reference column as an index (works as unexpected)

```
table_create CouponNames TABLE_HASH_KEY ShortText
table_create Coupons TABLE_HASH_KEY CouponNames

load --table Coupons
[
  {"_key": "fr1"},
  {"_key": "fr10"}
]

column_create CouponNames coupons_key COLUMN_INDEX Coupons _key

select Coupons --match_columns "_key" --query fr1
```
```
[
  [
    0,
    0.0,
    0.0
  ],
  [
    [
      [
        2
      ],
      [
        [
          "_id",
          "UInt32"
        ],
        [
          "_key",
          "ShortText"
        ],
        [
          "coupons_key",
          "Coupons"
        ]
      ],
      [
        1,
        "fr1",
        1
      ],
      [
        2,
        "fr10",
        1
      ]
    ]
  ]
]
```

### Expected behavior

Exact‐match searches on a hash table using a reference column should return **only** rows whose key exactly matches the query string regardless of whether the indexed column is plain text or a reference column.

```
table_create CouponNames TABLE_HASH_KEY ShortText
table_create Coupons TABLE_HASH_KEY CouponNames

load --table Coupons
[
  {"_key": "fr1"},
  {"_key": "fr10"}
]

[...]

PR #2341, opened 2025-05-26
db: enable fulltext search on `_key` when using a reference column's index

otegami:
GitHub: GH-1896

## Problem

Fulltext search on a table's `_key` column was not using the index when the index was defined via a reference column.
It means the following query with reference columns always falls back to a sequential scan with the reference column's index.

As a result, the following case, which was expected to return index‐based behavior, returned unexpected results (e.g. matching both "fr1" and "fr10" instead of only "fr1").

```
table_create CouponNames TABLE_HASH_KEY ShortText
table_create Coupons TABLE_HASH_KEY CouponNames

load --table Coupons
[
  {"_key": "fr1"},
  {"_key": "fr10"}
]

column_create CouponNames coupons_key COLUMN_INDEX Coupons _key

select Coupons --match_columns "_key" --query fr1
```
```
[
  [
    0,
    0.0,
    0.0
  ],
  [
    [
      [
        2
      ],
      [
        [
          "_id",
          "UInt32"
        ],
        [
          "_key",
          "ShortText"
        ],
        [
          "coupons_key",
          "Coupons"
        ]
      ],
      [
        1,
        "fr1",
        1
      ],
      [
        2,
        "fr10",
        1
      ]
    ]
  ]
]
```

## Cause

In the codepath that determines whether a key‐based search could use an index, the `GRN_OP_MATCH` operator
(fulltext search uses this operator) was not being treated as a "key search".

## Solution

We updated the logic so that `GRN_OP_MATCH` is treated as a key search when `_key` has been indexed
through a reference column.

Also, we updated some test cases to reflect index‐scan behavior instead of sequential‐scan behavior because
fulltext search for `_key` now leverages the index.

### DIFF
--- a/test/command/suite/select/filter/no_index/match/key_text.expected
+++ b/test/command/suite/select/filter/no_index/match/key_text.expected
@@ -6,4 +6,4 @@ load --table Memos
 ]
 [[0,0.0,0.0],1]
 select Memos --filter '_key @ "gROONGA"'
-[[0,0.0,0.0],[[[1],[["_id","UInt32"],["_key","ShortText"]],[1,"Groonga"]]]]
+[[0,0.0,0.0],[[[0],[["_id","UInt32"],["_key","ShortText"]]]]]
--- a/lib/db.c
+++ b/lib/db.c
@@ -15518,6 +15518,7 @@ grn_column_find_index_data_accessor_is_key_search(grn_ctx *ctx,
             (accessor->obj->header.flags & GRN_OBJ_KEY_WITH_SIS));
   case GRN_OP_EQUAL:
   case GRN_OP_NOT_EQUAL:
+  case GRN_OP_MATCH:
     switch (accessor->obj->header.type) {
     case GRN_TABLE_HASH_KEY:
     case GRN_TABLE_PAT_KEY:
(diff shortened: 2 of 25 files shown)


test/command/suite/select/filter/no_index/match/key_text.expected
```diff
@@ -6,4 +6,4 @@ load --table Memos
 ]
 [[0,0.0,0.0],1]
 select Memos --filter '_key @ "gROONGA"'
-[[0,0.0,0.0],[[[1],[["_id","UInt32"],["_key","ShortText"]],[1,"Groonga"]]]]
+[[0,0.0,0.0],[[[0],[["_id","UInt32"],["_key","ShortText"]]]]]
```
abetomo, line 9: Is it to be expected that the result of this test changes?
Since there is no index setting for `_key`, I would expect the same result.

otegami: Here, I want to make sure what the expected result is too.
In my current understanding, the `TABLE_PAT_KEY` could support index-search even if we don't specify an index.
ref: https://groonga.org/docs/reference/tables.html

otegami:
I just realized my understanding was incorrect. `TABLE_HASH_KEY` is for exact matches, not full‐text searches. So the change in my PR isn’t needed. I’m closing this PR.

PR #2348, opened 2025-05-30
docs: api/grn_inspect move grn_inspect_limited reference to the header file

otegami:
GitHub: GH-1817

### DIFF
--- a/include/groonga/util.h
+++ b/include/groonga/util.h
@@ -99,6 +99,42 @@ grn_inspect_indented(grn_ctx *ctx,
                      grn_obj *buffer,
                      grn_obj *obj,
                      const char *indent);
+/**
+ * \brief Inspect specified object and produce a length-limited textual
+ *        representation.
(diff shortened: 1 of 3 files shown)


include/groonga/util.h
```diff
@@ -99,6 +99,42 @@ grn_inspect_indented(grn_ctx *ctx,
...
+ * follows.
+ * ```
+ * #<table:pat Users key:ShortText value:(nil) size:7 columns:[] de...(502)
+ * ```
+ *
+ * \param ctx The context object.
+ * \param buffer The buffer where the truncated inspected text will be stored.
+ * \param obj The target object to inspect.
+ *
+ * \return The truncated inspected object in text
```
abetomo, line 136:
```suggestion
 * \return The truncated inspected object in text.
```

abetomo (approved):

komainu8 (approved):

Issue #2349, opened 2025-05-30
Add support for KEY_LARGE in TABLE_PAT_KEY


komainu8:
### What is your problem?

Currently, the total key size of TABLE_PAT_KEY table is limited 4GiB.
Therefore, we don't set index against huge data.

So, we want relieve this limitation.

---

- [x] enable `KEY_LARGE` flag on `TABLE_PAT_KEY` table 

  #2362 

- [x] expand `curr_key`

  #2350 
  #2361 
  #2370

- [x] enable `curr_key_large` on `TABLE_PAT_KEY` table with `KEY_LARGE` flag

  #2368

- [x] expand max of the number of segment for key of `TABLE_PAT_KEY` with  `KEY_LARGE` flag

  #2373 

- [x] add `grn_pat_node_large`

  #2376 

- [x] `put_key()` support `curr_key_large`

   #2368

---

### Utilities

- [x] grn_p_pat_node
- [x] grn_pat_inspect_node

### Related PAT cursor 

- [x] set_cursor_prefix
- [x] set_cursor_near
- [x] set_cursor_common_prefix
- [x] set_cursor_ascend
- [x] set_cursor_descend
- [x] sub_search

### Related PAT add/create

- [x] _grn_pat_create
- [x] grn_pat_add_internal
- [x] grn_pat_add_internal_find
- [x] grn_pat_add_node
- [x] pat_node_set_key
- [x] pat_node_set_shared_key
- [x] _grn_pat_get
- [x] _grn_pat_key
- [x] grn_pat_get_key
- [x] grn_pat_get_key2
- [x] grn_pat_delete_with_sis
- [x] grn_pat_cursor_next

### Fuzzy search

- [x] grn_pat_fuzzy_search_find_prefixed_start_node_id
- [x] grn_pat_fuzzy_search_recursive

### LCP search

- [x] grn_pat_lcp_search

### Related del info

- [x] delinfo_turn_2_internal
- [x] delinfo_turn_2
- [x] delinfo_compute_storage_size
- [x] delinfo_turn_3_internal
- [x] delinfo_turn_3

### Related WAL recover

- [x] grn_pat_wal_recover_add_entry
- [x] grn_pat_wal_recover_add_shared_entry
- [x] grn_pat_wal_recover_reuse_entry
- [x] grn_pat_wal_recover_reuse_shared_entry
- [x] grn_pat_wal_recover_delete_info_phase2
- [x] grn_pat_wal_recover_delete_info_phase3

---

[...]

PR #2350, opened 2025-05-30
pat:  grn_pat_total_key_size() support KEY_LARGE flag

komainu8:

I also modify the return value type of `grn_pat_total_key_size()`.
However, no problem that the return value type of `grn_pat_total_key_size()` is change to `uint64_t` from `uint32_t`.

Because the following two reasons.

1. Currently, `grn_pat_total_key_size()` use only as argument of `grn_ctx_output_uint64()` as below.

```c
grn_ctx_output_uint64(ctx, grn_pat_total_key_size(ctx, pat));
```

2. The second argument type of `grn_ctx_output_uint64()` is uint64_t as below. 

```c
void
grn_ctx_output_uint64(grn_ctx *ctx, uint64_t value)
{
  grn_output_uint64(ctx, ctx->impl->output.buf, ctx->impl->output.type, value);
}
```

### DIFF
--- a/lib/grn_pat.h
+++ b/lib/grn_pat.h
@@ -68,6 +68,7 @@ struct grn_pat_header {
   uint32_t n_entries;
   uint32_t curr_rec;
   uint32_t curr_key;
+  uint64_t curr_key_large;
   uint32_t curr_del;
   uint32_t curr_del2;
   uint32_t curr_del3;
@@ -76,7 +77,7 @@ struct grn_pat_header {
   uint32_t truncated;
   uint32_t n_dirty_opens;
   uint64_t wal_id;
-  uint32_t reserved[1000];
+  uint32_t reserved[998];
   grn_pat_delinfo delinfos[GRN_PAT_NDELINFOS];
   grn_id garbages[GRN_PAT_MAX_KEY_SIZE + 1];
 };
--- a/lib/grn_pat.h
+++ b/lib/grn_pat.h
@@ -136,7 +136,7 @@ grn_pat_fuzzy_search(grn_ctx *ctx,
                      grn_fuzzy_search_optarg *args,
                      grn_hash *h);
 
-uint32_t
+uint64_t
 grn_pat_total_key_size(grn_ctx *ctx, grn_pat *pat);
 
 bool
--- a/lib/grn_pat.h
+++ b/lib/grn_pat.h
@@ -68,7 +68,6 @@ struct grn_pat_header {
   uint32_t n_entries;
   uint32_t curr_rec;
   uint32_t curr_key;
-  uint64_t curr_key_large;
   uint32_t curr_del;
   uint32_t curr_del2;
   uint32_t curr_del3;
@@ -77,6 +76,7 @@ struct grn_pat_header {
   uint32_t truncated;
   uint32_t n_dirty_opens;
   uint64_t wal_id;
+  uint64_t curr_key_large;
   uint32_t reserved[998];
   grn_pat_delinfo delinfos[GRN_PAT_NDELINFOS];
   grn_id garbages[GRN_PAT_MAX_KEY_SIZE + 1];
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -5269,15 +5269,20 @@ grn_pat_inspect_nodes(grn_ctx *ctx, grn_pat *pat, grn_obj *buf)
 {
   pat_node *node;
   grn_obj key_buf;
+  bool is_large_mode = pat_key_is_large_total_key_size(pat);
 
   GRN_TEXT_PUTS(ctx, buf, "{");
   PAT_AT(pat, GRN_ID_NIL, node);
-  if (node->lr[1] != GRN_ID_NIL) {
-    GRN_TEXT_PUTS(ctx, buf, "\n");
-    GRN_OBJ_INIT(&key_buf, GRN_BULK, 0, pat->obj.header.domain);
-    grn_pat_inspect_node(ctx, pat, node->lr[1], -1, &key_buf, 0, "", buf);
-    GRN_OBJ_FIN(ctx, &key_buf);
-    GRN_TEXT_PUTS(ctx, buf, "\n");
+  if (is_large_mode) {
+    ; //TODO: implement if pat node have large key
+  } else {
+    if (node->lr[1] != GRN_ID_NIL) {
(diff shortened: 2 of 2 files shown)


lib/pat.c
```diff
@@ -92,6 +93,12 @@ typedef struct {
 #define PAT_DELETING  (1 << 1)
 #define PAT_IMMEDIATE (1 << 2)
 
+static inline bool
+pat_key_is_large_total_key_size(grn_pat *pat)
```
kou: Can we simplify this?

```suggestion
pat_is_key_large(grn_pat *pat)
```

lib/pat.c
```diff
@@ -5262,15 +5269,20 @@ grn_pat_inspect_nodes(grn_ctx *ctx, grn_pat *pat, grn_obj *buf)
...
+  if (is_large_mode) {
+    ; // TODO: implement if pat node have large key
+  } else {
+    if (node->lr[1] != GRN_ID_NIL) {
+      GRN_TEXT_PUTS(ctx, buf, "\n");
+      GRN_OBJ_INIT(&key_buf, GRN_BULK, 0, pat->obj.header.domain);
+      grn_pat_inspect_node(ctx, pat, node->lr[1], -1, &key_buf, 0, "", buf);
+      GRN_OBJ_FIN(ctx, &key_buf);
+      GRN_TEXT_PUTS(ctx, buf, "\n");
+    }
```
kou: Do we need this change in this PR...?

komainu8: I think also that this change is needless in this PR. However, when unused function exist, clang raise the following error. 

```
/home/runner/work/groonga/groonga/lib/pat.c:97:1: error: unused function 'pat_key_is_large_total_key_size' [-Werror,-Wunused-function]
   97 | pat_key_is_large_total_key_size(grn_pat *pat)
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
https://github.com/groonga/groonga/actions/runs/15344296603/job/43177061984#step:8:397 

So, I added this change in this PR.

komainu8: Should we add `-Wno-unused-function` as below?

```diff
modified   CMakeLists.txt
@@ -357,6 +357,7 @@ endif()
 if(GRN_C_COMPILER_GNU_LIKE)
   check_build_flag("-Wall")
   check_build_flag("-Wno-unused-but-set-variable")
+  check_build_flag("-Wno-unused-function")
   check_build_flag("-Wno-unused-parameter")
   check_cflag("-Wno-pointer-sign")
   check_build_flag("-Wfloat-equal")
```

kou: We should not disable it to detect unused functions.

kou: Can we implement a small feature that use `pat_is_key_large()`?

komainu8: I see. I'll try it.

komainu8: I implemented a small feature that use `pat_is_key_large()`.

lib/grn_pat.h
```diff
@@ -68,6 +68,7 @@ struct grn_pat_header {
   uint32_t n_entries;
   uint32_t curr_rec;
   uint32_t curr_key;
+  uint64_t curr_key_large;
```
abetomo: It should be added at the end because it is bad to change the order.

komainu8: Oh...Thank you for your comment.
I'll fix it.

komainu8: I've fixed it.

lib/grn_pat.h
```diff
@@ -76,7 +76,8 @@ struct grn_pat_header {
   uint32_t truncated;
   uint32_t n_dirty_opens;
   uint64_t wal_id;
-  uint32_t reserved[1000];
+  uint64_t curr_key_large;
+  uint32_t reserved[998];
```
otegami, line 80: 📝 `uint32_t` is guaranteed to be exactly 32 bits (4 bytes). Since we added a `uint64_t` field (8 bytes), we reduced the reserved array by two elements (2 × 4 bytes) to keep the struct’s total size unchanged.

lib/grn_pat.h
otegami, line 1: Will you add your copyright here too?

lib/pat.c
```diff
@@ -92,6 +93,12 @@ typedef struct {
 #define PAT_DELETING  (1 << 1)
 #define PAT_IMMEDIATE (1 << 2)
 
+static inline bool
+pat_is_key_large(grn_pat *pat)
+{
+  return (pat->header->flags & GRN_OBJ_KEY_LARGE) == GRN_OBJ_KEY_LARGE;
```
otegami, line 99: Could we write the following? 
```suggestion
  return pat->header->flags & GRN_OBJ_KEY_LARGE;
```

otegami: 📝 The result could be converted to 0 or 1.
```c
$ cat pat_is_key_large.c
#include <stdio.h>
#include <stdbool.h>

static inline bool is_nonzero(int x) {
    return x;
}

int main(void) {
    int a = 0;
    int b = 42;

    printf("is_nonzero(a＝0): %d\n", is_nonzero(a));
    printf("is_nonzero(b＝42): %d\n", is_nonzero(b));

    return 0;
}
```
```console
$ gcc pat_is_key_large.c
$ ./a.out
is_nonzero(a＝0): 0
is_nonzero(b＝42): 1
```

otegami: 📝 The existed code in `hash.c` uses the same approach now here.
https://github.com/groonga/groonga/blob/58a20e712ac4178bcb13d1792c24701cd29d97c0/lib/hash.c#L5160-L5164

otegami: My comment isn't strong opinion because both approaches are good to me.

komainu8: The implement of this function like `grn_hash_is_large_total_key_size()`(https://github.com/groonga/groonga/blob/v15.1.1/lib/hash.c#L5161).

Also, the return value type of this function is bool.
However, the result of `pat->header->flags & GRN_OBJ_KEY_LARGE` is `0` or `4096`.
Probably, the above value is typecasted implicit, but I think that we should not except behavior implicit.

lib/grn_pat.h
komainu8: Changes of this file are not originality. So, I think that I have not copyright.

otegami (approved):
LGTM 👍🏾

PR #2354, opened 2025-06-03
docs: api/grn_inspect move grn_inspect_name reference to the header file

otegami:
GitHub: GH-1817

### DIFF
--- a/include/groonga/util.h
+++ b/include/groonga/util.h
@@ -137,6 +137,38 @@ grn_inspect_indented(grn_ctx *ctx,
  */
 GRN_API grn_obj *
 grn_inspect_limited(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
+/**
+ * \brief Inspect specified object and return the name of the specified object.
+ *
+ * This function inspects the given \p obj and writes its name into \p buffer.
+ * If \p obj has a name, that name is copied into \p buffer.
+ * Otherwise:
+ * - If \p obj's id is a \ref GRN_ID_NIL, the `(nil)` is stored in \p buffer.
+ * - If \p obj exists but has no name, the `(anonymous:xx)` is stored.
+ *
+ * For example usage:
+ * ```c
+ * grn_obj name;
+ * GRN_TEXT_INIT(&name, 0);
+ * grn_inspect_name(ctx, &name, obj);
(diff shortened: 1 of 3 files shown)


include/groonga/util.h
```diff
@@ -137,6 +137,38 @@ grn_inspect_indented(grn_ctx *ctx,
...
+ * printf("%.*s\n", (int)GRN_TEXT_LEN(&name),
+ *        GRN_TEXT_VALUE(&name));
+ * GRN_OBJ_FIN(ctx, &name);
+ * ```
+ *
+ * For example output:
+ * ```
+ * Users
+ * (nil)
+ * (anonymous:42)
```
abetomo, line 164: How about adding that one of these can be got?
It looks like we get 3 rows in one run.

otegami: fix: [9adc4ae](https://github.com/groonga/groonga/pull/2354/commits/9adc4ae24baccb133893de5357af651c8a4aba60) Thanks.

include/groonga/util.h
```diff
@@ -137,6 +137,39 @@ grn_inspect_indented(grn_ctx *ctx,
...
 GRN_API grn_obj *
 grn_inspect_limited(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
+/**
+ * \brief Inspect specified object and return the name of the specified object.
+ *
+ * This function inspects the given \p obj and writes its name into \p buffer.
+ * If \p obj has a name, that name is copied into \p buffer.
+ * Otherwise:
+ * - If \p obj's id is a \ref GRN_ID_NIL, the `(nil)` is stored in \p buffer.
+ * - If \p obj exists but has no name, the `(anonymous:xx)` is stored.
```
abetomo:
```suggestion
 * - If \p obj exists but has no name, the `(anonymous:ID)` is stored.
```

I think `ID` is better than `xx`.
It is easier to understand that an ID is set.

otegami: [2e091f7](https://github.com/groonga/groonga/pull/2354/commits/2e091f7bc8ba4d538c784d1fcde8c54a5009d7d2) Sure!

otegami:
Thank you for reviewing. I've just fixed it.

komainu8 (approved):

PR #2355, opened 2025-06-04
ci: cmake: Utilize `rubygems-requirements-system`

abetomo:
Related: GitHub GH-2353

It automatically installs dependent libraries.

PR #2353, opened 2025-06-03
ci: benchmark: Utilize `rubygems-requirements-system`

abetomo:
It automatically installs dependent libraries.

PR #2356, opened 2025-06-05
packages: test: Utilize `rubygems-requirements-system`

abetomo:
Related: GitHub GH-2353, GH-2355
It automatically installs dependent libraries.

### DIFF
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -63,6 +63,7 @@ apt install -V -y \
   make \
   ruby-dev \
   tzdata
+gem install rubygems-requirements-system
 MAKEFLAGS=-j$(nproc) gem install grntest
 
 if groonga --version | grep -q apache-arrow; then
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -72,6 +72,7 @@ if [ "${run_test}" = "yes" ]; then
   if [ ${os} != "amazon-linux" ]; then
     ${DNF} install -y redhat-rpm-config
   fi
+  gem install rubygems-requirements-system
   MAKEFLAGS=-j$(nproc) gem install grntest
 
   export TZ=Asia/Tokyo


packages/apt/test.sh
```diff
@@ -63,6 +63,7 @@ apt install -V -y \
   make \
   ruby-dev \
   tzdata
+gem install rubygems-requirements-system
```
otegami, line 66: We need `gem install rubygems-requirements-system` when we want to install red-arrow.
How about moving the following place?
```diff
  if groonga --version | grep -q apache-arrow; then
    apt install -V -y \
      g++ \
      libre2-dev
+   gem install rubygems-requirements-system
    MAKEFLAGS=-j$(nproc) gem install red-arrow
  fi
```
https://github.com/groonga/groonga/actions/runs/15455913902/job/43508273535#step:14:2915

packages/yum/test.sh
```diff
@@ -72,6 +72,7 @@ if [ "${run_test}" = "yes" ]; then
   if [ ${os} != "amazon-linux" ]; then
     ${DNF} install -y redhat-rpm-config
   fi
+  gem install rubygems-requirements-system
```
otegami, line 75: We don't use gem red-arrow here, so I think we don't need here.

packages/apt/test.sh
abetomo: You are right in the sense that it avoids errors when installing `red-arrow`.
But there should be cases where `rubygems-requirements-system` also works effectively when installing other gems.
That is why I installed it first.

packages/yum/test.sh
abetomo: I added `rubygems-requirements-system` because there should be some gems that work effectively.
I also thought it would be easier to understand since the process would be similar to `apt/test.sh`.

packages/apt/test.sh
otegami: I was worried that we might assume other gems also require dependencies, but since installing it doesn’t introduce any overhead when it’s not needed, it seems fine to include it here.

PR #2361, opened 2025-06-09
check: add a new item for outputting curr_key_large

komainu8:
Currently, tests of `check` command are not exist.
We will implements these tests after we supported `KEY_LARGE`.


abetomo (approved):

lib/pat.c
```diff
@@ -5118,6 +5118,8 @@ grn_pat_check(grn_ctx *ctx, grn_pat *pat)
   GRN_OUTPUT_INT64(h->curr_rec);
   GRN_OUTPUT_CSTR("curr_key");
   GRN_OUTPUT_INT64(h->curr_key);
+  GRN_OUTPUT_CSTR("curr_key_large");
+  GRN_OUTPUT_UINT64(h->curr_key_large);
```
kou: Should we output both of `curr_key` and `curr_key_large`?

komainu8: I agree with you.
I've fixed it and implement test for this change, but this test is failure. Because we don't specify `KEY_LARGE` flag on `TABLE_PAT_KEY` table yet.
So, I'll implement that specify `KEY_LARGE` flag on `TABLE_PAT_KEY` table in GH-2362 before this modification.

lib/pat.c
```diff
@@ -5124,8 +5124,13 @@ grn_pat_check(grn_ctx *ctx, grn_pat *pat)
...
   GRN_OUTPUT_INT64(h->curr_rec);
-  GRN_OUTPUT_CSTR("curr_key");
-  GRN_OUTPUT_INT64(h->curr_key);
+  if (!pat_is_key_large(pat)) {
+    GRN_OUTPUT_CSTR("curr_key");
+    GRN_OUTPUT_INT64(h->curr_key);
+  } else {
+    GRN_OUTPUT_CSTR("curr_key_large");
+    GRN_OUTPUT_UINT64(h->curr_key_large);
+  }
```
otegami: How about reversing the condition as follows? It will be easy to understand it.
```suggestion
  if (pat_is_key_large(pat)) {
    GRN_OUTPUT_CSTR("curr_key_large");
    GRN_OUTPUT_UINT64(h->curr_key_large)
  } else {;
    GRN_OUTPUT_CSTR("curr_key");
    GRN_OUTPUT_INT64(h->curr_key);
  }
```

abetomo:
```suggestion
  GRN_OUTPUT_CSTR("curr_key");
  if (!pat_is_key_large(pat)) {
    GRN_OUTPUT_INT64(h->curr_key);
  } else {
    GRN_OUTPUT_UINT64(h->curr_key_large);
  }
```

Is it not better to keep the key as `curr_key` and change only the value?

komainu8: For sure. I'll change only the value.

test/command/suite/check/pat/default.test
```diff
@@ -0,0 +1,6 @@
+table_create --name Users --flags TABLE_PAT_KEY --key_type ShortText
+load --table Users
+[
+{"_key":"Alice"}
+]
+check Users
```
abetomo:
```suggestion
check --obj Users
```

I think the http test will be successful with this change.
(The `.expected` needs to be updated as well.)

komainu8: Oh... Thank you very match!
I'll fixed it.

abetomo (approved):

PR #2362, opened 2025-06-09
pat: allow use KEY_LARGE flag on TABLE_PAT_KEY

komainu8:

Currently, we can only specify KEY_LARGE flag on TABLE_PAT_KEY table. Even if we specify `KEY_LARGE`, behavior is not almost change currently.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -1598,10 +1598,11 @@ grn_table_create_validate(grn_ctx *ctx,
     return ctx->rc;
   }
 
-  if ((flags & GRN_OBJ_KEY_LARGE) && table_type != GRN_OBJ_TABLE_HASH_KEY) {
+  if ((flags & GRN_OBJ_KEY_LARGE) && (table_type != GRN_OBJ_TABLE_HASH_KEY
+                                      && table_type != GRN_OBJ_TABLE_PAT_KEY) ) {
     ERR(GRN_INVALID_ARGUMENT,
         "[table][create] "
-        "large key support is available only for TABLE_HASH_KEY key table: "
+        "large key support is available for TABLE_HASH_KEY and TABLE_PAT_KEY key table: "
         "<%.*s>(%s)",
         name_size,
         name,
--- a/test/command/suite/table_create/key_large/double_array_trie.test
+++ b/test/command/suite/table_create/key_large/double_array_trie.test
@@ -1,0 +1,2 @@
+table_create Users TABLE_DAT_KEY|KEY_LARGE ShortText
+
--- a/test/command/suite/dump/schema/table/patricia_trie/key_large_flag.expected
+++ b/test/command/suite/dump/schema/table/patricia_trie/key_large_flag.expected
@@ -1,0 +1,4 @@
+table_create Users TABLE_PAT_KEY|KEY_LARGE ShortText
+[[0,0.0,0.0],true]
+dump
+table_create Users TABLE_PAT_KEY|KEY_LARGE ShortText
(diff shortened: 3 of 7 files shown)


otegami (approved):

test/command/suite/table_create/key_large/double_array_trie.test
```diff
@@ -0,0 +1,2 @@
+table_create Users TABLE_DAT_KEY|KEY_LARGE ShortText
+
```
abetomo:
```suggestion
```

lib/db.c
```diff
@@ -1598,10 +1598,12 @@ grn_table_create_validate(grn_ctx *ctx,
...
 
-  if ((flags & GRN_OBJ_KEY_LARGE) && table_type != GRN_OBJ_TABLE_HASH_KEY) {
+  if ((flags & GRN_OBJ_KEY_LARGE) && (table_type != GRN_OBJ_TABLE_HASH_KEY &&
+                                      table_type != GRN_OBJ_TABLE_PAT_KEY)) {
     ERR(GRN_INVALID_ARGUMENT,
         "[table][create] "
-        "large key support is available only for TABLE_HASH_KEY key table: "
+        "large key support is available for TABLE_HASH_KEY and TABLE_PAT_KEY "
+        "key table: "
         "<%.*s>(%s)",
```
abetomo:
```suggestion
        "large key support is available for TABLE_HASH_KEY and TABLE_PAT_KEY "
        "key table: <%.*s>(%s)",
```

No need for a line break?

lib/db.c
```diff
@@ -1598,10 +1598,12 @@ grn_table_create_validate(grn_ctx *ctx,
     return ctx->rc;
   }
 
-  if ((flags & GRN_OBJ_KEY_LARGE) && table_type != GRN_OBJ_TABLE_HASH_KEY) {
+  if ((flags & GRN_OBJ_KEY_LARGE) && (table_type != GRN_OBJ_TABLE_HASH_KEY &&
+                                      table_type != GRN_OBJ_TABLE_PAT_KEY)) {
```
abetomo, line 1602:
```suggestion
  if ((flags & GRN_OBJ_KEY_LARGE) && table_type != GRN_OBJ_TABLE_HASH_KEY &&
                                      table_type != GRN_OBJ_TABLE_PAT_KEY) {
```

No need for brackets `()`?

lib/db.c
komainu8: Ah... This is correct.
Thank you for your comment.

lib/db.c
komainu8: I think that `table_type != GRN_OBJ_TABLE_HASH_KEY && table_type != GRN_OBJ_TABLE_PAT_KEY` is one group.
So, I use `()`. Because I think I should use `()` to easy to read.

abetomo:
```suggestion
  if ((flags & GRN_OBJ_KEY_LARGE) &&
    (table_type & (GRN_OBJ_TABLE_HASH_KEY | GRN_OBJ_TABLE_PAT_KEY))) {
```

Can we write it like this?

komainu8: No. we don't write it like that.

For example, if table is `TABLE_PAT_KEY`, `table_type & (GRN_OBJ_TABLE_HASH_KEY | GRN_OBJ_TABLE_PAT_KEY)` is `true`. However, I don't want to execute output error log, if  table is `TABLE_PAT_KEY`.

kou: `table_type` isn't a flag: https://github.com/groonga/groonga/blob/ca2683888109333ceb2ec75449d8186afb622ec8/include/groonga/groonga.h#L685-L689

So we can't use bit operations.

otegami (approved):

PR #2363, opened 2025-06-09
ci cmake: run daily

otegami:
During the last release flow, our CMake build job failed and blocked the Groonga release.
This change makes the CMake workflow run daily, preventing it from silently becoming a blocker.

We’ve scheduled the CMake build to run daily at 23:00 UTC (which finishes around 08:00 JST).
This timing also helps us catch any breakages early on release day.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -60,6 +60,9 @@ on:
       - 'test/command_line/**'
       - 'test/mruby/**'
       - 'vendor/mruby/**'
+  schedule:
+    - cron: |
+        0 0 * * *
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -60,6 +60,9 @@ on:
       - 'test/command_line/**'
       - 'test/mruby/**'
       - 'vendor/mruby/**'
+  schedule:
+    - cron: |
+        0 0 * * *
```
abetomo:
```suggestion
        0 21 * * *
```

0:00 would end around 11:00 JST, so how about setting it to end around 8:00 JST?
If something problems are found on the release date, we will have time to fix them.

PR #2366, opened 2025-06-10
docs: api/grn_inspect move grn_inspect_type reference to the header file

otegami:
GitHub: GH-1817

### DIFF
--- a/include/groonga/util.h
+++ b/include/groonga/util.h
@@ -214,6 +214,80 @@ grn_inspect_name(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
  */
 GRN_API grn_obj *
 grn_inspect_encoding(grn_ctx *ctx, grn_obj *buffer, grn_encoding encoding);
+/**
+ * \brief Inspect the given type and produce its name.
+ *
+ * \since 4.0.8
+ *
+ * For example usage:
+ * ```c
+ * grn_obj type_name;
+ * GRN_TEXT_INIT(&type_name, 0);
+ * grn_inspect_type(ctx, &type_name, obj->header.type);
+ * printf("#=> %.*s\n", (int)GRN_TEXT_LEN(&type_name),
+ * GRN_TEXT_VALUE(&type_name)); GRN_OBJ_FIN(ctx, &type_name);
+ * ```
+ *
+ * For example output:
+ * Depending on \p type, it will output one of the following lines:
+ *
+ * Table types:
+ * ```
+ * GRN_DB
+ * GRN_TABLE_DAT_KEY
+ * GRN_TABLE_HASH_KEY
+ * GRN_TABLE_NO_KEY
+ * GRN_TABLE_PAT_KEY
+ * ```
+ *
+ * Column types:
+ * ```
+ * GRN_COLUMN_FIX_SIZE
+ * GRN_COLUMN_INDEX
+ * GRN_COLUMN_VAR_SIZE
+ * ```
+ *
+ * Data container types:
+ * ```
+ * GRN_BULK
+ * GRN_PTR
+ * GRN_PVECTOR
+ * GRN_UVECTOR
+ * GRN_VECTOR
+ * ```
+ *
+ * Cursor types:
+ * ```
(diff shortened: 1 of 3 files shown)


include/groonga/util.h
```diff
@@ -214,6 +214,80 @@ grn_inspect_name(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
...
+ *
+ * \since 4.0.8
+ *
+ * For example usage:
+ * ```c
+ * grn_obj type_name;
+ * GRN_TEXT_INIT(&type_name, 0);
+ * grn_inspect_type(ctx, &type_name, obj->header.type);
+ * printf("#=> %.*s\n", (int)GRN_TEXT_LEN(&type_name),
+ * GRN_TEXT_VALUE(&type_name)); GRN_OBJ_FIN(ctx, &type_name);
```
abetomo:
```suggestion
 * printf("%.*s\n", (int)GRN_TEXT_LEN(&type_name), GRN_TEXT_VALUE(&type_name));
 * GRN_OBJ_FIN(ctx, &type_name);
```

I think it would be better without `#=>` to match the output example.

otegami: fix: [ea80160](https://github.com/groonga/groonga/pull/2366/commits/ea80160252f13a72a63b10d7156fc06ca1483808) Thanks!

include/groonga/util.h
```diff
@@ -214,6 +214,82 @@ grn_inspect_name(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
...
+ * GRN_STRING
+ * GRN_TYPE
+ * GRN_VOID
+ * (unknown: 0x42)
+ * ```
+ *
+ * \param ctx The context object.
+ * \param buffer The buffer where the type name will be stored.
+ * \param type The type to inspect. This is typically obtained from an object's
+ *             header.type field.
```
abetomo: `\ref grn_obj_header::type` might be better.

abetomo (approved):

include/groonga/util.h
otegami: fix b7c019e: Thanks.

include/groonga/util.h
```diff
@@ -214,6 +214,82 @@ grn_inspect_name(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
...
+ *        GRN_TEXT_VALUE(&type_name));
+ * GRN_OBJ_FIN(ctx, &type_name);
+ * ```
+ *
+ * For example output:
+ * Depending on \p type, it will output one of the following lines:
+ *
+ * Table types:
+ * ```
+ * GRN_DB
```
komainu8: `GRN_DB` is not table.
So, I believe that `GRN_DB` is not table types.

otegami: Should we move to `Other object types` here?

include/groonga/util.h
```diff
@@ -214,6 +214,82 @@ grn_inspect_name(grn_ctx *ctx, grn_obj *buffer, grn_obj *obj);
...
+ * GRN_EXPR
+ * GRN_MSG
+ * GRN_PATSNIP
+ * GRN_PROC
+ * GRN_QUERY
+ * GRN_SNIP
+ * GRN_STRING
+ * GRN_TYPE
+ * GRN_VOID
+ * (unknown: 0x42)
```
komainu8, line 287: If type is `unknown`, `grn_inspect_type()` is not always return `unknown: 0x42` by implement of this function. ( https://github.com/groonga/groonga/blob/main/lib/util.c#L195-L206 )

include/groonga/util.h
komainu8: I think it's good.

include/groonga/util.h
otegami: I added the example outputs here. So is it better for us to show a example like `0xXX`?

otegami: fix: [36c912f](https://github.com/groonga/groonga/pull/2366/commits/36c912f2c2945fd08c16190ffd2d0b836d67a01c) I added the above explanation.

komainu8 (approved):

PR #2367, opened 2025-06-10
ci windows: add a built package with memory debug

otegami:
This change adds a new build matrix entry for Windows MSVC builds that enables `--preset=memory-debug`.
This package is useful for debugging memory-related issues on Windows.


komainu8 (approved):

abetomo:
I think that debug build will be built in each user's local environment and used for testing.
Do we need to build with groonga/groonga's CI?

otegami:
This change is only required for a specific case and isn’t applicable upstream, so I’m closing this.

PR #2376, opened 2025-06-13
pat: add support for pat node has large key offset

komainu8:
This commit is part of the work to implement #2349.

We add a new struct for pat node has large key.
Because we don't allocate a needless memory space when we don't use `KEY_LARGE` flag.

We implement only definition of a new struct and a new union in this commit.
We will implement codes use this struct and union by the other commit.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -96,6 +96,24 @@ typedef struct {
   uint16_t bits;
 } pat_node;
 
+/* This struct almost the same pat_node.
+   We only use this struct when we specify `KEY_LARGE`
+   This struct is different from pat_node only in that the size of key.
+   The size of pat_node.key is uint32_t.
+   The size of pat_node_large.key is uint64_t.
+ */
+typedef struct {
+  grn_id lr[2];
+  uint64_t key;
+  uint16_t check;
+  uint16_t bits;
+} pat_node_large;
+
+typedef union {
+  pat_node default_node;
+  pat_node_large large_node;
+} pat_node_common;
+
 #define PAT_DELETING  (1 << 1)
 #define PAT_IMMEDIATE (1 << 2)
 
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -96,7 +96,8 @@ typedef struct {
   uint16_t bits;
 } pat_node;
 
-/* This struct almost the same pat_node.
+/*
+   This struct is almost the same as `pat_node`.
    We only use this struct when we specify `KEY_LARGE`
    This struct is different from pat_node only in that the size of key.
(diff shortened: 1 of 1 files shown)


lib/pat.c
```diff
@@ -96,6 +96,24 @@ typedef struct {
   uint16_t bits;
 } pat_node;
 
+/* This struct almost the same pat_node.
+   We only use this struct when we specify `KEY_LARGE`
+   This struct is different from pat_node only in that the size of key.
+   The size of pat_node.key is uint32_t.
+   The size of pat_node_large.key is uint64_t.
```
otegami:
```suggestion
/*
   This struct is almost the same as `pat_node`.
   We only use this struct when we specify `KEY_LARGE`
   This struct is different from pat_node only in that the size of key.
   The size of pat_node.key is uint32_t.
   The size of pat_node_large.key is uint64_t.
```

lib/pat.c
```diff
@@ -96,6 +96,24 @@ typedef struct {
...
+ */
+typedef struct {
+  grn_id lr[2];
+  uint64_t key;
+  uint16_t check;
+  uint16_t bits;
+} pat_node_large;
```
otegami, line 111: Should we define the `_pat_node_large` ?
```
typedef struct _pat_node_large pat_node_large;
```

komainu8: Sorry, I don't know why should we define the `_pat_node_large`.

otegami: I think it will resolve circular dependencies using forward-declarations.
And also, internal structs uses  `_grn_` prefix in Groonga. 
- https://stackoverflow.com/questions/55978064/resolving-circular-dependencies-using-forward-declarations

abetomo (approved):

lib/pat.c
otegami: We don't use the type which is defined in other files here. So we don't need it here.

otegami: Forward‐declaring a struct is only necessary in the following cases.
So, in this case we don't need it here.
- A type needs to refer to itself as a member (e.g. a node in a linked list)
- Two types mutually refer to each other (A uses B and B uses A).

komainu8: I see.

lib/pat.c
```diff
@@ -96,6 +96,25 @@ typedef struct {
...
+
+typedef union {
+  pat_node default_node;
```
kou, line 114: Why do we need `default_` prefix here?

komainu8: Because I want to emphasis that `pat_node` is default.
But this prefix is redundant. I'll remove this.

PR #2386, opened 2025-06-18
pat: fix displaying wrong max total key size when we specify KEY_LARGE flag

komainu8:
This commit is part of the work to implement #2349.


lib/pat.c
```diff
@@ -1150,7 +1150,7 @@ key_put(grn_ctx *ctx, grn_pat *pat, const uint8_t *key, uint32_t len)
         "max=%u: current=%" GRN_FMT_INT64U ": new key size=%u",
```
kou: Should we update `max=%u` too?

komainu8: Oh No...
Sorry, I will fix it...

otegami (approved):

PR #2393, opened 2025-06-23
pat: add pat_at() instead of PAT_AT macro

komainu8:
Because macro is difficult debug.

This commit is part of the work to implement https://github.com/groonga/groonga/issues/2349.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -1065,24 +1065,7 @@ static inline pat_node_common *
 pat_at(grn_ctx *ctx, grn_pat *pat, grn_id id)
 {
   int flags = 0;
-  pat_node_common *n;
-
-  if (pat_is_key_large(pat)) {
-    pat_node_large *node =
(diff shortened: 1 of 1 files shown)


kou:
Should we remove `PAT_AT()`?

lib/pat.c
```diff
@@ -1061,6 +1061,13 @@ grn_pat_wal_add_entry_data_set_record_direction(
     n = grn_io_array_at(ctx, pat->io, SEGMENT_PAT, id, &flags);                \
   } while (0)
 
+static inline pat_node_common *
+pat_at(grn_ctx *ctx, grn_pat *pat, grn_id id)
+{
+  int flags = 0;
```
otegami, line 1067: [nits] Do we have a constant here instead of hard-coding 0?

komainu8:
> Should we remove `PAT_AT()`?

Ah...
We don't need to remove `PAT_AT()`.
So, I close this PR.

PR #2399, opened 2025-06-26
Fix Apache Arrow output for slice and drilldown

otegami:
When running slice or drilldown with `output_type=apache-arrow`, Groonga would crash because it attempted to set a label in the Arrow output, which isn’t supported in that context.

To prevent this, we added `GROONGA:data_type` and `GROONGA:label` metadata to the Apache Arrow output for slice and drilldown results.

### DIFF
--- a/lib/ctx.cpp
+++ b/lib/ctx.cpp
@@ -452,12 +452,10 @@ grn_ctx_impl_init(grn_ctx *ctx)
                       grn_msgpack_buffer_write);
 #endif
   grn_timeval_now(ctx, &ctx->impl->tv);
-#ifdef GRN_WITH_APACHE_ARROW
   ctx->impl->output.arrow_stream_writer = NULL;
   ctx->impl->output.arrow_metadata_data_type = NULL;
   GRN_RAW_STRING_INIT(ctx->impl->output.arrow_metadata_label);
   ctx->impl->arrow_stream_loader = NULL;
-#endif
   grn_loader_init(&ctx->impl->loader);
   ctx->impl->plugin_path = NULL;
 
@@ -854,13 +852,11 @@ grn_ctx_impl_fin(grn_ctx *ctx)
   if (ctx->impl->parallel.task_executor) {
     delete static_cast<grn::TaskExecutor *>(ctx->impl->parallel.task_executor);
   }
-#ifdef GRN_WITH_APACHE_ARROW
   if (ctx->impl->output.arrow_stream_writer) {
     grn_arrow_stream_writer_close(ctx, ctx->impl->output.arrow_stream_writer);
   }
   ctx->impl->output.arrow_metadata_data_type = NULL;
   GRN_RAW_STRING_INIT(ctx->impl->output.arrow_metadata_label);
-#endif
   GRN_OBJ_FIN(ctx, &ctx->impl->output.names);
   GRN_OBJ_FIN(ctx, &ctx->impl->output.levels);
   rc = grn_obj_close(ctx, ctx->impl->output.buf);
--- a/lib/ctx.cpp
+++ b/lib/ctx.cpp
@@ -451,9 +451,11 @@ grn_ctx_impl_init(grn_ctx *ctx)
                       ctx,
                       grn_msgpack_buffer_write);
 #endif
-  ctx->impl->output.arrow_stream_writer = NULL;
   grn_timeval_now(ctx, &ctx->impl->tv);
 #ifdef GRN_WITH_APACHE_ARROW
+  ctx->impl->output.arrow_stream_writer = NULL;
+  ctx->impl->output.arrow_metadata_data_type = NULL;
+  GRN_RAW_STRING_INIT(ctx->impl->output.arrow_metadata_label);
   ctx->impl->arrow_stream_loader = NULL;
 #endif
   grn_loader_init(&ctx->impl->loader);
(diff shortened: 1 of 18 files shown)


komainu8:
Could you update "when using command_version=3." in this PR's description?

otegami:
> Could you update "when using command_version=3." in this PR's description?

Sure. I've just updated it.

lib/output.c
```diff
@@ -2710,6 +2710,27 @@ grn_output_result_set_open_metadata(grn_ctx *ctx,
...
+        "GROONGA:data_type",
+        ctx->impl->output.arrow_metadata_data_type);
+      grn_arrow_stream_writer_add_metadata(
+        ctx,
+        ctx->impl->output.arrow_stream_writer,
+        "GROONGA:label",
+        GRN_TEXT_VALUE(&label_buffer));
+      GRN_OBJ_FIN(ctx, &label_buffer);
+    }
   }
```
komainu8: Is `#ifdef GRN_WITH_APACHE_ARROW ... #endif` unnecessary in this block?

otegami: I think we don't need it because we have already checked whether the output format is Apache Arrow format or not.

komainu8: Does "checked whether the output format is Apache Arrow format or not" mean `if (output_type == GRN_CONTENT_APACHE_ARROW)` condition?

I think if we use `GRN_WITH_APACHE_ARROW`, this process is efficiency.
If `#ifdef GRN_WITH_APACHE_ARROW ... #endif` is exist, Groonga is not process `if (output_type == GRN_CONTENT_APACHE_ARROW)`condition.
Because compiler exclude this codes from compile target

otegami: The existed code doesn't handle it at all when it checks the e.x. `GRN_CONTENT_APACHE_ARROW`. Do we really need it here? If so, do we also add it in the other places too?

kou: We don't need `#ifdef GRN_WITH_APACHE_ARROW` here.

`grn_arrow_stream_writer_*()` are always defined to avoid many `#ifdef GRN_WITH_APACHE_ARROW` in many places.

komainu8: I see.

komainu8 (approved):

lib/ctx.cpp
```diff
@@ -451,9 +451,11 @@ grn_ctx_impl_init(grn_ctx *ctx)
                       ctx,
                       grn_msgpack_buffer_write);
 #endif
-  ctx->impl->output.arrow_stream_writer = NULL;
   grn_timeval_now(ctx, &ctx->impl->tv);
 #ifdef GRN_WITH_APACHE_ARROW
```
kou: We can remove this `#ifdef`.

lib/ctx.cpp
```diff
@@ -451,9 +451,11 @@ grn_ctx_impl_init(grn_ctx *ctx)
...
+  ctx->impl->output.arrow_stream_writer = NULL;
+  ctx->impl->output.arrow_metadata_data_type = NULL;
+  GRN_RAW_STRING_INIT(ctx->impl->output.arrow_metadata_label);
```
kou, line 457: We should close all `ctx->impl->output.*` initialization.

lib/proc/proc_select.cpp
```diff
@@ -4553,6 +4553,12 @@ grn_select_output_slice_label_v1(grn_ctx *ctx,
                                  grn_select_data *data,
                                  Slice *slice)
 {
+  if (grn_ctx_get_output_type(ctx) == GRN_CONTENT_APACHE_ARROW) {
+    ctx->impl->output.arrow_metadata_data_type = "slice";
+    ctx->impl->output.arrow_metadata_label = slice->label;
+    return;
+  }
+
   GRN_OUTPUT_STR(slice->label.value, slice->label.length);
```
kou: We should add a logical `grn_output_XXX()` function and use it instead of passing information via `ctx->impl->output`.

kou: This approach breaks the `grn_output_*()` abstraction.

lib/ctx.cpp
otegami: fix: [4e9ca04](https://github.com/groonga/groonga/pull/2399/commits/4e9ca04e8cc85ebd10c7f872f24c2aeefc755345)

lib/ctx.cpp
otegami: fix: [6b3174e](https://github.com/groonga/groonga/pull/2399/commits/6b3174e407fbddde0cdd34deb4313e5b8a340942) Sure.

lib/proc/proc_select.cpp
otegami: fix: [869a3b6](https://github.com/groonga/groonga/pull/2399/commits/869a3b6e2e71c83230af0d5e8a6b9c1dc347b631) I added `grn_output_arrow_metadata()` function as a abstraction layer.

otegami:
Thank you for your review. I’ve addressed all of the review comments here.

lib/proc/proc_select.cpp
kou: Can we abstract more? We want to use `grn_ctx_get_output_type(ctx) == GRN_CONTENT_APACHE_ARROW` only in `output.c`.

otegami:
I think these changes don't work as we expected. I will re-check it.

lib/proc/proc_select.cpp
otegami: fix: [53837a7](https://github.com/groonga/groonga/pull/2399/commits/53837a72db658aac6a345970d5ed7c22865d77c7) Sure. I've extracted the logic of setting the label.

otegami:
Thank you for your review. I’ve addressed all of the review comments here.

lib/grn_output.h
```diff
@@ -90,6 +90,11 @@ grn_obj_format_set_columns(grn_ctx *ctx,
                            const char *columns,
                            unsigned int columns_len);
 
+void
+grn_output_arrow_stream_open_with_metadata(grn_ctx *ctx,
+                                           grn_obj *outbuf,
+                                           grn_content_type output_type);
```
kou, line 96: Why do we need to export this?

lib/proc/proc_select.cpp
kou: Hmm.

I feel that `grn_output_result_set_label()` -> `grn_output_result_set_open_metadata()` order is strange. Because `open` is a constructor name in Groonga. Calling other functions before `open()` is strange. (I feel that this mismatch introduces `ctx->impl->output.arrow_metadata_*`.)

Can we set label as a metadata? Can we use `grn_obj_format` for it?

otegami:
I will use the different approach for this PR. So, I will close this one now.

PR #2402, opened 2025-06-27
docs: api/grn_inspect move grn_thread_set_get_limit_func reference to the header file

otegami:
GitHub: GH-1817

### DIFF
--- a/include/groonga/thread.h
+++ b/include/groonga/thread.h
@@ -86,7 +86,6 @@ typedef uint32_t (*grn_thread_get_limit_func)(void *data);
  *
  * uint32_t max_threads = 100;
  * grn_thread_set_get_limit_func(my_get_thread_limit, &max_threads);
- * grn_init();
  * ```
  *
  * \param func The custom function that returns the max number of threads.
--- a/include/groonga/thread.h
+++ b/include/groonga/thread.h
@@ -49,7 +49,49 @@ grn_thread_get_limit_with_ctx(grn_ctx *ctx);
 GRN_API void
 grn_thread_set_limit_with_ctx(grn_ctx *ctx, uint32_t new_limit);
 
+/**
+ * \brief The type of function that returns the max number of threads.
+ *
+ * \since 5.0.7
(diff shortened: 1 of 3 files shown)


abetomo (approved):

include/groonga/thread.h
```diff
@@ -49,7 +49,49 @@ grn_thread_get_limit_with_ctx(grn_ctx *ctx);
...
+ * static uint32_t
+ * my_get_thread_limit(void *data)
+ * {
+ *   uint32_t *max_threads = (uint32_t *)data;
+ *   return *max_threads;
+ * }
+ *
+ * uint32_t max_threads = 100;
+ * grn_thread_set_get_limit_func(my_get_thread_limit, &max_threads);
+ * grn_init();
```
komainu8: Is `grn_init()` need?

include/groonga/thread.h
```diff
@@ -49,7 +49,49 @@ grn_thread_get_limit_with_ctx(grn_ctx *ctx);
...
+ * ```
```
komainu8, line 89: In my opinion, the above example is artificial example.
How about we show actual use case as below? 

```c
static uint32_t max_threads_in_my_command = 100;

static uint32_t
my_get_thread_limit(void *data)
{
  return max_threads_in_my_command;
}

grn_thread_set_get_limit_func(my_get_thread_limit, NULL);
```

include/groonga/thread.h
otegami: fix: [96901a1](https://github.com/groonga/groonga/pull/2402/commits/96901a1bce51e52c98c2ee32dda166021e291888) Thanks, we don't need it.

include/groonga/thread.h
otegami: The  above example doesn't use the data parameter, so I want to keep using the current example here.

komainu8: I see.

komainu8 (approved):

PR #2407, opened 2025-06-30
ii order-near-phrase: fix interval caclulation

komainu8:
## Problem

Currently, we can not calculate correctly interval between phrase.
We calculate interval between phrase as below expression.

```
interval = pos - previous_pos - token_info->n_tokens_in_phrase;
```

However, the avobe expression can not calculate correctly interval.
Because we use token_info->n_tokens_in_phrase.
We should use the n_tokens_in_phrase of previous phrase.

This problem only occurs when we use max element intervals.
Because if we don't use max element intervals, Groonga does not reach posint that cause the problem.

---

For example, if we calculate interval between phrase "abc defghi jklmnop" in "abc123456defghi".
(We use `TokenNgram("unify_alphabet", false, "unify_digit", false)` and `NormalizerNFKC121` in this case.)

### Calculate interval between "abc" and "defghi":

Curretnnly:

```
interval = pos - previous_pos - token_info->n_tokens_in_phrase;
```

pos is position of first token of second phrase.(So, the head position of "defghi".)
Therefore, pos is 9.

previous_pos is position of first token of first phrase.(So, the head position of "abc".)
Therefore, previous_pos is 0.

token_info->n_tokens_in_phrase is the number of tokens in second phrase.(So, the number of tokens in "defghi".)
Therefore, token_info->n_tokens_in_phrase is 6.
Because "defghi" is tokenized as below.

```
"defghi" -> "de/ef/fg/gh/hi/i"
```

Therefore the interval between "abc" and "defghi" is 9 - 0 - 6 = 3.

"abc123456defghi" is hit.
However, correctly the interval between"abc" and "defghi" is 6.
We expect "abc123456defghi" is not hit.

Revision:

```
interval = pos - previous_pos - previous_n_tokens_in_phrase;
```

pos is position of first token of second phrase.(So, the head position of "defghi".)
Therefore, pos is 9.

previous_pos is position of first token of first phrase.(So, the head position of "abc".)
Therefore, previous_pos is 0.

previous_n_tokens_in_phrase is the number of tokens in first phrase.(So, the number of tokens in "abc".)
Therefore, token_info->n_tokens_in_phrase is 3.
Because "abc" is tokenized as below.

```
"abc" -> "ab/bc/c"
```

Therefore the interval between "abc" and "defghi" is 9 - 0 - 3 = 6.
"abc123456defghi" is not hit.

We can get expected value by using the n_tokens_in_phrase of previous phrase instead of token_info->n_tokens_in_phrase.

This problem does not detect current tests.
Because the number of tokens in phrase are same value in all phrase like "abc def ghi".
So, I update current tests.

### DIFF
--- a/test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_tokens/all.expected
+++ b/test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_tokens/all.expected
@@ -8,13 +8,13 @@ column_create Terms entries_content COLUMN_INDEX|WITH_POSITION Entries content
 [[0,0.0,0.0],true]
 load --table Entries
 [
-{"content": "abc12345def123456ghi"},
-{"content": "abc12345def1234567ghi"},
-{"content": "abc123456def123456ghi"},
-{"content": "abc123456def1234567ghi"}
+{"content": "abc12345defghi123456jklmnop"},
+{"content": "abc12345defghi1234567jklmnop"},
+{"content": "abc123456defghi123456jklmnop"},
+{"content": "abc123456defghi1234567jklmnop"}
 ]
 [[0,0.0,0.0],4]
-select Entries --filter 'content *ONP-1,0,5|6 "abc def ghi"' --output_columns '_score, content'
+select Entries --filter 'content *ONP-1,0,5|6 "abc defghi jklmnop"' --output_columns '_score, content'
 [
   [
     0,
@@ -38,7 +38,7 @@ select Entries --filter 'content *ONP-1,0,5|6 "abc def ghi"' --output_columns '_
       ],
       [
         1,
-        "abc12345def123456ghi"
+        "abc12345defghi123456jklmnop"
       ]
     ]
   ]
--- a/lib/ii.cpp
+++ b/lib/ii.cpp
@@ -13118,6 +13118,8 @@ grn_ii_select_data_check_near_element_intervals(grn_ctx *ctx,
     }
     uint32_t previous_phrase_id = data->token_infos[0]->phrase_id;
     int32_t previous_pos = data->token_infos[0]->pos;
+    int32_t previous_n_tokens_in_phrase =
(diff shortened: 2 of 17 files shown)


test/command/suite/select/filter/ordered_near_phrase/max_element_intervals/multi_tokens/all.expected
```diff
@@ -8,13 +8,13 @@ column_create Terms entries_content COLUMN_INDEX|WITH_POSITION Entries content
...
-{"content": "abc123456def123456ghi"},
-{"content": "abc123456def1234567ghi"}
+{"content": "abc12345defghi123456jklmnop"},
+{"content": "abc12345defghi1234567jklmnop"},
+{"content": "abc123456defghi123456jklmnop"},
+{"content": "abc123456defghi1234567jklmnop"}
 ]
 [[0,0.0,0.0],4]
-select Entries --filter 'content *ONP-1,0,5|6 "abc def ghi"' --output_columns '_score, content'
+select Entries --filter 'content *ONP-1,0,5|6 "abc defghi jklmnop"' --output_columns '_score, content'
```
otegami, line 17: In my current understanding, the expectation wouldn't be changed here because the `abc12345def123456ghi` is satisfied with the following conditions.
`'content *ONP-1,0,5|6 "abc def ghi"'`
- `abc def ghi` must be included in this order.
- The interval between `abc` and `def` is within 5.
- The interval between `def` and `ghi` is within 6.
- After the last phrase `ghi`, there is no interval. 


https://groonga.org/docs/reference/grn_expr/script_syntax.html#ordered-near-phrase-search-operator

otegami: My above understanding is correct. But this time, we want to check the behavior which has the different number tokens in a phrase because all phrases have 3 tokens in the previous expectations. So, we couldn't find the case that the counting was wrong like this issue.

abetomo (approved):

PR #2423, opened 2025-07-11
pat: use pat_node_common instead of pat_node in set_cursor_rk

komainu8:

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -5921,9 +5921,11 @@ set_cursor_rk(grn_ctx *ctx,
   }
   byte_len = rk_conv(key, key_len, keybuf, GRN_TABLE_MAX_KEY_SIZE, &state);
   PAT_AT(pat, 0, node);
-  id = pat_node_get_right(pat, node);
-  if ((id = sub_search(ctx, pat, id, &c0, keybuf, byte_len))) {
-    search_push(ctx, pat, c, keybuf, byte_len, state, id, c0, flags);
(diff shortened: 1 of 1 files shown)


lib/pat.c
```diff
@@ -5912,17 +5912,16 @@ set_cursor_rk(grn_ctx *ctx,
...
+  pat_node_common *node;
   int c0 = -1;
-  uint32_t len, byte_len;
+  uint32_t byte_len;
   uint8_t keybuf[GRN_TABLE_MAX_KEY_SIZE];
   if (flags & GRN_CURSOR_SIZE_BY_BIT) {
     return GRN_OPERATION_NOT_SUPPORTED;
   }
   byte_len = rk_conv(key, key_len, keybuf, GRN_TABLE_MAX_KEY_SIZE, &state);
-  len = byte_len * 16;
```
komainu8, line 5923: `len` does not use in function.
So, I removed this variable.

lib/pat.c
```diff
@@ -5912,17 +5912,16 @@ set_cursor_rk(grn_ctx *ctx,
...
-  PAT_AT(pat, 0, pn);
-  id = pn->lr[1];
+  PAT_AT(pat, 0, node);
+  id = pat_node_get_right(pat, node);
```
otegami: Should we check whether the node exists or not here?

komainu8: I've added `NULL` check.

PR #2424, opened 2025-07-11
pat: use pat_node_common instead of pat_node in grn_pat_del_data structure 

komainu8:

I added the following function temporary.

- _grn_pat_wal_add_entry_data_set_record_direction()
- _pat_node_get_key()
- _grn_pat_next_location()

Because if we modify the above functions, we must modify many place in lib/pat.c to prevent build error.
So, I added the above functions temporary to reduce affect.

The above functions will change to functions that have no "_" prefix at the completion of GH-2349.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -6757,12 +6757,12 @@ grn_pat_wal_recover_delete_entry(grn_ctx *ctx,
     return;
   }
   data.proot = pat_node_get_right_address(pat, root_node);
-  data.p0 = pat_node_get_child_address(pat,
-                                       grandparent_node,
-                                       entry->parent_record_direction);
   data.p = pat_node_get_child_address(pat,
                                       data.node_previous,
                                       entry->record_direction);
+  data.p0 = pat_node_get_child_address(pat,
+                                       grandparent_node,
+                                       entry->parent_record_direction);
   data.n_garbages = entry->n_garbages;
   data.n_entries = entry->n_entries;
   grn_pat_del_internal(ctx, pat, &data);


otegami (commented):
Looks nice to me except for the format 👍🏾 
How about installing the pre-commit?

lib/pat.c
```diff
@@ -6570,11 +6570,26 @@ grn_pat_wal_recover_delete_entry(grn_ctx *ctx,
...
+                              wal_error_tag,
+                              "failed to refer root node");
+    return;
+  }
+  if (pat_is_key_large(pat)) {
+    data.proot = &(root_node->node_large.lr[DIRECTION_RIGHT]);
+    data.p0 =
+      &(grandparent_node->node_large.lr[entry->parent_record_direction]);
+  } else {
+    data.proot = &(root_node->node.lr[DIRECTION_RIGHT]);
```
abetomo: Could we use pat_node_get_right()?

komainu8: Ah...
Thank you. I'll fix it.

lib/pat.c
```diff
@@ -6570,11 +6570,51 @@ grn_pat_wal_recover_delete_entry(grn_ctx *ctx,
...
+    grn_wal_set_recover_error(ctx,
+                              GRN_NO_MEMORY_AVAILABLE,
+                              (grn_obj *)pat,
+                              entry,
+                              wal_error_tag,
+                              "failed to refer root node");
+    return;
+  }
+  /*
+   * We don't use "pat_node_get_right()" to get "data.proot" value.
```
kou: Can we add `pat_node_get_right_address()`?

komainu8: Oh. I see.
I added it.

lib/pat.c
```diff
@@ -6583,11 +6593,25 @@ grn_pat_wal_recover_delete_entry(grn_ctx *ctx,
...
+                              (grn_obj *)pat,
+                              entry,
+                              wal_error_tag,
+                              "failed to refer root node");
+    return;
+  }
+  data.proot = pat_node_get_right_address(pat, root_node);
+  if (pat_is_key_large(pat)) {
+    data.p0 =
+      &(grandparent_node->node_large.lr[entry->parent_record_direction]);
```
kou: Can we add `pat_node_get_child_address()` or something?

lib/pat.c
```diff
@@ -6583,11 +6593,25 @@ grn_pat_wal_recover_delete_entry(grn_ctx *ctx,
...
+  } else {
+    data.p0 = &(grandparent_node->node.lr[entry->parent_record_direction]);
+  }
```
otegami: How about adding `pat_node_get_child_address` and using it here?

otegami: Please ignore this comment because this is duplicated content.

lib/pat.c
```diff
@@ -1238,6 +1350,26 @@ pat_node_get_key(grn_ctx *ctx, grn_pat *pat, pat_node *n)
   }
 }
 
+static inline uint8_t *
+_pat_node_get_key(grn_ctx *ctx, grn_pat *pat, pat_node_common *node)
```
komainu8, line 1346: This function that does the same `pat_node_get_key()`.
If we change `pat_node_get_key()`, we modify many functions.
Because `pat_node_get_key()` uses in many place.
So, I implement function instead of `pat_node_get_key()` temporary.
I will change to `pat_node_get_key()` from `_pat_node_get_key()` in the near feature.

lib/pat.c
```diff
@@ -156,8 +166,82 @@ pat_node_get_left(grn_pat *pat, pat_node_common *node)
...
+pat_node_get_child_address(grn_pat *pat,
+                           pat_node_common *node,
+                           pat_node_direction direction)
+{
+  if (direction == DIRECTION_RIGHT) {
+    return pat_node_get_right_address(pat, node);
+  } else {
+    return pat_node_get_left_address(pat, node);
+  }
+}
```
kou: Can we use `pat_node_get_child_address()` in `pat_node_get_{right,left}_address()`?
If we do it, we'll be able to remove `direction == DIRECTION_RIGHT` condition.

lib/pat.c
```diff
@@ -3487,57 +3619,62 @@ grn_pat_del_internal(grn_ctx *ctx, grn_pat *pat, grn_pat_del_data *data)
   grn_id id = data->id;
   uint16_t check = data->check;
   uint16_t check0 = data->check0;
-  pat_node *rn = data->rn;
-  pat_node *rn0 = data->rn0;
-  pat_node *rno = data->rno;
+  pat_node_common *refer_node = data->rn;
```
kou: Hmm. I don't think that "r" means "refer"... I think that "r" means "root" or "relative".
If "r" is "root", it's not a good name because we reuse "rn" for all nodes.
If "r" is "relative", it's not a good name. We can use "current" or something instead. But I'm not sure "current" is meaningful in this context.

So I think that we can use "node" here because "r" doesn't provide meaningful additional information.

lib/pat.c
```diff
@@ -3487,57 +3619,62 @@ grn_pat_del_internal(grn_ctx *ctx, grn_pat *pat, grn_pat_del_data *data)
...
+  pat_node_common *refer_parent_node = data->rn0;
```
kou: I think that "0" means "previous" in our code.

lib/pat.c
```diff
@@ -3487,57 +3619,62 @@ grn_pat_del_internal(grn_ctx *ctx, grn_pat *pat, grn_pat_del_data *data)
...
+  pat_node_common *refer_otherside_node = data->rno;
```
kou: Using the same prefix ("refer_node_XXX") than putting additional information to middle ("refer_XXX_node") will be better for readability.

```c
refer_node;
refer_node_parent;
refer_node_otherside;
```

```c
refer_node;
refer_parent_node;
refer_otherside_node;
```

lib/pat.c
komainu8: I am using `pat_node_get_child_address()` when direction is decided dynamically such as below.

```c
data.p = pat_node_get_child_address(pat, data.rn0, entry->record_direction);
```

In the above case, `entry->record_direction`'s value is changed dynamically.
So, I think we can't remove `direction == DIRECTION_RIGHT` condition.

kou:
```c
static inline grn_id *
pat_node_get_child_address(grn_pat *pat,
                           pat_node_common *node,
                           pat_node_direction direction)
{
  if (pat_is_key_large(pat)) {
    return &(node->node_large.lr[direction]);
  } else {
    return &(node->node.lr[direction]);
  }
}

static inline grn_id *
pat_node_get_left_address(grn_pat *pat, pat_node_common *node)
{
  return pat_node_get_child_address(pat, node, DIRECTION_LEFT);
}

static inline grn_id *
pat_node_get_right_address(grn_pat *pat, pat_node_common *node)
{
  return pat_node_get_child_address(pat, node, DIRECTION_RIGHT);
}
```

otegami: Ahhh I see. I couldn't come up with it mm

komainu8: Oh. I see.
Thank you.

lib/pat.c
komainu8: I see.
I used "node" instead of "rn", "node_previous" instead of "rn0", and "node_otherside" instead of "rno".

komainu8:
I've reflected the feedback.

otegami:
Build step seems to be failed in CI.

komainu8:
Ah...Thank you.
Unused functions remaining...

lib/pat.c
```diff
@@ -3469,9 +3613,9 @@ typedef struct {
   grn_id id;
   uint16_t check;
   uint16_t check0;
-  pat_node *rn;
-  pat_node *rn0;
-  pat_node *rno;
+  pat_node_common *rn;
+  pat_node_common *rn0;
+  pat_node_common *rno;
```
otegami: 📝 (It's just memo) We want to rename it for readability.

lib/pat.c
```diff
@@ -1307,6 +1431,26 @@ grn_pat_next_location(grn_ctx *ctx,
...
+_grn_pat_next_location(grn_ctx *ctx,
+                       grn_pat *pat,
+                       pat_node_common *node,
+                       const uint8_t *key,
+                       uint16_t check,
+                       uint16_t check_max)
+{
+  if (PAT_CHECK_IS_TERMINATED(check)) {
+    /* check + 1: delete terminated flag and increment bit differences */
+    if (check + 1 < check_max) {
```
otegami, line 1444: 📝 (it's just memo) We want to replace this logic with the named function.

lib/pat.c
komainu8: I've applied it.

lib/pat.c
komainu8: I'll take care of this comment at another PR.

lib/pat.c
```diff
@@ -3487,57 +3631,62 @@ grn_pat_del_internal(grn_ctx *ctx, grn_pat *pat, grn_pat_del_data *data)
...
+         * node_otherside becomes the new first
          * node, is not an output node even if its check is zero.
          */
-        const uint8_t *k = pat_node_get_key(ctx, pat, rno);
+        const uint8_t *k = _pat_node_get_key(ctx, pat, node_otherside);
         int direction = k ? (*k >> 7) : 1;
-        rno->lr[direction] = otherside;
-        rno->lr[!direction] = 0;
+        pat_node_set_child(pat, node_otherside, direction, otherside);
+        pat_node_set_child(pat, node_otherside, !direction, 0);
```
abetomo: この `0` は `GRN_ID_NIL` のほうが良さげ？

komainu8: たしかに。
`GRN_ID_NIL`を使うようにしました。

lib/pat.c
```diff
@@ -3549,66 +3698,71 @@ grn_pat_del_internal(grn_ctx *ctx, grn_pat *pat, grn_pat_del_data *data)
...
         }
-        PAT_DEL_OFF(rn0);
+        pat_node_set_deleting_off(pat, node_previous);
         ddi->stat = DL_PHASE2;
         di->stat = DL_PHASE1;
         di->ld = ddi->ld;
         di->d = id;
         /*
-        PAT_DEL_OFF(rn0);
+        PAT_DEL_OFF(node_previous);
```
abetomo: このコメントアウトの経緯まで把握してないですが、今後は `PAT_DEL_OFF` ではなくて、`pat_node_set_deleting_off` の方を使うと思うので、コメントをまるっと削除でも良いのでは？

abetomo (approved):

lib/pat.c
komainu8: なるほど。
このコメントは、最初のコミットからすでに存在するコメントで、いままで変化がないので削除しようと思います。
https://github.com/groonga/groonga/blob/a4bff03400bf32e00bf874ad7fa34460711b5817/lib/pat.c#L1042

PR #2425, opened 2025-07-14
pat: use pat_node_common instead of pat_node in defrag

komainu8:

I added "_pat_get()" function temporary.
Because if we modify the above functions, we must modify many place in lib/pat.c to prevent build error.
So, I added the above functions temporary to reduce affect.

The above functions will change to functions that have no "_" prefix at the completion of GH-2349.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -1215,6 +1215,16 @@ pat_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
   return grn_io_array_at(ctx, pat->io, SEGMENT_PAT, id, &flags);
 }
 
+static inline pat_node_common *
+_pat_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
+{
+  int flags = GRN_TABLE_ADD;
+  if (id > GRN_ID_MAX) {
+    return NULL;
+  }
+  return grn_io_array_at(ctx, pat->io, SEGMENT_PAT, id, &flags);
+}
+
 /* sis operation */
 
 static inline sis_node *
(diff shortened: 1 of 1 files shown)


otegami (approved):

lib/pat.c
```diff
@@ -158,6 +158,15 @@ pat_node_get_left(grn_pat *pat, pat_node_common *node)
 
 #define PAT_DEL(x) ((x)->bits & PAT_DELETING)
 #define PAT_IMD(x) ((x)->bits & PAT_IMMEDIATE)
+static inline bool
+pat_is_key_immediate(grn_pat *pat, pat_node_common *node)
+{
+  if (pat_is_key_large(pat)) {
+    return (node->node_large.bits & PAT_IMMEDIATE);
+  } else {
+    return (node->node.bits & PAT_IMMEDIATE);
```
otegami: [nits]
```suggestion
    return node->node_large.bits & PAT_IMMEDIATE;
  } else {
    return node->node.bits & PAT_IMMEDIATE;
```

lib/pat.c
```diff
@@ -158,6 +158,15 @@ pat_node_get_left(grn_pat *pat, pat_node_common *node)
 
 #define PAT_DEL(x) ((x)->bits & PAT_DELETING)
 #define PAT_IMD(x) ((x)->bits & PAT_IMMEDIATE)
+static inline bool
+pat_is_key_immediate(grn_pat *pat, pat_node_common *node)
```
kou: We should use `pat_node_` prefix for `pat_node_common` methods.

lib/pat.c
```diff
@@ -1330,6 +1340,33 @@ key_put(grn_ctx *ctx, grn_pat *pat, const uint8_t *key, uint32_t len)
...
+{
+  if (pat_is_key_large(pat)) {
+    node->node_large.key = position;
+  } else {
+    node->node.key = position;
+  }
+}
+
+/*
+ *  We must not call this function when pat_node_is_key_immediate() is true.
```
komainu8:
```suggestion
 * We must not call this function when pat_node_is_key_immediate() is true.
```

lib/pat.c
```diff
@@ -1215,6 +1215,16 @@ pat_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
   return grn_io_array_at(ctx, pat->io, SEGMENT_PAT, id, &flags);
 }
 
+static inline pat_node_common *
+_pat_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
+{
+  int flags = GRN_TABLE_ADD;
+  if (id > GRN_ID_MAX) {
+    return NULL;
+  }
```
otegami: We could skip the unnecessary assignment.
```suggestion
  if (id > GRN_ID_MAX) {
    return NULL;
  }
  int flags = GRN_TABLE_ADD;
```

lib/pat.c
```diff
@@ -1215,6 +1215,16 @@ pat_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
   return grn_io_array_at(ctx, pat->io, SEGMENT_PAT, id, &flags);
 }
 
+static inline pat_node_common *
+_pat_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
```
otegami: Should we use the same prefix `pat_node`?
```suggestion
_pat_node_get(grn_ctx *ctx, grn_pat *pat, grn_id id)
```

lib/pat.c

lib/pat.c
komainu8: I've fixed them.

PR #2427, opened 2025-07-16
bulk: Cast the return value of get()

abetomo:
Since `GRN_DB_INT64` and larger already casts the return value, cast it for other types.

### DIFF
--- a/include/groonga/bulk.hpp
+++ b/include/groonga/bulk.hpp
@@ -83,7 +83,7 @@ namespace grn {
     {
       switch (bulk->header.domain) {
       case GRN_DB_BOOL:
-        return GRN_BOOL_VALUE(bulk);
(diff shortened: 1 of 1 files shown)


include/groonga/bulk.hpp
```diff
@@ -85,17 +85,17 @@ namespace grn {
       case GRN_DB_BOOL:
         return GRN_BOOL_VALUE(bulk);
```
komainu8: Is it needless `static_cast<TYPE>` here?

otegami (approved):

include/groonga/bulk.hpp
abetomo: In many cases, I thought it was enough to convert implicitly from bool types, but I added it just in case.

komainu8:
I'll merge this PR after finishing CI.

PR #2429, opened 2025-07-17
pat: refactor get_tc()

komainu8:
We use more meaningful name to readability.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -2907,30 +2907,30 @@ grn_pat_nextid(grn_ctx *ctx, grn_pat *pat, const void *key, uint32_t key_size)
 static void
 get_transitive_closure(grn_ctx *ctx,
                        grn_pat *pat,
-                       grn_hash *h,
-                       pat_node_common *rn)
+                       grn_hash *hash,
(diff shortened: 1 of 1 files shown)


lib/pat.c
```diff
@@ -2905,29 +2905,32 @@ grn_pat_nextid(grn_ctx *ctx, grn_pat *pat, const void *key, uint32_t key_size)
 }
 
 static void
-get_tc(grn_ctx *ctx, grn_pat *pat, grn_hash *h, pat_node_common *rn)
+get_transitive_closure(grn_ctx *ctx,
+                       grn_pat *pat,
+                       grn_hash *hash,
+                       pat_node_common *root_node)
```
otegami: How about renaming `root_node` to `current_node` (or similar)? After the initial call, it actually refers to the node currently being processed in the recursion, not the true root node.

```suggestion
                       pat_node_common *current_node)
```

komainu8: Oh... I see.
I've fixed it.

otegami (approved):

komainu8:
I'll try again later this modification.

Issue #1641, opened 2023-12-08
Suppress -Wconversion warnings


kou:
### What is your problem?

We need this to update bundled Groonga in MariaDB. MariaDB treats this warning as an error.

We should do this step by step. We must not open one large PR that includes all changes. We can't review it.

abetomo:
Suppress C4244 warnings as well.

It is now disabled.
https://github.com/groonga/groonga/blob/ef1c63b6322e474135e895d69f5762fd6d70e05a/CMakeLists.txt#L387-L390

PR #2441, opened 2025-07-19
grn_error: Cast when assigning to ctx::errlvl

abetomo:
This is part of the work of GitHub GH-1641.

`level` is `grn_log_level`, but `ctx::errlvl` is `unsigned char`, so cast it.

https://github.com/groonga/groonga/blob/57809bdcf99041ad4e7fcca87df2642240f81d05/include/groonga/groonga.h#L354

### DIFF
--- a/include/groonga/groonga.h
+++ b/include/groonga/groonga.h
@@ -187,7 +187,7 @@ typedef enum {
 #define GRN_COMMAND_VERSION_STABLE GRN_COMMAND_VERSION_1
 #define GRN_COMMAND_VERSION_MAX    GRN_COMMAND_VERSION_3
 
-typedef enum {
+typedef enum /* : unsigned char */ {
   GRN_LOG_NONE = 0,
   GRN_LOG_EMERG,
   GRN_LOG_ALERT,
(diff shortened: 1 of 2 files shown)


otegami (approved):

komainu8:
This is minor suggestion.
We can't input value over `255` or under `0`into `grn_log_level`.
Because `errlvl`'s type is `unsigned char`.

So, How about show clearly `grn_log_level`'s value range as below.
 
```c
typedef enum /* : unsigned char */ {
  GRN_LOG_NONE = 0,
  GRN_LOG_EMERG,
  GRN_LOG_ALERT,
  GRN_LOG_CRIT,
  GRN_LOG_ERROR,
  GRN_LOG_WARNING,
  GRN_LOG_NOTICE,
  GRN_LOG_INFO,
  GRN_LOG_DEBUG,
  GRN_LOG_DUMP
} grn_log_level;
```

If we use C23, we can specify `enum`'s type like `typedef enum : unsigned char`.
However, C23 can't use in default. Because C23 is still new.
So, I added `: unsigned char` as a comment in the above example.

However, we don't input value over `255` into `grn_log_level` in the future as well probably.
So, It's OK to apply this suggestion or not apply.

komainu8 (approved):

PR #1520, opened 2023-02-10
near search: add `min_interval`

HashidaTKS:
Add an argument to specify `min_interval` for the near search family.
The intervals of phrases(words) should be greater or equals to `min_interval`.

The default value of `min_interval` is INT32_MIN(`-2147483648`).

New syntax:

```
*N${MAX_INTERVAL},${MAX_TOKEN_INTERVAL_1}|${MAX_TOKEN_INTERVAL_2}|...,${MIN_INTERVAL} "word1 word2 ..."
*NP${MAX_INTERVAL},${ADDITIONAL_LAST_INTERVAL},${MAX_PHRASE_INTERVAL_1}|${MAX_PHRASE_INTERVAL_2}|...,${MIN_INTERVAL} "phrase1 phrase2 ..."
*NPP${MAX_INTERVAL},${ADDITIONAL_LAST_INTERVAL},${MAX_PHRASE_INTERVAL_1}|${MAX_PHRASE_INTERVAL_2}|...,${MIN_INTERVAL} "(phrase1-1 phrase1-2 ...) (phrase2-1 phrase2-2 ...) ..."
*ONP${MAX_INTERVAL},${ADDITIONAL_LAST_INTERVAL},${MAX_PHRASE_INTERVAL_1}|${MAX_PHRASE_INTERVAL_2}|...,${MIN_INTERVAL} "phrase1 phrase2 ..."
*ONPP${MAX_INTERVAL},${ADDITIONAL_LAST_INTERVAL},${MAX_PHRASE_INTERVAL_1}|${MAX_PHRASE_INTERVAL_2}|...,${MIN_INTERVAL} "(phrase1-1 phrase1-2 ...) (phrase2-1 phrase2-2 ...) ..."
```

We should also specify `${MAX_INTERVAL}`, `${ADDITIONAL_LAST_INTERVAL}`, `${MAX_PHRASE_INTERVAL_1}`( exclude `*N`) in order to specify `${MIN_INTERVAL}` because it is added as the 4th argument.

Example:

```

table_create Entries TABLE_NO_KEY
[[0,0.0,0.0],true]
column_create Entries content COLUMN_SCALAR Text
[[0,0.0,0.0],true]
table_create Terms TABLE_PAT_KEY ShortText   --default_tokenizer 'TokenNgram("unify_alphabet", false,                                   "unify_digit", false)'   --normalizer NormalizerNFKC121
[[0,0.0,0.0],true]
[...]

PR #2442, opened 2025-07-23
ii ordered-near-phrase-product: fix interval caclulation

komainu8:
This problem is similar problem to GH-2407.

Currently, we can not calculate correctly interval between phrase in `*ONPP` operator too.

We should use the number of tokens in phrase of previous phrase in `*ONPP` operator too.
However, `*ONPP` operator use the number of tokens in phrase of current phrase in caculating interval between phrases.

So, we modify that use it of previous phrase in this commit.

### DIFF
--- a/test/command/suite/select/filter/ordered_near_phrase_product/min_interval/with_max_element_intervals/match.expected
+++ b/test/command/suite/select/filter/ordered_near_phrase_product/min_interval/with_max_element_intervals/match.expected
@@ -8,10 +8,10 @@ column_create Terms entries_content COLUMN_INDEX|WITH_POSITION Entries content
 [[0,0.0,0.0],true]
 load --table Entries
 [
-{"content": "abc123456789def"}
+{"content": "abc1234567890def"}
 ]
 [[0,0.0,0.0],1]
-select Entries   --filter 'content *ONPP-1,0,10,9"(abc bcd) (def)"'   --output_columns '_score, content'
+select Entries   --filter 'content *ONPP-1,0,10,9"(abc bcd) (de)"'   --output_columns '_score, content'
 [
   [
     0,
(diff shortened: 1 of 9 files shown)


test/command/suite/select/filter/ordered_near_phrase_product/min_interval/with_max_element_intervals/match.expected
```diff
@@ -8,10 +8,10 @@ column_create Terms entries_content COLUMN_INDEX|WITH_POSITION Entries content
 [[0,0.0,0.0],true]
 load --table Entries
 [
-{"content": "abc123456789def"}
+{"content": "abc1234567890def"}
 ]
 [[0,0.0,0.0],1]
-select Entries   --filter 'content *ONPP-1,0,10,9"(abc bcd) (def)"'   --output_columns '_score, content'
+select Entries   --filter 'content *ONPP-1,0,10,9"(abc bcd) (de)"'   --output_columns '_score, content'
```
otegami, line 14: Could we specify the `9` here because the syntax is as follows ?
`column *ONPP${MAX_INTERVAL},${ADDITIONAL_LAST_INTERVAL},${MAX_PHRASE_INTERVAL_1}|${MAX_PHRASE_INTERVAL_2}|...`
```suggestion
select Entries   --filter 'content *ONPP-1,0,10"(abc bcd) (de)"'   --output_columns '_score, content'
```

otegami (commented):
Thanks you so much!
If there is the example about before and after this modification, we can easily review it.

test/command/suite/select/filter/ordered_near_phrase_product/min_interval/with_max_element_intervals/match.expected
komainu8: `9` is `MIN_INTERVAL`. So, the `9` is need in this test.
But this parameter undocumented...

otegami: 📝 We can specify the min_interval as a fourth value here. This feature is undocumented so we need to write the documentation later.
ref: https://github.com/groonga/groonga/pull/1520

otegami (approved):

PR #2452, opened 2025-07-29
cmake: suppress unused-function warning for llama.cpp

otegami:
## Issue

```
/home/buildbot/_deps/llama_cpp-src/ggml/src/ggml-impl.h:111:13: error: 'void ggml_set_op_params(ggml_tensor*, const void*, size_t)' defined but not used [-Werror=unused-function]
  111 | static void ggml_set_op_params(struct ggml_tensor * tensor, const void * params, size_t params_size) {
      |             ^~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

## Cause

When building Groonga with MariaDB, the bundled
llama.cpp triggers `-Werror=unused-function`.

## Solution

Since upstream llama.cpp defines these static helpers even if we don’t call them now, add
`-Wno-unused-function` to Debug and RelWithDebInfo builds to suppress the error.


CMakeLists.txt
```diff
@@ -1685,6 +1685,9 @@ if(NOT "${GRN_WITH_LLAMA_CPP}" STREQUAL "no")
         # https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996
         string(APPEND CMAKE_C_FLAGS " /wd4996")
         string(APPEND CMAKE_CXX_FLAGS " /wd4996")
+      elseif(GRN_C_COMPILER_GNU_LIKE)
+        string(APPEND CMAKE_C_FLAGS_DEBUG " -Wno-unused-function")
+        string(APPEND CMAKE_C_FLAGS_RELWITHDEBINFO " -Wno-unused-function")
```
komainu8, line 1690: Do these option only enable in `grn_build_llama_cpp()` function?

otegami: Yes it is. The `-Wno-unused-function` is applied for only llama.cpp.

```console
$ grep -oP '/_deps/\K[^/]+' build.log | sort -u
h3-build
h3-build 
h3-build && 
h3-build && ccache 
h3-src
h3-src 
llama_cpp-build
llama_cpp-src
msgpack-c-build
msgpack-c-build 
msgpack-c-src
msgpack-c-src 
simdjson-build
simdjson-build 
simdjson-src
simdjson-src 
xsimd-src
$ grep "\-Wno-unused-function" build.log | grep -oP '/_deps/\K[^/]+' | sort -u
llama_cpp-build
llama_cpp-src
```

[build.log](https://github.com/user-attachments/files/21499657/build.log)

komainu8: Thank you for your replay. I see.

PR #2426, opened 2025-07-16
cmake msvc: enable possible loss of data warning

abetomo:
We plan to check and fix the code that causes C4244 warning.
Therefore, we will enable C4244 so that it can be checked.

The build succeeds, but with more warning output.

PR #2460, opened 2025-08-01
cmake: Enable C4244 when GRN_WARN_CONVERSION=ON

abetomo:
C4244 is also a "possible loss of data" warning as well as C4267, so add it.

C4244: 'argument' : conversion from 'type1' to 'type2', possible loss of data
C4267: 'var' : conversion from 'size_t' to 'type', possible loss of data

Add `/W3` because the default is `/W1`, so level2 and level3 are not warnings.

https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level
> /W1 displays level 1 (severe) warnings. /W1 is the default setting in the command-line compiler.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -767,6 +767,7 @@ jobs:
           ccache --show-stats --verbose --version
       - name: Install Groonga
         shell: cmd
+        continue-on-error: ${{ matrix.warn_conversion == 'ON' }}
         run: |
           call "%VC_PREFIX%\Auxiliary\Build\vcvarsall.bat" %VC_ARCHITECTURE%
           set CMAKE_ARGS=-S .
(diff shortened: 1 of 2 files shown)


abetomo:
#2426

.github/workflows/cmake.yml
```diff
@@ -769,19 +774,24 @@ jobs:
...
+          set CMAKE_ARGS=%CMAKE_ARGS% "-DGRN_WARN_CONVERSION=${{ matrix.warn_conversion }}"
           set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_WITH_APACHE_ARROW=ON
           set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_WITH_BLOSC=auto
           set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_WITH_MRUBY=ON
           set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_WITH_XXHASH=bundled
           set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_WITH_ZLIB=no
           set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_WITH_ZSTD=bundled
           set CMAKE_ARGS=%CMAKE_ARGS% -DGROONGA_NORMALIZER_MYSQL_DOC_DIR=share/groonga-normalizer-mysql
+          if "${{ matrix.warn_conversion }}" == "OFF" (
+            set CMAKE_ARGS=%CMAKE_ARGS% -DGRN_ALLOW_WARNING=OFF
```
otegami: Could you tell me why we disable here when we enable warn conversion?
Is it because we don't want to mix two warning here?

kou: We haven't suppressed conversion warnings yet.

We don't want to allow other warnings accidentally. We want to detect the case by a CI failure.

----

How about always enabling `-DGRN_ALLOW_WARNING=ON` and ignore build failures for conversion warnings for now?

otegami (approved):

PR #2461, opened 2025-08-01
cmake: suppress declaration-after-statement warning for llama.cpp

otegami:
## Issue

```
/home/buildbot/_deps/llama_cpp-src/ggml/src/ggml.c: In function 'ggml_quantize_chunk':
/home/buildbot/_deps/llama_cpp-src/ggml/src/ggml.c:6405:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 6405 |     const size_t start_row = start / n_per_row;
      |     ^~~~~
cc1: all warnings being treated as errors
```

## Cause

When building Groonga with MariaDB, the bundled
llama.cpp triggers `-Werror`.

## Solution

Upstream llama.cpp doesn't use C99 standardeven since this project is new project. It is unlikely to
change this style, so we suppress the warning by
explicitly passing -Wno-declaration-after-statement in Debug and RelWithDebInfo builds.


CMakeLists.txt
```diff
@@ -1686,7 +1686,10 @@ if(NOT "${GRN_WITH_LLAMA_CPP}" STREQUAL "no")
         string(APPEND CMAKE_C_FLAGS " /wd4996")
         string(APPEND CMAKE_CXX_FLAGS " /wd4996")
       elseif(GRN_C_COMPILER_GNU_LIKE)
+        string(APPEND CMAKE_C_FLAGS_DEBUG " -Wno-declaration-after-statement")
         string(APPEND CMAKE_C_FLAGS_DEBUG " -Wno-unused-function")
+        string(APPEND CMAKE_C_FLAGS_RELWITHDEBINFO
+               " -Wno-declaration-after-statement")
```
abetomo:
```
        string(APPEND CMAKE_C_FLAGS " /wd4805")
        string(APPEND CMAKE_CXX_FLAGS " /wd4805")

        string(APPEND CMAKE_C_FLAGS " /wd4996")
        string(APPEND CMAKE_CXX_FLAGS " /wd4996")
```

Since the other settings are in the order above, how about doing the same?

```suggestion
        string(APPEND CMAKE_C_FLAGS_DEBUG " -Wno-declaration-after-statement")
        string(APPEND CMAKE_C_FLAGS_RELWITHDEBINFO
               " -Wno-declaration-after-statement")
        string(APPEND CMAKE_C_FLAGS_DEBUG " -Wno-unused-function")
```

otegami: Thanks. I got it!

komainu8 (approved):

PR #2473, opened 2025-08-04
pat: expand embeddable key size

komainu8:

We expand key size in `pat_node_large` to `uint64_t` from `uint32_t` when we specify `KEY_LARGE`.
Therefore, we can embeddable key utile 8byte into `pat_node_large.key` directly.

If key is embedded `pat_node_large.key` directly, `curr_key` is not count. Because if key is embedded `pat_node_large.key` directly, Groonga doesn't use segment.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -128,15 +128,19 @@ pat_is_key_large(grn_pat *pat)
 }
 
 static inline bool
-pat_key_is_embeddable(uint32_t key_size)
+pat_key_is_embeddable(grn_pat *pat, uint32_t key_size)
 {
-  return key_size <= sizeof(uint32_t);
+  if (pat_is_key_large(pat)) {
+    return key_size <= sizeof(uint64_t);
+  } else {
+    return key_size <= sizeof(uint32_t);
+  }
 }
 
 static inline uint32_t
-pat_key_storage_size(uint32_t key_size)
+pat_key_storage_size(grn_pat *pat, uint32_t key_size)
 {
-  return pat_key_is_embeddable(key_size) ? 0 : key_size;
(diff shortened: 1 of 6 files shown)


lib/pat.c
```diff
@@ -125,15 +125,19 @@ pat_is_key_large(grn_pat *pat)
...
+  if (pat_is_key_large(pat)) {
+    return key_size <= sizeof(uint64_t);
+  } else {
+    return key_size <= sizeof(uint32_t);
+  }
 }
 
 static inline uint32_t
-pat_key_storage_size(uint32_t key_size)
+pat_key_storage_size(grn_pat *pat, uint32_t key_size)
```
abetomo, line 141: `uint64_t` ?

komainu8: The `key_size` is one key size( not total key size ).
So, The max value of the `key_size` is 4KiB. Therefore, I don't change the `key_size` type.

abetomo (approved):

test/command/suite/check/pat/large_key_with_embeddable_key.expected
otegami, line 1: Could you teach me how to figure out where the key is embedded?

komainu8: If a key is embedded, `curr_key` that is return value of `check` is `0`.
So, I confirm that a key is embedded for the reason of `curr_key` is `0`.

Did I answer your question?

otegami: I got it! Thank you so much.

otegami (approved):

PR #2474, opened 2025-08-04
pat: remove unused macros

komainu8:

These macros have already replaced inline functions.
So, I removed them.

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -75,7 +75,7 @@ typedef struct {
    */
   grn_id lr[2];
   /*
-    pat_node_is_key_immediate(pat, node) == 0: key bytes offset in memory map.
+    pat_node_is_key_immediate(pat, node) == false: key bytes offset in memory map.
     pat_node_is_key_immediate(pat, node) == 1: the key bytes.
    */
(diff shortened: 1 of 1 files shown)


otegami (approved):

lib/pat.c
```diff
@@ -75,8 +75,8 @@ typedef struct {
    */
   grn_id lr[2];
   /*
-    PAT_IMD(node) == 0: key bytes offset in memory map.
-    PAT_IMD(node) == 1: the key bytes.
+    pat_node_is_key_immediate(pat, node) == 0: key bytes offset in memory map.
```
kou:
```suggestion
    pat_node_is_key_immediate(pat, node) == false: key bytes offset in memory map.
```

lib/pat.c
```diff
@@ -75,8 +75,8 @@ typedef struct {
...
+    pat_node_is_key_immediate(pat, node) == 1: the key bytes.
```
kou:
```suggestion
    pat_node_is_key_immediate(pat, node) == true: the key bytes.
```

PR #2475, opened 2025-08-05
pat: commonalize computation of need number of bits to express a entry size

komainu8:

### DIFF
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -1819,15 +1819,16 @@ _grn_pat_create(grn_ctx *ctx,
   }
   {
     grn_io_array_spec array_spec[3];
-    uint32_t w_of_pat_entry = pat_node_get_w_of_pat_entry(flags);
+    uint32_t segment_width_of_pat_entry = pat_node_get_w_of_pat_entry(flags);
     array_spec[SEGMENT_KEY].w_of_element = 0;
     if (flags & GRN_OBJ_KEY_LARGE) {
       array_spec[SEGMENT_KEY].max_n_segments = GRN_PAT_MAX_N_SEGMENTS_LARGE;
     } else {
       array_spec[SEGMENT_KEY].max_n_segments = GRN_PAT_MAX_N_SEGMENTS;
     }
-    array_spec[SEGMENT_PAT].w_of_element = w_of_pat_entry;
-    array_spec[SEGMENT_PAT].max_n_segments = 1 << (30 - (22 - w_of_pat_entry));
+    array_spec[SEGMENT_PAT].w_of_element = segment_width_of_pat_entry;
+    array_spec[SEGMENT_PAT].max_n_segments =
+      1 << (30 - (22 - segment_width_of_pat_entry));
     array_spec[SEGMENT_SIS].w_of_element = w_of_element;
     array_spec[SEGMENT_SIS].max_n_segments = 1 << (30 - (22 - w_of_element));
     io = grn_io_create_with_array(ctx,
--- a/lib/pat.c
+++ b/lib/pat.c
@@ -1777,6 +1777,24 @@ delinfo_new(grn_ctx *ctx,
 
 /* pat operation */
 
+static inline uint32_t
+pat_node_get_w_of_pat_entry(uint32_t flags)
+{
+  size_t node_size;
+  if (flags & GRN_OBJ_KEY_LARGE) {
+    node_size = sizeof(pat_node_large);
+  } else {
+    node_size = sizeof(pat_node);
+  }
+
+  uint32_t w_of_pat_entry;
+  for (w_of_pat_entry = 0; (size_t)(1 << w_of_pat_entry) < node_size;
+       w_of_pat_entry++) {
+    /* nop */
+  }
+  return w_of_pat_entry;
+}
+
 static inline grn_pat *
 _grn_pat_create(grn_ctx *ctx,
                 grn_pat *pat,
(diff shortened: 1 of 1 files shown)


lib/pat.c
```diff
@@ -1804,16 +1822,15 @@ _grn_pat_create(grn_ctx *ctx,
   }
   {
     grn_io_array_spec array_spec[3];
+    uint32_t w_of_pat_entry = pat_node_get_w_of_pat_entry(flags);
```
abetomo: How about a name like `segment_w_of_element` since it is the value of `array_spec[SEGMENT_PAT].w_of_element`?

lib/pat.c
```diff
@@ -1780,6 +1780,24 @@ delinfo_new(grn_ctx *ctx,
 
 /* pat operation */
 
+static inline uint32_t
+pat_node_get_w_of_pat_entry(uint32_t flags)
```
kou: * We should not use the `pat_node_` prefix here because this is not a method of `pat_node`
* We should not use the `get` here because this doesn't get something

lib/pat.c
```diff
@@ -1780,6 +1780,24 @@ delinfo_new(grn_ctx *ctx,
...
+    node_size = sizeof(pat_node_large);
+  } else {
+    node_size = sizeof(pat_node);
+  }
+
+  uint32_t w_of_pat_entry;
+  for (w_of_pat_entry = 0; (size_t)(1 << w_of_pat_entry) < node_size;
+       w_of_pat_entry++) {
+    /* nop */
+  }
```
kou: If we create a function to compute width of element, can https://github.com/groonga/groonga/blob/31a70fa45525b0fa533f572159c36b06084a4f02/lib/pat.c#L1794-L1797 also reuse the extracted bit width computation feature?

lib/pat.c
komainu8: I used `segment_width_of_element` instead of `w_of_pat_entry`.

lib/pat.c
komainu8: I changed to `compute_bit_width_of_pat_entry` from `pat_node_get_w_of_pat_entry`.

lib/pat.c
komainu8: I've extracted common code into `compute_bit_width_of_pat_entry`.

lib/pat.c
```diff
@@ -1788,31 +1799,32 @@ _grn_pat_create(grn_ctx *ctx,
...
+    array_spec[SEGMENT_PAT].max_n_segments =
+      1 << (30 - (22 - segment_width_of_pat_entry));
+    if (flags & GRN_OBJ_KEY_WITH_SIS) {
+      entry_size = sizeof(sis_node) + value_size;
+    } else {
+      entry_size = value_size;
     }
-    array_spec[SEGMENT_SIS].w_of_element = w_of_element;
-    array_spec[SEGMENT_SIS].max_n_segments = 1 << (30 - (22 - w_of_element));
+    segment_width_of_pat_entry = compute_bit_width_of_pat_entry(entry_size);
```
otegami: The `segment_width_of_pat_entry` is updated here. But the previous code wasn't updated. Is it intentional change?

lib/pat.c
```diff
@@ -1777,6 +1777,17 @@ delinfo_new(grn_ctx *ctx,
 
 /* pat operation */
 
+static inline uint32_t
+compute_bit_width_of_pat_entry(uint32_t entry_size)
```
kou:
```suggestion
compute_needed_bits(uint32_t size_in_byte)
```

lib/pat.c
```diff
@@ -1777,6 +1777,17 @@ delinfo_new(grn_ctx *ctx,
...
+{
+  uint32_t width_of_pat_entry;
```
kou:
```suggestion
  uint32_t needed_bits;
```

lib/pat.c
komainu8: Yes. this is intentional change.

`array_spec[SEGMENT_PAT].w_of_element` and `array_spec[SEGMENT_SIS].w_of_element`are different values.
These values compute in `compute_bit_width_of_pat_entry()`, but `entry_size` of the argument of this function is different if we handle `SEGMENT_PAT` and `SEGMENT_SIS`

We need to update `segment_width_of_pat_entry` because of the process of compute`array_spec[SEGMENT_PAT].w_of_element` and `array_spec[SEGMENT_SIS].w_of_element` were commonalized.

lib/pat.c
```diff
@@ -1788,31 +1799,32 @@ _grn_pat_create(grn_ctx *ctx,
...
-      array_spec[SEGMENT_PAT].w_of_element = 4;
-      array_spec[SEGMENT_PAT].max_n_segments = 1 << (30 - (22 - 4));
+      entry_size = sizeof(pat_node);
+    }
+    needed_bits_of_pat_entry = compute_needed_bits(entry_size);
+    array_spec[SEGMENT_PAT].w_of_element = needed_bits_of_pat_entry;
+    array_spec[SEGMENT_PAT].max_n_segments =
+      1 << (30 - (22 - needed_bits_of_pat_entry));
+    if (flags & GRN_OBJ_KEY_WITH_SIS) {
+      entry_size = sizeof(sis_node) + value_size;
```
kou: We should not reuse `entry_size` for `SEGMENT_PAT` and `SEGMENT_SIS` for readability.

lib/pat.c
```diff
@@ -1788,31 +1799,32 @@ _grn_pat_create(grn_ctx *ctx,
...
+    } else {
+      entry_size = value_size;
     }
-    array_spec[SEGMENT_SIS].w_of_element = w_of_element;
-    array_spec[SEGMENT_SIS].max_n_segments = 1 << (30 - (22 - w_of_element));
+    needed_bits_of_pat_entry = compute_needed_bits(entry_size);
```
kou: We should not reuse `needed_bits_of_pat_entry` for `SEGMENT_SIS`. It's for `SEGMENT_PAT`. (See `pat` in `needed_bits_of_pat_entry`.)

lib/pat.c
komainu8: I used `sis_entry_size` instead of `entry_size`.

lib/pat.c
komainu8: I used `needed_bits_of_sis_entry` instead of `needed_bits_of_pat_entry`.

lib/pat.c
```diff
@@ -1788,31 +1799,33 @@ _grn_pat_create(grn_ctx *ctx,
   grn_io *io;
   pat_node_common *node0;
   struct grn_pat_header *header;
-  uint32_t entry_size, w_of_element;
+  uint32_t entry_size;
```
otegami: [nits] How about naming it for consistency with sis_entry_size?
```suggestion
  uint32_t pat_entry_size;
```

otegami (approved):

komainu8:
I'll marge after CI passed.

PR #2483, opened 2025-08-15
cmake: suppress no-unused-function warning for llama.cpp

otegami:
## Issue

```
/home/buildbot/_deps/llama_cpp-src/ggml/src/ggml-impl.h:111:13: error: 'void ggml_set_op_params(ggml_tensor*, const void*, size_t)' defined but not used [-Werror=unused-function]
  111 | static void ggml_set_op_params(struct ggml_tensor * tensor, const void * params, size_t params_size) {
      |             ^~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

## Cause

When building Groonga with MariaDB, the bundled llama.cpp triggers `-Werror`.

## Solution

Upstream llama.cpp doesn't setup unused-function, so we suppress the warning by explicitly passing `-Wno-unused-function` in Debug and RelWithDebInfo builds.


CMakeLists.txt
```diff
@@ -1701,6 +1701,8 @@ if(NOT "${GRN_WITH_LLAMA_CPP}" STREQUAL "no")
         string(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " -Wno-frame-larger-than")
         string(APPEND CMAKE_CXX_FLAGS_DEBUG " -Wno-non-virtual-dtor")
         string(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " -Wno-non-virtual-dtor")
+        string(APPEND CMAKE_CXX_FLAGS_DEBUG " -Wno-unused-function")
+        string(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " -Wno-unused-function")
```
abetomo, line 1705: Same as lines 1698-1699?

otegami: Yes, the target are different. Those lines support for `CMAKE_C_FLAGS_DEBUG`.
The current changes support for `CMAKE_CXX_FLAGS_DEBUG`.

abetomo: They looked like the same string, but I see there's a difference in `XX`.

abetomo: How about sorting by effect?

```
        string(APPEND CMAKE_C_FLAGS_DEBUG " -Wno-unused-function")
        string(APPEND CMAKE_C_FLAGS_RELWITHDEBINFO " -Wno-unused-function")
        string(APPEND CMAKE_CXX_FLAGS_DEBUG " -Wno-unused-function")
        string(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " -Wno-unused-function")

        string(APPEND CMAKE_CXX_FLAGS_DEBUG " -Wno-frame-larger-than")
        string(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " -Wno-frame-larger-than")
...
```

みたいな。

otegami: fix: [dc78a18](https://github.com/groonga/groonga/pull/2483/commits/dc78a18f00ad6de6af8eea13fd4197991f4a6404) Sure! The Windows options are handled in the same way. So I've just aligned them.

otegami:
Thank you for reviewing. I’ve just addressed all of the review comments here

Issue #2484, opened 2025-08-20
MIssing a "KEY_LARGE" flag from table after executing a "truncate" command.


komainu8:
### What happend?

A "KEY_LARGE" flag is miss after we execute "truncate" command against "TABLE_PAT_KEY" and "TABLE_HASH_KEY" that have a "KEY_LARGE" flag.

### How to reproduce it

- `TABLE_PAT_KEY`

```
> table_create Diaries TABLE_PAT_KEY|KEY_LARGE ShortText
> truncate Diaries
> dump
```

- `TABLE_HASH_KEY`

```
> table_create Diaries TABLE_HASH_KEY|KEY_LARGE ShortText
> truncate Diaries
> dump
```

### Expected behavior

We expect that a "KEY_LARGE" flag remain in table even after executing a "truncate" command.

### Environment

- OS: Debian
- OS Version: 12
- Groonga version: main branch

PR #2485, opened 2025-08-20
table: use value that can be expressed within uint16_t

komainu8:
Fix: Gh-2484

Because the flag value of TABLE_PAT_KEY and TABLE_HASH_KEY are also used as grn_db_obj.header.

Groonga does not refer grn_pat.header, but grn_db_obj.header when Groonga execute a "truncate" command.

However, the type of grn_db_obj.header is uint16_t. So, we must set the flag value of TABLE_PAT_KEY and TABLE_HASH_KEY within uint16_t.

### DIFF
--- a/include/groonga/groonga.h
+++ b/include/groonga/groonga.h
@@ -819,7 +819,7 @@ typedef uint32_t grn_column_flags;
 
 /* flags only for grn_table_flags */
 
-#define GRN_OBJ_KEY_LARGE (0x01 << 16)
+#define GRN_OBJ_KEY_LARGE (0x01 << 11)
 
 /* flags only for grn_column_flags */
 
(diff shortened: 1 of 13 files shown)


include/groonga/groonga.h
```diff
@@ -819,7 +819,7 @@ typedef uint32_t grn_column_flags;
 
 /* flags only for grn_table_flags */
 
-#define GRN_OBJ_KEY_LARGE (0x01 << 16)
+#define GRN_OBJ_KEY_LARGE (0x01 << 11)
```
kou, line 822: We can't accept this because this breaks backward compatibility.

See also the above comment that "flags only for grn_table_flags". We should use 16 or larger for table only flags like we did for `grn_column_flags` and column only flags.

komainu8: Ah.. I see. 
I close this PR. I'll start again.

PR #2486, opened 2025-08-20
table: use the header of table object instead of the header of grn_obj

komainu8:
Fix: GH-2484

Groonga refers to the header of grn_obj when executing a truncate command.
However, the header of `grn_obj` does not include the flag information of individual tables.

Currently, Groonga cannot set the correct flags when executing a truncate command on a table.
As a result, `TABLE_PAT_KEY` and `TABLE_HASH_KEY` lose the `KEY_LARGE` flag after the truncate command is executed.

With this modification, Groonga now refers to the header of each individual table.


lib/hash.c
```diff
@@ -2263,7 +2263,7 @@ grn_hash_truncate(grn_ctx *ctx, grn_hash *hash)
   }
   key_size = hash->key_size;
   value_size = hash->value_size;
-  flags = hash->obj.header.flags;
+  flags = hash->header.common->flags;
```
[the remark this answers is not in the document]
komainu8: Header layout of `grn_hash` and `grn_pat` is different as below.

---

`grn_pat

```
struct _grn_pat {
...
  struct grn_pat_header *header;
...
};

struct grn_pat_header {
  uint32_t flags;
...
};
```

`grn_hash`

```
struct _grn_hash {
...
  union {
    grn_hash_header_common *common;
    grn_hash_header_normal *normal;
    grn_hash_header_large  *large;
  } header;
...
};

#define GRN_HASH_HEADER_COMMON_FIELDS\
  uint32_t flags;\
  grn_encoding encoding;\
  uint32_t key_size;\
  uint32_t value_size;\
  grn_id tokenizer;\
  uint32_t curr_rec;\
  uint32_t curr_key_normal;\
  uint32_t idx_offset;\
  uint32_t entry_size;\
  uint32_t max_offset;\
  uint32_t n_entries;\
  uint32_t n_garbages;\
  uint32_t lock;\
  grn_id normalizer;\
  uint32_t truncated;\
  uint64_t curr_key_large;\
  uint64_t wal_id;\
  uint32_t reserved[10]

struct _grn_hash_header_common {
  GRN_HASH_HEADER_COMMON_FIELDS;
};
```

abetomo (approved):

lib/hash.c
otegami, line 2266: Do we need to preserve the existing hash->obj.header.flags, or is it okay to overwrite them?

otegami (commented):
[nits] Fix the minor grammatical errors in the PR description.
```
Groonga refers to the header of grn_obj when executing a truncate command.
However, the header of `grn_obj` does not include the flag information of individual tables.

Currently, Groonga cannot set the correct flags when executing a truncate command on a table.
As a result, `TABLE_PAT_KEY` and `TABLE_HASH_KEY` lose the `KEY_LARGE` flag after the truncate command is executed.

With this modification, Groonga now refers to the header of each individual table.
```

lib/hash.c
otegami: Made a call with @komainu8 
It’s fine to overwrite them.
The `hash->header.common->flags` already includes the existing information, so nothing will be lost.

PR #2490, opened 2025-08-20
NormalizerNFKC unify_iteration_mark: support single ideographic iteration mark (U+3005)

otegami:
GitHub: GH-2487

Add support for the ideographic iteration mark 々(U+3005) to the unify_iteration_mark option.
This allows proper normalization of Japanese kanji reduplication like 時々 → 時時.

### DIFF
--- a/lib/nfkc.c
+++ b/lib/nfkc.c
@@ -74,6 +74,7 @@ grn_nfkc_normalize_options_init(grn_ctx *ctx,
   options->unify_kana_hyphen = false;
   options->unify_kana_prolonged_sound_mark = false;
   options->unify_katakana_trailing_o = false;
+  options->unify_iteration_mark = false;
   options->unify_to_romaji = false;
   options->unify_to_katakana = false;
   options->remove_blank = false;
@@ -396,6 +397,12 @@ grn_nfkc_normalize_options_apply(grn_ctx *ctx,
                                     raw_options,
                                     i,
                                     options->unify_katakana_trailing_o);
+    } else if (GRN_RAW_STRING_EQUAL_CSTRING(name_raw, "unify_iteration_mark")) {
+      options->unify_iteration_mark =
+        grn_vector_get_element_bool(ctx,
+                                    raw_options,
+                                    i,
+                                    options->unify_iteration_mark);
     } else if (GRN_RAW_STRING_EQUAL_CSTRING(name_raw, "unify_to_romaji")) {
       options->unify_to_romaji =
         grn_vector_get_element_bool(ctx,
--- a/lib/normalizer.c
+++ b/lib/normalizer.c
@@ -3429,6 +3429,57 @@ grn_nfkc_normalize_unify_romaji(grn_ctx *ctx,
 }
 
 static const unsigned char *
+grn_nfkc_normalize_unify_iteration_mark(grn_ctx *ctx,
+                                        const unsigned char *start,
+                                        const unsigned char *current,
+                                        const unsigned char *end,
+                                        size_t *n_used_bytes,
+                                        size_t *n_used_characters,
+                                        unsigned char *unified_buffer,
+                                        size_t *n_unified_bytes,
+                                        size_t *n_unified_characters,
+                                        void *user_data)
+{
+  size_t char_length;
+
+  char_length = (size_t)grn_charlen_(ctx, current, end, GRN_ENC_UTF8);
+
+  *n_used_bytes = char_length;
+  *n_used_characters = 1;
+
+  /* U+3005 IDEOGRAPHIC ITERATION MARK 々 */
+  if (char_length == 3 && current[0] == 0xe3 && current[1] == 0x80 &&
+      current[2] == 0x85) {
+    if (current > start) {
+      const unsigned char *prev = current - 1;
+
+      while (prev > start && (*prev & 0xc0) == 0x80) {
+        prev--;
+      }
+
+      size_t prev_char_length =
+        (size_t)grn_charlen_(ctx, prev, current, GRN_ENC_UTF8);
+
+      if (prev_char_length > 0) {
+        size_t i;
+        for (i = 0; i < prev_char_length; i++) {
+          unified_buffer[(*n_unified_bytes)++] = prev[i];
+        }
+        (*n_unified_characters)++;
+        *n_used_bytes = prev_char_length;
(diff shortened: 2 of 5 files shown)


lib/normalizer.c
```diff
@@ -3428,6 +3428,23 @@ grn_nfkc_normalize_unify_romaji(grn_ctx *ctx,
...
+                                        const unsigned char *end,
+                                        size_t *n_used_bytes,
+                                        size_t *n_used_characters,
+                                        unsigned char *unified_buffer,
+                                        size_t *n_unified_bytes,
+                                        size_t *n_unified_characters,
+                                        void *user_data)
+{
+  /* TODO: Implement iteration mark (踊り字) expansion.
+   * For now, just return the current position unchanged. */
```
kou: How about implementing the easiest one pattern instead of nothing?
It seems that we can't test whether this works or not without implementation.

lib/normalizer.c
```diff
@@ -3852,6 +3869,23 @@ grn_nfkc_normalize_unify(grn_ctx *ctx, grn_nfkc_normalize_data *data)
...
+    if (need_swap) {
+      grn_nfkc_normalize_context_swap(ctx, &(data->context), &unify);
+      grn_nfkc_normalize_context_rewind(ctx, &unify);
+    }
+    grn_nfkc_normalize_unify_stateful(ctx,
+                                      data,
+                                      &unify,
+                                      grn_nfkc_normalize_unify_iteration_mark,
+                                      NULL,
+                                      "[unify][iteration_mark]");
```
kou:
```suggestion
                                      "[unify][iteration-mark]");
```

lib/normalizer.c
```diff
@@ -3428,6 +3428,93 @@ grn_nfkc_normalize_unify_romaji(grn_ctx *ctx,
...
+        iteration_count++;
+        pos += pos_char_length;
+      } else {
+        break;
+      }
+    }
+
+    if (current > start && iteration_count > 0) {
+      const unsigned char *search_pos = current;
+      const unsigned char *char_starts[iteration_count];
```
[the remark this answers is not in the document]
askdkc: 「Groonga 本体では 基本的に VLA は避けて grn_malloc を使う方針 が良いです。」

とAIも言い出すくらいに、GroongaもAIに学習されてるのね👀

test/command/suite/normalizers/nfkc/unify_iteration_mark/ideographic/reduplication.expected
```diff
@@ -0,0 +1,19 @@
+normalize   'NormalizerNFKC("unify_iteration_mark", true)'   "部分々々"   WITH_TYPES
```
komainu8: Should we also use "version" option as below?
`NormalizerNFKC("unify_iteration_mark", true, "version", "NFKC_VERSION")`

komainu8:
Some tests are failure. ( e.g. https://github.com/groonga/groonga/actions/runs/17231634694/job/48886774915?pr=2490 )
We will need to fix them.

lib/nfkc.c
```diff
@@ -396,6 +397,12 @@ grn_nfkc_normalize_options_apply(grn_ctx *ctx,
                                     raw_options,
                                     i,
                                     options->unify_katakana_trailing_o);
+    } else if (GRN_RAW_STRING_EQUAL_CSTRING(name_raw, "unify_iteration_mark")) {
```
komainu8, line 400: Should we use `raw_name` instead of `name_raw` aligned with `raw_options` ( https://github.com/groonga/groonga/pull/2490/files#diff-3fa3dff3156d509df3c960fe781da737f6e1a2b511d45d204191eea2c94f1032R403 )?

kou:
Can we focus on only the 時々 → 時時 case in this PR to make this PR more smaller?

otegami:
Thank you for reviewing it. I've addressed all of the review comments here.

otegami:
I will close this PR now because supporting `々` has two more cases and it's too complicated at first step.
- 部分々々 -> 部分部分
- 古々々米 -> 古古古米

lib/normalizer.c
```diff
@@ -3428,6 +3428,57 @@ grn_nfkc_normalize_unify_romaji(grn_ctx *ctx,
...
+
+  /* U+3005 IDEOGRAPHIC ITERATION MARK 々 */
+  if (char_length == 3 && current[0] == 0xe3 && current[1] == 0x80 &&
+      current[2] == 0x85) {
+    if (current > start) {
+      const unsigned char *prev = current - 1;
+
+      while (prev > start && (*prev & 0xc0) == 0x80) {
+        prev--;
+      }
```
kou, line 3458: In general, we should not backward characters in UTF-8. We can detect the first byte by the top 2 bits but it has overhead like this.

Can we keep the previous character length in `user_data` like `grn_nfkc_normalize_unify_kana_prolonged_sound_mark_like()` does?

lib/normalizer.c
```diff
@@ -3428,6 +3428,57 @@ grn_nfkc_normalize_unify_romaji(grn_ctx *ctx,
...
+        (size_t)grn_charlen_(ctx, prev, current, GRN_ENC_UTF8);
+
+      if (prev_char_length > 0) {
+        size_t i;
+        for (i = 0; i < prev_char_length; i++) {
+          unified_buffer[(*n_unified_bytes)++] = prev[i];
+        }
+        (*n_unified_characters)++;
+        *n_used_bytes = prev_char_length;
+        (*n_used_characters)++;
```
kou, line 3470: Is this correct?

test/command/suite/normalizers/nfkc/unify_iteration_mark/ideographic/single.test
```diff
@@ -0,0 +1,6 @@
+#@add-substitution /"version",\s"(?:.+?)"/ "\"version\", \"#{ENV['NFKC'] || '10.0.0'}\"" "\"version\", \"NFKC_VERSION\""
+normalize \
+  'NormalizerNFKC("unify_iteration_mark", true)' \
+  "時々" \
```
kou, line 4: Could you add at least one character before and after "時々" to detect wrong N bytes/characters computation?

otegami:
Thank you for review comments. I've addressed the above comments in new GH-2499.

Issue #2493, opened 2025-08-25
grndb: misses `${DB}.0000000` check


kou:
### What happend?

It seems that `${DB}.0000000` check misses:

```diff
diff --git a/lib/mrb/scripts/command_line/grndb.rb b/lib/mrb/scripts/command_line/grndb.rb
index 55bcc435d..c72486bcd 100644
--- a/lib/mrb/scripts/command_line/grndb.rb
+++ b/lib/mrb/scripts/command_line/grndb.rb
@@ -538,6 +538,7 @@ module Groonga
 
           suffixes = [
             "",
+            ".0000000",
             ".001",
             ".conf",
             ".options",
```

### How to reproduce it

Not tested. I just noticed this when I read the `grndb` source.

### Expected behavior

Detect problems in `${DB}.0000000` too.

### Environment

- OS: Not related
- OS Version: Not related
- Groonga version: main

PR #2494, opened 2025-08-25
grndb: add a missing check of "${DB}.0000000"

komainu8:
Fix: GH-2493.

### DIFF
--- a/test/command_line/suite/grndb/test_check.rb
+++ b/test/command_line/suite/grndb/test_check.rb
@@ -897,6 +897,34 @@ Empty file exists: <#{empty_file_path_no_object}>
                    ])
     end
 
+    def test_database_objects
+      remove_groonga_log
+      FileUtils.touch("#{@database_path}.0000000",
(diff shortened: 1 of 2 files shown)


test/command_line/suite/grndb/test_check.rb
```diff
@@ -897,6 +897,34 @@ def test_database_keys
                    ])
     end
 
+    def test_database_objects
```
kou: Ah, `.0000000` is a file for specs...

test/command_line/suite/grndb/test_check.rb
```diff
@@ -897,6 +897,34 @@ def test_database_keys
...
+                   [
+                     result.output,
+                     result.error_output,
+                     normalized_groonga_log_content,
+                   ])
+    end
+
     def test_database_specs
       remove_groonga_log
       FileUtils.touch("#{@database_path}.001",
```
kou, line 930: `.001` (`.XXX`) is a sub file of keys (DAT).

Issue #2503, opened 2025-08-29
release: GitHub workflows use incorrect version when we specified version with `rake release`


otegami:
### What happend?

When releasing Groonga with a specific version using `VERSION=x.x.x rake release`,
the GitHub Actions workflows fail to use the correct version for package creation.
This happens because the workflows read the version from the `base_version` file,
which is not updated during the release process when the `VERSION` environment variable is used.

This issue becomes particularly problematic when skipping versions (e.g., going from 15.1.1 directly to 15.1.3),
resulting in packages being created with the wrong version number.

Here are the affected workflow files and lines which are using ``version=$(cat base_version)``:
- `.github/workflows/package.yml`
- `.github/workflows/cmake.yml`
- `.github/workflows/document.yml`

### How to reproduce it

For example:

1. Have `base_version` file containing `15.1.1`
2. Run `VERSION=15.1.3 rake release` to skip version 15.1.2
3. The release task will:
   - Update package versions to 15.1.3
   - Create git tag v15.1.3
   - Update `base_version` to 15.1.4 (next development version)
4. When GitHub Actions workflows run for the v15.1.3 tag, they read `base_version` (which could be 15.1.1 or 15.1.4 depending on timing) instead of using 15.1.3
5. Packages are created with the wrong version

### Expected behavior

GitHub Actions workflows should respect the specifed version when we execute the `rake release` command.

### Environment

GitHub Actions workflows

kou:
We should report an error on `VERSION=15.1.3 rake release`.

otegami:
Since Mroonga’s version may be ahead, there could be cases where we want to intentionally skip Groonga’s version and release with a higher version number than Mroonga.
In that case, is the idea to run rake release multiple times until we reach the desired version?

otegami:
I'm thinking of the following changes to resolve this issue.

```diff
git diff
diff --git a/Rakefile b/Rakefile
index 28ff0ecfa..3d94434aa 100644
--- a/Rakefile
+++ b/Rakefile
@@ -214,6 +214,12 @@ namespace :release do
       raise "failed to detect deb package version: #{changelog}"
     end
 
+    if ENV["VERSION"]
+      File.write("base_version", version)
+      sh("git", "add", "base_version")
+      sh("git", "commit", "-m", "Skip to #{version}")
+    end
+
     sh("git",
        "tag",
        "v#{version}",
```

kou:
```diff
diff --git a/Rakefile b/Rakefile
[...]

PR #2504, opened 2025-08-29
release: prevent `VERSION=` usage with rake release when releasing skipped version

otegami:
GitHub: GH-2503

When `VERSION=x.x.x rake release` was used to skip versions,
GitHub Actions workflows would read the wrong version from base_version file,
which wasn't updated during the release process.
This caused packages to be built with incorrect version numbers.

Now developers must use `rake dev:version:bump NEW_VERSION=x.x.x`  before `rake release` to properly update `base_version`, ensuring workflows use the correct version.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -203,7 +203,7 @@ namespace :release do
   end
 
   desc "Tag"
-  task :tag do
+  task tag: "version:validate" do
     latest_news = Dir.glob("doc/source/news/*.*").max do |a, b|
       File.basename(a).to_f - File.basename(b).to_f
     end


otegami:
```console
$ VERSION=15.1.8 rake release
git ls-files
rake aborted!
You must NOT specify VERSION= for 'rake release'. You must run 'rake dev:version:bump NEW_VERSION=15.1.8 before 'rake release'.
/home/otegami/work/c/groonga/Rakefile:155:in 'block (3 levels) in <top (required)>'
Tasks: TOP => release => release:version:update
(See full trace by running task with --trace)
```

Rakefile
```diff
@@ -148,6 +148,13 @@ namespace :release do
   namespace :version do
     desc "Update versions for a new release"
     task :update do
+      if version != BASE_VERSION
+        message = "You must NOT specify VERSION= for 'rake release'. "
+        message += "You must run 'rake dev:version:bump NEW_VERSION=#{version}'"
```
kou:
```suggestion
        message += "You must run 'rake dev:version:bump NEW_VERSION=#{version}' "
```

kou (commented):
@komainu8 @abetomo What approach do you prefer?

abetomo:
I was planning to check the specified version and base_version.
But I think it will be checked in `release` task, not `release:version:update` task.

komainu8:
In my opinion, If we will execute `VERSION=x.x.x rake release`, we will also update `base_version` into specify version and push to repository as below.

1. Execute `VERSION=x.x.x rake release`
2. Confirm the `base_version` and the `VERSION`(the environment value) whether  same value or not
3. If these values are different, we overwrite the `base_version` to the `VERSION`

I think that I am easier to use than I execute `rake dev:version:bump NEW_VERSION=#{version}` and `VERSION=x.x.x rake release` individually.

kou:
In general, "difficult to do wrong thing" approach is better.

If we want to make `VERSION=X.Y.Z rake release` is the standard release process, the force overwriting approach is fine.
If we don't want to make `VERSION=X.Y.Z rake release` is the standard release process, I feel that we should prohibit it. In general, exceptional processes will cause accidents.

komainu8:
> In general, "difficult to do wrong thing" approach is better.
> 
> If we want to make `VERSION=X.Y.Z rake release` is the standard release process, the force overwriting approach is fine. If we don't want to make `VERSION=X.Y.Z rake release` is the standard release process, I feel that we should prohibit it. In general, exceptional processes will cause accidents.

I see.
I don't want to make VERSION=X.Y.Z rake release is the standard release process.

otegami:
@abetomo 

I understand your opinion about placing the check in the release task instead of `release:version:update`. I initially placed it in `release:version:update` because I felt it semantically fits there, since this task is responsible for handling version-related updates, it seemed natural for it to also validate version consistency.

Rakefile
otegami: fix: [a64af44](https://github.com/groonga/groonga/pull/2504/commits/a64af4434c925c5ccb1f017f062905a04d4bdd22)

abetomo:
The task that fails when `VERSION=` and `base_version` do not match is the `release:tag` task.
So, it seems better to check within that task.

But since `release:version:update` runs first when we run `rake release`, it still seems safer to check at the first step when `rake release` is run.

`VERSION=` と `base_version` が一致しなくて失敗するのは、 `release:tag` です。
（Rakeタスクが失敗するのではなくて、tagのpushで動き出したCIが失敗する。）
ので、チェックするならそこがよさそう。
だけど、`rake release` をすると、 先に `release:version:update` が動くので、やはり、`rake release` をしたときの一番最初でチェックするのが無難なように思えます。
（`VERSION=` と `base_version` が一致しなくて失敗するのは、`release:version:update` ではないので、そこでチェックをするのに違和感があります。）

otegami:
@abetomo 

fix: [442d087](https://github.com/groonga/groonga/pull/2504/commits/442d0879186fab11b7313dd52ac4186ad8cdcc92) Sure. I added the new task to validate the release version and kick it at the beginning of  `rake release`.

Rakefile
```diff
@@ -306,6 +316,7 @@ end
 
 desc "Release"
 task release: [
+  ":release:version:validate",
```
kou:
```suggestion
  "release:version:validate",
```

Rakefile
```diff
@@ -165,6 +165,16 @@ namespace :release do
...
+
+    desc "Validate version for release"
+    task :validate do
+      if version != BASE_VERSION
+        message = "You must NOT specify VERSION= for 'rake release'. "
+        message += "You must run 'rake dev:version:bump NEW_VERSION=#{version}' "
+        message += "before 'rake release'."
+        raise message
+      end
+    end
```
kou: Could you move this before `task :update`?

Rakefile
otegami: fix: [78f5f18](https://github.com/groonga/groonga/pull/2504/commits/78f5f187d548df8af6918d3c2bd5dc09f9b7ea7d)

Rakefile
otegami: fix: [8b0ad9f](https://github.com/groonga/groonga/pull/2504/commits/8b0ad9f10d4eb4094fa14c2b04e9c7a3f1596aaa) Sure.

otegami:
Thank you for reviewing. I’ve just addressed all of the review comments here.

otegami:
> Rakeタスクって依存関係みたいなの設定できなかったですか？
validateが必要なタスクに設定してみてはどうかと思いました。そこまでやらなくても良さげ？

fix: [436a1cb](https://github.com/groonga/groonga/pull/2504/commits/436a1cbfc8ec2c71df379171863b3f1a205045ee) より安全に倒す形で設定してみました！

```console
$ VERSION=15.1.8 rake release
git ls-files
rake aborted!
You must NOT specify VERSION= for 'rake release'. You must run 'rake dev:version:bump NEW_VERSION=15.1.8' before 'rake release'.
/home/otegami/work/c/groonga/Rakefile:155:in 'block (3 levels) in <top (required)>'
Tasks: TOP => release => release:version:update => release:version:validate
(See full trace by running task with --trace)
```

otegami:
I've just rebased now. I didn't change any codes.

Rakefile
```diff
@@ -146,8 +146,18 @@ end
...
+
     desc "Update versions for a new release"
-    task :update do
+    task update: :validate do
```
abetomo, line 160: Could you add the same thing to `release:tag` as well?

otegami: fix: [3838f2e](https://github.com/groonga/groonga/pull/2504/commits/3838f2e50256d87fa5a2c16980eb2e3651833b25) Sure!

```console
$ VERSION=15.1.8 rake release:tag
git ls-files
rake aborted!
You must NOT specify VERSION= for 'rake release'. You must run 'rake dev:version:bump NEW_VERSION=15.1.8' before 'rake release'.
/home/otegami/work/c/groonga/Rakefile:155:in 'block (3 levels) in <top (required)>'
Tasks: TOP => release:tag => release:version:validate
(See full trace by running task with --trace)
```

PR #2511, opened 2025-09-01
grndb: Don't stop processing by File.stat error

kou:


lib/mrb/scripts/command_line/grndb.rb
```diff
@@ -5,6 +5,33 @@ module CommandLine
...
+        def file_mtime(path)
+          begin
+            stat = File.stat(path)
+          rescue SystemCallError => error
+            failed("Failed to get last modified time: <#{path}>",
+                   error.message)
+            Time.at(0)
+          else
+            stat.mtime
+          end
```
otegami, line 18: 📝 Although I wrote the code like this before, the current way is better since it minimizes the scope where an error can occur.
```suggestion
          begin
            stat = File.stat(path)
            stat.mtime
          rescue SystemCallError => error
            failed("Failed to get last modified time: <#{path}>",
                   error.message)
            Time.at(0)
          end
```

otegami (approved):

PR #2528, opened 2025-09-09
package: Use only `gpg_uid_rsa4096`

abetomo:
`gpg_uid` is because it is a low strength key.
Since we've been using `gpg_uid_rsa4096` recently, removing `gpg_uid` will have no impact.


abetomo:
The gpg_uid file itself is also used by GNU Autotools, so we have not deleted it.

otegami:
📝 `gpg_uid` is used as follows.
https://github.com/groonga/groonga/blob/4086f5ac37d2601696a561896d529ff017aa2758/configure.ac#L1644-L1645

kou:
Is the registered `GPG_UID` used? If it's not used, we can remove it.

abetomo:
I deleted it because we don't use it.

Check after deletion:

```
$ grep -ir gpg_uid *
packages/repository-helper.rb:      File.read(File.join(top_directory, "gpg_uid_rsa4096")).strip,
```

Issue #1411, opened 2022-09-12
 doc: Add description about `CMAKE_BUILD_PARALLEL_LEVEL` and parallel options


HashidaTKS:
We can use the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable for parallel build with ``cmake --build`: https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html

_Originally posted by @kou in https://github.com/groonga/groonga/pull/1403#discussion_r967935280_

HashidaTKS:
https://github.com/groonga/groonga/pull/1403#discussion_r968048123

PR #2529, opened 2025-09-10
docs: Use Ninja on Windows, when building with CMake

abetomo:
GitHub fixes GH-1411

Using Ninja enables parallel builds.

### DIFF
--- a/doc/source/install/cmake.md
+++ b/doc/source/install/cmake.md
@@ -294,3 +294,6 @@ Here is a command line to build and install Groonga by `cmake
 
 You should specify `--config Debug` instead of `--config Release` when
 debugging.
+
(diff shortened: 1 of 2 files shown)


abetomo:
Related: #2489

doc/source/install/cmake.md
```diff
@@ -294,3 +294,6 @@ Here is a command line to build and install Groonga by `cmake
 
 You should specify `--config Debug` instead of `--config Release` when
 debugging.
+
+We showed an example specifying `-G "Visual Studio 17 2022"` for the generator, but we recommend specifying Ninja for the generator, like `-G Ninja`.
+Ninja builds in parallel by default using all CPU cores, so you can expect fast builds.
```
kou, line 299: Note that we need to call `vcvarsall.bat` like https://github.com/groonga/groonga/blob/4086f5ac37d2601696a561896d529ff017aa2758/.github/workflows/cmake.yml#L795 to use Ninja + Visual Studio.

Issue #2530, opened 2025-09-12
Add `command_list` command


kou:
### What is your problem?

We may be able to generate client library API automatically by this introspection API.

This API will return a list of information of available commands. Each information includes:

* Name
* Summary
* Description
* Arguments
  * Argument name
  * Argument summary
  * Argument description

This API may be useful to implement Groonga MCP server.

PR #2532, opened 2025-09-12
command_list: add

kou:
GitHub: GH-2530

This returns only ID and name. We can add more information later.

### DIFF
--- a/lib/proc/proc_command_list.c
+++ b/lib/proc/proc_command_list.c
@@ -35,7 +35,7 @@ command_command_list(grn_ctx *ctx,
   {
     if (id == GRN_DB_MECAB) {
       /* TokenMecab must not be a command. TokenMecab may exist as a
-       * dummy object for historically reason. */
+       * dummy object for historical reason. */
       continue;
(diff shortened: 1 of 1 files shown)


lib/proc/proc_command_list.c
```diff
@@ -0,0 +1,113 @@
...
+{
+  const bool is_close_opened_object_mode = (grn_thread_get_limit() == 1);
+  grn_obj command_ids;
+
+  GRN_RECORD_INIT(&command_ids, GRN_OBJ_VECTOR, GRN_ID_NIL);
+  GRN_DB_EACH_BEGIN_BY_KEY(ctx, cursor, id)
+  {
+    if (id == GRN_DB_MECAB) {
+      /* TokenMecab must not be a command. TokenMecab may exist as a
+       * dummy object for historically reason. */
```
otegami: [nits]
```suggestion
       * dummy object for historical reason. */
```

kou: Thanks!

PR #2535, opened 2025-09-22
doc install amazonlinux: Remove the install procudere of groonga-token-mecab package

komainu8:
Because we don't provide the groonga-token-mecab package in Amazon Linux 2023.


kou:
Could you rebase this?

komainu8:
I've rebased.

PR #2547, opened 2025-09-30
Use C++20 or later

kou:
We want to use `std::string::starts_with()`.


kou:
Hmm. llama.cpp doesn't like C++20 for now.

Let's revisit this later.

PR #2548, opened 2025-10-01
language-model: add support for auto download from Hugging Face

kou:
URL:
* `hf:///${USER}/${MODEL}`
  * Example: `hf:///groonga/bge-m3-Q4_K_M-GGUF` for https://huggingface.co/groonga/bge-m3-Q4_K_M-GGUF
* `hf:///${USER}/${MODEL}#${TAG}`
  * Example: `hf:///Qwen/Qwen3-Embedding-8B-GGUF#Q8_0` for Q8_0 tag of https://huggingface.co/Qwen/Qwen3-Embedding-8B-GGUF

### DIFF
--- a/lib/language_model.cpp
+++ b/lib/language_model.cpp
@@ -747,7 +747,7 @@ namespace grn {
 
       return ensure_model(model_file_name);
 #else
-      ERR(GRN_FUNCTION_NOT_IMPLEMENTED, "%s simdjson isn't enabled", TAG)
+      ERR(GRN_FUNCTION_NOT_IMPLEMENTED, "%s simdjson isn't enabled", TAG);
       return false;
 #endif
     }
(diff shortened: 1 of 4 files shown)


lib/language_model.cpp
```diff
@@ -14,17 +14,27 @@
...
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
 
-#include "grn_language_model.hpp"
 #include "grn_db.h"
+#include "grn_http_client.h"
+#include "grn_language_model.hpp"
+#include "grn_util.h"
 
-#include "groonga/smart_obj.hpp"
+#include <groonga/smart_obj.hpp>
```
otegami, line 22: 📝 When we want to include the system's header files, we can use `<>`.
- https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html#Include-Syntax-1

PR #2559, opened 2025-10-13
compressor: add

kou:
It abstracts compression routines for zlib, LZ4, Zstandard and Blosc2.

grn_ja uses this instead of raw zlib, LZ4, Zstandard and Blosc2 API.

grn_ja_reader is also changed but it may not work. We will revisit this when we use grn_ja_reader. There is no grn_ja_reader user for now. We need streaming API and cache API when we revisit this.

This is for adding new compression routine such as OpenZL and adding support for table key compression.

### DIFF
--- a/lib/compressor.c
+++ b/lib/compressor.c
@@ -51,7 +51,7 @@
  *
  * Metadata:
  * * Upper 32bit is flags
- * * Lower 32bit is data size in byte of the original data
+ * * Lower 32bit is data length in byte of the original data
  *
  * Available flags:
  *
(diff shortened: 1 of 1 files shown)


lib/store.c
```diff
@@ -3620,147 +3620,6 @@ grn_ja_element_info(grn_ctx *ctx,
...
 }
 
-#define COMPRESSED_VALUE_META_FLAG(meta) ((meta) & 0xf000000000000000)
-#define COMPRESSED_VALUE_META_FLAG_RAW   0x1000000000000000
-#define COMPRESSED_VALUE_META_UNCOMPRESSED_LEN(meta)                           \
-  ((meta) & 0x0fffffffffffffff)
-
-#define COMPRESS_THRESHOLD_BYTE        256
-#define COMPRESS_PACKED_VALUE_SIZE_MAX 257
-/* COMPRESS_THRESHOLD_BYTE - 1 + sizeof(uint64_t) = 257 */
```
kou, line 3630: This calculation was wrong... This must be 263 not 257...

This was used for allocating a memory on stack. Existing code had a buffer overflow problem...

lib/store.c
```diff
@@ -3620,147 +3620,6 @@ grn_ja_element_info(grn_ctx *ctx,
...
-                  grn_ja *ja,
-                  grn_id id,
-                  void *value,
-                  uint32_t value_len,
-                  int flags,
-                  uint64_t *cas)
-{
-  char *packed_value[COMPRESS_PACKED_VALUE_SIZE_MAX];
-  uint32_t packed_value_len;
-  uint64_t packed_value_meta;
```
kou, line 3676: `COMPRESS_PACKED_VALUE_SIZE_MAX` doesn't have enough size. It's 6 bytes short.

`packed_value_len` and `packed_value_meta` may be overwritten by the following `gem_memcpy()`. `packed_value_meta` isn't used after the `grn_memcpy()` but `grn_value_len` is used after the `grn_memcpy()`. `grn_value_len` may be broken...?

PR #2561, opened 2025-10-15
cmake: enable -Wtype-limits warning and fix comparison

otegami:
## Issue

```
/home/buildbot/extra/groonga/lib/raw_string.c:132:24: error: comparison of unsigned expression in '< 0' is always false [-Werror=type-limits]
```

### Hot to reproduce

We can reproduce it by specifying `-Werror=type-limits`.
```console
cmake \
  -S . \
  -B ../groonga.build \
  --preset=debug-default \
  -DCMAKE_INSTALL_PREFIX=/tmp/local \
  -DGRN_WITH_APACHE_ARROW=ON \
  -DCMAKE_C_FLAGS="-Werror=type-limits" && \
cmake --build ../groonga.build/
```

## Causes

In MariaDB’s build environment, compiler warnings are promoted to errors.
As a result, any `-Wtype-limits` diagnostics fail the build.

## Solution

This PR fixes the existing violation in
`grn_raw_string_substring`() where `size_t` was
compared with < 0, which is always false since
`size_t` is unsigned.

And also, it enables the `-Wtype-limits` compiler
warning to catch comparisons that are always true or false due to the type's range.

### DIFF
--- a/lib/operator.c
+++ b/lib/operator.c
@@ -202,10 +202,21 @@ static const char *operator_script_syntaxes[] = {
 
 #define GRN_OP_LAST GRN_OP_ORDERED_NEAR_PHRASE_PRODUCT
 
+static inline bool
+grn_operator_is_valid(grn_operator op)
+{
+  /* We can't assume grn_operator is signed or unsigned. If
+   * grn_operator is unsigned, "op >= 0" may cause a warning. But "op
+   * >= 0" is needed when grn_operator is signed. We can suppress the
+   * warning with unsigned by "(int)op >= 0" and keep the check for
+   * signed. */
+  return (int)op >= 0 && op <= GRN_OP_LAST;
+}
+
 const char *
 grn_operator_to_string(grn_operator op)
 {
-  if (op <= GRN_OP_LAST) {
+  if (grn_operator_is_valid(op)) {
     return operator_names[op];
   } else {
     return "unknown";
@@ -215,7 +226,7 @@ grn_operator_to_string(grn_operator op)
 const char *
 grn_operator_to_script_syntax(grn_operator op)
 {
-  if (op <= GRN_OP_LAST) {
+  if (grn_operator_is_valid(op)) {
     return operator_script_syntaxes[op];
   } else {
     return "unknown";


abetomo (approved):

otegami:
There are build errors with mruby. I will check them now.

vendor/mruby/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ if(GRN_WITH_MRUBY)
     string(APPEND MRUBY_C_COMPILE_FLAGS " -Wno-float-equal")
     string(APPEND MRUBY_C_COMPILE_FLAGS " -Wno-missing-field-initializers")
     string(APPEND MRUBY_C_COMPILE_FLAGS " -Wno-sign-compare")
+    string(APPEND MRUBY_C_COMPILE_FLAGS " -Wno-type-limits")
```
otegami, line 121: I'm thinking I will fix the following warning later in the upstream as follows.
```
/home/runner/work/groonga/groonga/vendor/mruby-source/mrbgems/mruby-io/src/io.c:1019:46: error: result of comparison 'off_t' (aka 'long') > 9223372036854775807 is always false [-Werror,-Wtautological-type-limit-compare]
 1019 |   if (sizeof(off_t) > sizeof(mrb_int) && pos > (off_t)MRB_INT_MAX) {
      |                                          ~~~ ^ ~~~~~~~~~~~~~~~~~~
1 error generated.
```
ref: https://github.com/mruby/mruby/commit/83ecb2f078234bca084b934bb086e9f13cb967f6

lib/mrb/mrb_bulk.c
```diff
@@ -192,11 +192,15 @@ grn_mrb_value_from_bulk(mrb_state *mrb, grn_obj *bulk)
...
-      if ((uint64_t)value <= (uint64_t)MRB_INT_MAX) {
+#  if UINT32_MAX > MRB_INT_MAX
+      if (value <= MRB_INT_MAX) {
         mrb_value_ = mrb_int_value(mrb, value);
       } else {
         mrb_value_ = mrb_float_value(mrb, value);
       }
+#  else
+      mrb_value_ = mrb_int_value(mrb, value);
+#  endif
```
kou: Why do we need this?

FYI: We always define `MRB_INT64`:

https://github.com/groonga/groonga/blob/bc604c9bc5cca5170e1d7776845207f361f00e5b/vendor/mruby/CMakeLists.txt#L101

lib/operator.c
```diff
@@ -205,7 +205,7 @@ static const char *operator_script_syntaxes[] = {
 const char *
 grn_operator_to_string(grn_operator op)
 {
-  if (op >= 0 && op <= GRN_OP_LAST) {
+  if (op <= GRN_OP_LAST) {
```
kou: Can we assume that `grn_operator` is always unsigned?

lib/mrb/mrb_bulk.c
otegami: Thanks, I thought MRB_INT_MAX would be INT32_MAX when `MRB_INT64` was not defined.
So we don't need this condition anymore, right?

lib/operator.c
otegami: I think no. The underlying type of an enum is implementation-defined. Some compilers may choose an unsigned underlying type, while others may use a signed type.

I removed `(op >= 0` here because of the following reasons. If there is better way, could you share it with me? 🙏🏾 
For our use case, I think all enumerators are non-negative and `op` only comes from our parser/dispatcher, so the lower-bound check 0 <= op is redundant. Keeping only the upper bound avoids Clang’s `-Wtautological-unsigned-enum-zero-compare` and preserves validation on both signed and unsigned implementations.

```console
$ cat test_enum_signedness.c
#include <stdio.h>

/* Simplified grn_operator enum */
typedef enum {
  GRN_OP_PUSH = 0,
  GRN_OP_POP,
  GRN_OP_NOP,
  GRN_OP_EQUAL = 30,
  GRN_OP_NOT_EQUAL,
  GRN_OP_LESS,
  GRN_OP_LAST = 100
} grn_operator;

const char*
check_operator_old(grn_operator op)
{
  if (op >= 0 && op <= GRN_OP_LAST) {
    return "valid (old check)";
  } else {
    return "invalid (old check)";
  }
}

const char*
check_operator_new(grn_operator op)
{
  if (op <= GRN_OP_LAST) {
    return "valid (new check)";
  } else {
    return "invalid (new check)";
  }
}

int main(void)
{
  printf("Testing grn_operator enum signedness\n");
  printf("=====================================\n\n");

  printf("Enum value tests:\n");
  printf("  GRN_OP_PUSH = %d\n", GRN_OP_PUSH);
  printf("  GRN_OP_EQUAL = %d\n", GRN_OP_EQUAL);
  printf("  GRN_OP_LAST = %d\n\n", GRN_OP_LAST);

  printf("Size of enum: %zu bytes\n\n", sizeof(grn_operator));

  /* Test with valid values */
  grn_operator op1 = GRN_OP_EQUAL;
  printf("Testing op = GRN_OP_EQUAL (30):\n");
  printf("  Old check: %s\n", check_operator_old(op1));
  printf("  New check: %s\n\n", check_operator_new(op1));

  /* Test with invalid value */
  grn_operator op2 = (grn_operator)200;
  printf("Testing op = 200 (invalid):\n");
  printf("  Old check: %s\n", check_operator_old(op2));
  printf("  New check: %s\n\n", check_operator_new(op2));

  /* Test if -1 cast to enum */
  grn_operator op3 = (grn_operator)(-1);
  printf("Testing op = -1 (cast from negative):\n");
  printf("  Value as int: %d\n", (int)op3);
  printf("  Value as unsigned: %u\n", (unsigned)op3);
  printf("  Old check: %s\n", check_operator_old(op3));
  printf("  New check: %s\n\n", check_operator_new(op3));

  return 0;
}
$ clang -Wtype-limits test_enum_signedness.c -o test_enum_signedness
test_enum_signedness.c:17:10: warning: result of comparison of unsigned enum expression >= 0 is always true [-Wtautological-unsigned-enum-zero-compare]
   17 |   if (op >= 0 && op <= GRN_OP_LAST) {
      |       ~~ ^  ~
1 warning generated.
$ gcc -Wtype-limits test_enum_signedness.c -o test_enum_signedness
$ ./test_enum_signedness 
Testing grn_operator enum signedness
=====================================

Enum value tests:
  GRN_OP_PUSH = 0
  GRN_OP_EQUAL = 30
  GRN_OP_LAST = 100

Size of enum: 4 bytes

Testing op = GRN_OP_EQUAL (30):
  Old check: valid (old check)
  New check: valid (new check)

Testing op = 200 (invalid):
  Old check: invalid (old check)
  New check: invalid (new check)

Testing op = -1 (cast from negative):
  Value as int: -1
  Value as unsigned: 4294967295
  Old check: invalid (old check)
  New check: invalid (new check)
```

lib/mrb/mrb_bulk.c
otegami: fix: [4412d4d](https://github.com/groonga/groonga/pull/2561/commits/4412d4d205bbdd79b719999d0be52e876e7199ba)

lib/operator.c
kou: It doesn't work if the `enum` is signed:

```diff
--- test_enum_signedness.c.orig	2025-10-16 10:32:57.388973395 +0900
+++ test_enum_signedness.c	2025-10-16 10:34:17.976780507 +0900
@@ -1,7 +1,7 @@
 #include <stdio.h>
 
 /* Simplified grn_operator enum */
-typedef enum {
+typedef enum : int {
   GRN_OP_PUSH = 0,
   GRN_OP_POP,
   GRN_OP_NOP,
```

```console
$ ./test_enum_signedness
Testing grn_operator enum signedness
=====================================

Enum value tests:
  GRN_OP_PUSH = 0
  GRN_OP_EQUAL = 30
  GRN_OP_LAST = 100

Size of enum: 4 bytes

Testing op = GRN_OP_EQUAL (30):
  Old check: valid (old check)
  New check: valid (new check)

Testing op = 200 (invalid):
  Old check: invalid (old check)
  New check: invalid (new check)

Testing op = -1 (cast from negative):
  Value as int: -1
  Value as unsigned: 4294967295
  Old check: invalid (old check)
  New check: valid (new check)
```

kou: > `op` only comes from our parser/dispatcher

This is a public API. So users can also use this.

> so the lower-bound check `0 <= op` is redundant

If we define that caller is responsible for passing a valid value, we can also remove upper-bound check.

kou: Can we use something like this?

```diff
diff --git a/lib/operator.c b/lib/operator.c
index 60dea84ed..562a2b369 100644
--- a/lib/operator.c
+++ b/lib/operator.c
@@ -202,10 +202,21 @@ static const char *operator_script_syntaxes[] = {
 
 #define GRN_OP_LAST GRN_OP_ORDERED_NEAR_PHRASE_PRODUCT
 
+static inline bool
+grn_operator_is_valid(grn_operator op)
+{
+  /* We can't assume grn_operator is signed or unsigned. If
+   * grn_operator is unsigned, "op >= 0" may cause a warning. But "op
+   * >= 0" is needed when grn_operator is signed. We can suppress the
+   * warning with unsigned by "(int)op >= 0" and keep the check for
+   * signed. */
+  return (int)op >= 0 && op <= GRN_OP_LAST;
+}
+
 const char *
 grn_operator_to_string(grn_operator op)
 {
-  if (op >= 0 && op <= GRN_OP_LAST) {
+  if (grn_operator_is_valid(op)) {
     return operator_names[op];
   } else {
     return "unknown";
@@ -215,7 +226,7 @@ grn_operator_to_string(grn_operator op)
 const char *
 grn_operator_to_script_syntax(grn_operator op)
 {
-  if (op >= 0 && op <= GRN_OP_LAST) {
+  if (grn_operator_is_valid(op)) {
     return operator_script_syntaxes[op];
   } else {
     return "unknown";
```

otegami: fix: [17b3fca](https://github.com/groonga/groonga/pull/2561/commits/17b3fca54361e26ca9653b207083635d074f52f2) Thank you so much. I understood it. I make sure it works too using the example codes.

```
$ ./test_enum_signedness
Testing grn_operator enum signedness
=====================================

Enum value tests:
  GRN_OP_PUSH = 0
  GRN_OP_EQUAL = 30
  GRN_OP_LAST = 100

Size of enum: 4 bytes

Testing op = GRN_OP_EQUAL (30):
  Old check: valid (old check)
  New check: valid (new check)

Testing op = 200 (invalid):
  Old check: invalid (old check)
  New check: invalid (new check)

Testing op = -1 (cast from negative):
  Value as int: -1
  Value as unsigned: 4294967295
  Old check: invalid (old check)
  New check: invalid (new check)
```

otegami:
Thank you for reviews. I've addressed all of review comments so far.

PR #2568, opened 2025-10-17
io: use grn_io_mapinfo for grn_mmap()/grn_unmap()

kou:
We don't want to use `HANDLE` as `grn_mmap()`/`grn_munmap()` argument.


kou:
I abandon this for now...

I wanted to generalize the mmap abstraction in `io.c` and reuse it in other code. I'll create a new mmap abstraction.

PR #2588, opened 2025-10-21
Add support for RaBitQ based ANN search

kou:
This is an experimental feature. API and DB schema may be changed.

This is implemented as `language_model/knn` plugin but its name may be changed.

This requires Faiss 1.12.0 or later. Bundled Faiss isn't supported yet.

OR works but AND doesn't work.

We need to improve performance.

We'll add custom sorter later.


kou:
We need to add support for bundled Faiss for CI.

kou:
The `language_model/knn` plugin doesn't work on GitHub Actions but I'll merge this.
Let's fix it later.

PR #2613, opened 2025-10-29
grn_path_copy: Fix `system call error: File exists`

abetomo:
An error occurs when saving the model cache in language_model.

```
|e| system call error: File exists: [path][copy] failed to open destination path: <path/src> -> <path/dest>
```

If the file already exists, do nothing and handle it as a success.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -1779,14 +1779,20 @@ grn_path_copy(grn_ctx *ctx,
   }
 
   int destination_fd;
+  errno = 0;
   grn_open(destination_fd,
            destination_path,
            O_WRONLY | O_CREAT | O_EXCL | GRN_OPEN_FLAG_BINARY);
   if (destination_fd == -1) {
-    SERR("%s failed to open destination path: <%s> -> <%s>",
-         tag,
-         source_path,
-         destination_path);
+    if (errno == EEXIST) {
+      /* If the file already exists, it will succeed without overwriting it. */
+      ctx->rc = GRN_SUCCESS;
+    } else {
+      SERR("%s failed to open destination path: <%s> -> <%s>",
+           tag,
+           source_path,
+           destination_path);
+    }
     grn_close(source_fd);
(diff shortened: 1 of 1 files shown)


lib/util.c
```diff
@@ -1779,14 +1779,20 @@ grn_path_copy(grn_ctx *ctx,
   }
 
   int destination_fd;
+  errno = 0;
   grn_open(destination_fd,
            destination_path,
            O_WRONLY | O_CREAT | O_EXCL | GRN_OPEN_FLAG_BINARY);
```
abetomo, line 1785:
```suggestion
           O_WRONLY | O_CREAT | O_TRUNC | GRN_OPEN_FLAG_BINARY);
```

Is overwriting better?

kou: If we want to overwrite, we just need to remove `O_EXCL`. We don't need `O_TRUNC`.

It may be better that `cp ${source_path} ${destination_path}.${pid} && mv ${destination_path}.${pid} ${destination_path}` instead of `flock()`...

lib/util.c
```diff
@@ -1779,14 +1779,20 @@ grn_path_copy(grn_ctx *ctx,
   }
 
   int destination_fd;
+  errno = 0;
```
komainu8, line 1782: Is this need?
`open` or `_sopen_s` is set errno appropriately when `destination_fd == -1` is met.
Therefore, we don't need to initialize errno. 

Ref:
`open()`:
 https://linuxjm.sourceforge.io/html/LDP_man-pages/man2/open.2.html
「open(), openat(), and creat() return the new file descriptor (a nonnegative integer), or -1 if an error occurred (in which case, errno is set appropriately). 」

`_sopen_s()`
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sopen-s-wsopen-s?view=msvc-170
「A nonzero return value indicates an error; in that case errno is set to one of the following values.」

Issue #2616, opened 2025-10-30
ci: The "Docker (arch-linux)" job is failing


kou:
### What happend?

https://github.com/groonga/groonga/actions/runs/18901396279/job/53949276925?pr=2614#step:9:3351

```text
  :: Retrieving packages...
  error: failed retrieving file 'aws-sdk-cpp-s3-1.11.667-1-x86_64.pkg.tar.zst' from geo.mirror.pkgbuild.com : The requested URL returned error: 404
  error: failed retrieving file 'aws-sdk-cpp-iam-1.11.667-1-x86_64.pkg.tar.zst' from geo.mirror.pkgbuild.com : The requested URL returned error: 404
  error: failed retrieving file 'aws-sdk-cpp-core-1.11.667-1-x86_64.pkg.tar.zst' from geo.mirror.pkgbuild.com : The requested URL returned error: 404
  warning: too many errors from geo.mirror.pkgbuild.com, skipping for the remainder of this transaction
  error: failed retrieving file 'aws-sdk-cpp-core-1.11.667-1-x86_64.pkg.tar.zst' from mirror.rackspace.com : The requested URL returned error: 404
  error: failed retrieving file 'aws-sdk-cpp-iam-1.11.667-1-x86_64.pkg.tar.zst' from mirror.rackspace.com : The requested URL returned error: 404
  error: failed retrieving file 'aws-sdk-cpp-s3-1.11.667-1-x86_64.pkg.tar.zst' from mirror.rackspace.com : The requested URL returned error: 404
  warning: too many errors from mirror.rackspace.com, skipping for the remainder of this transaction
  error: failed retrieving file 'aws-sdk-cpp-s3-1.11.667-1-x86_64.pkg.tar.zst' from mirror.leaseweb.net : The requested URL returned error: 404
  error: failed retrieving file 'aws-sdk-cpp-core-1.11.667-1-x86_64.pkg.tar.zst' from mirror.leaseweb.net : The requested URL returned error: 404
  error: failed retrieving file 'aws-sdk-cpp-iam-1.11.667-1-x86_64.pkg.tar.zst' from mirror.leaseweb.net : The requested URL returned error: 404
  warning: too many errors from mirror.leaseweb.net, skipping for the remainder of this transaction
  warning: failed to retrieve some files
   arrow-21.0.0-7-x86_64 downloading...
  error: failed to commit transaction (failed to retrieve some files)
   grpc-1.75.1-2-x86_64 downloading...
[...]

PR #2618, opened 2025-10-30
ci: archlinux: passing two refresh flag to force a refresh of all package databases

komainu8:
Fix: GH-2616

### Cause

Versions of some packages such as ruby-3.4.7-1-x86_64.pkg.tar.zst and aws-sdk-cpp-s3-1.11.667-1-x86_64.pkg.tar.zst are old.
These are not exist in mirror.
Therefore, we can't download require some packages currently.

### Solution

We update repository data forcefully.
Arch Linux's CI is successful as below by this modification.

https://github.com/groonga/groonga/actions/runs/18928862551/job/54041595746

### DIFF
--- a/dockerfiles/arch-linux.dockerfile
+++ b/dockerfiles/arch-linux.dockerfile
@@ -17,7 +17,7 @@
 FROM archlinux
 
 RUN \
-  pacman --sync --refresh --refresh --upgrade --noconfirm \
+  pacman --sync --refresh --refresh --sysupgrade --noconfirm \
     # mecab-ipadic must have this but it doesn't have it.
     # So we install this in base environment.
     autoconf \


dockerfiles/arch-linux.dockerfile
```diff
@@ -17,8 +17,7 @@
 FROM archlinux
 
 RUN \
-  pacman --sync --noconfirm --refresh --sysupgrade && \
-  pacman --sync --noconfirm \
+  pacman -Syyu --noconfirm \
```
kou: Can we use long options for readability?

komainu8: I see.

komainu8: I've fixed it.

komainu8:
Arch Linux's CI is successful in the latest modification.
https://github.com/komainu8/groonga/actions/runs/18929903979/job/54044659651

PR #2636, opened 2025-11-06
packages ubuntu: add missing require packages for enabling Faiss

komainu8:
Because if Faiss is disable, the installing of groonga-plugin-language-model is failure as below.

```
dh_install: warning: Cannot find (any matches for) "usr/lib/*/groonga/plugins/language_model/" (tried in ., debian/tmp)

dh_install: warning: groonga-plugin-language-model missing files: usr/lib/*/groonga/plugins/language_model/
dh_install: error: missing files, aborting
make: *** [debian/rules:15: binary] Error 255
```

### DIFF
--- a/packages/Rakefile
+++ b/packages/Rakefile
@@ -60,7 +60,7 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
     if os.start_with?("ubuntu-")
       options.concat( ["--env", "DEB_EMULATE_LAUNCHPAD_ENVIRONMENT=yes"])
     end
-    if os == "ubuntu-22.04"
+    if os == "ubuntu-jammy"
       options.concat(["--env", "CAN_BUILD_LIBFAISS=no"])
     end
     options


otegami:
The same error is raised on Ubuntu Jammy arm64.
```
dh_install: warning: Cannot find (any matches for) "usr/lib/*/groonga/plugins/language_model/" (tried in ., debian/tmp)

dh_install: warning: groonga-plugin-language-model missing files: usr/lib/*/groonga/plugins/language_model/
dh_install: error: missing files, aborting
make: *** [debian/rules:15: binary] Error 25
```
https://github.com/groonga/groonga/actions/runs/19127638103/job/54661474073?pr=2636#step:10:6256

packages/debian/rules
```diff
@@ -18,7 +18,9 @@ override_dh_auto_configure:
 	if [ "$${DEB_EMULATE_LAUNCHPAD_ENVIRONMENT}" = "yes" ]; then	\
 	  echo > /etc/resolv.conf;					\
 	fi
-	if dpkg -l libarrow-dev > /dev/null 2>&1; then		\
+	if [ "$${CAN_BUILD_LIBFAISS}" = "no" ]; then			\
+	  GRN_WITH_FAISS=OFF;						\
+	fi	if dpkg -l libarrow-dev > /dev/null 2>&1; then		\
```
otegami:
```suggestion
	fi
	if dpkg -l libarrow-dev > /dev/null 2>&1; then		\
```

packages/Rakefile
```diff
@@ -63,6 +63,11 @@ class GroongaPackageTask < PackagesGroongaOrgPackageTask
     else
       []
     end
+    if os == "ubuntu-22.04"
+      [
+        "--env", "CAN_BUILD_LIBFAISS=no",
+      ]
+    end
```
otegami, line 65: It might not work but I think we should return array at the end.
The current implementation returns `nil` at the end because we don't specify anything except for `os == "ubuntu-22.04"`'s case.
```suggestion
options = []
if os.start_with?("ubuntu-")
  options.concat( ["--env", "DEB_EMULATE_LAUNCHPAD_ENVIRONMENT=yes"])
end
if os == "ubuntu-22.04"
  options.concat(["--env", "CAN_BUILD_LIBFAISS=no"])
end
options
```
https://github.com/groonga/groonga/actions/runs/19161989852/job/54774847292?pr=2636#step:10:47

kou:
#2641 fixed this.

PR #2640, opened 2025-11-07
language_model_knn: support `k` option

abetomo:
Specify the number of records to return using `k` option.

### DIFF
--- a/test/command/suite/language_model/options/empty.test
+++ b/test/command/suite/language_model/options/empty.test
@@ -17,8 +17,11 @@ load --table Data
 ]
 
 table_create RaBitQ TABLE_HASH_KEY ShortBinary \
-  --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
-                                             "code_column", "rabitq_code")'
+  --default_tokenizer \
+    'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
+                           "code_column", "rabitq_code", \
+                           "passage_prefix", "passage: ", \
+                           "query_prefix", "query: ")'
 # This is for waiting for downloading model.
 #@timeout 300
 column_create RaBitQ data_text COLUMN_INDEX Data text
--- a/test/command/suite/language_model/options/invalid/unknown_option.test
+++ b/test/command/suite/language_model/options/invalid/unknown_option.test
@@ -17,8 +17,11 @@ load --table Data
 ]
 
 table_create RaBitQ TABLE_HASH_KEY ShortBinary \
-  --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
-                                             "code_column", "rabitq_code")'
+  --default_tokenizer \
+    'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
+                           "code_column", "rabitq_code", \
+                           "passage_prefix", "passage: ", \
+                           "query_prefix", "query: ")'
 # This is for waiting for downloading model.
 #@timeout 300
 column_create RaBitQ data_text COLUMN_INDEX Data text
--- a/test/command/suite/language_model/options/k/minus/one.test
+++ b/test/command/suite/language_model/options/k/minus/one.test
@@ -25,8 +25,11 @@ load --table Data
 ]
 
 table_create RaBitQ TABLE_HASH_KEY ShortBinary \
-  --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
-                                             "code_column", "rabitq_code")'
+  --default_tokenizer \
+    'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
+                           "code_column", "rabitq_code", \
+                           "passage_prefix", "passage: ", \
+                           "query_prefix", "query: ")'
 # This is for waiting for downloading model.
 #@timeout 300
 column_create RaBitQ data_text COLUMN_INDEX Data text
--- a/test/command/suite/language_model/options/k/zero.expected
+++ b/test/command/suite/language_model/options/k/zero.expected
@@ -13,7 +13,7 @@ load --table Data
 {"text": "Groonga is a full text search engine."}
 ]
 [[0,0.0,0.0],3]
-table_create RaBitQ TABLE_HASH_KEY ShortBinary   --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF",                                              "code_column", "rabitq_code")'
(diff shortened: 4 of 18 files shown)


test/command/suite/language_model/options/k/zero.expected
```diff
@@ -0,0 +1,22 @@
...
+{"text": "This is an apple."},
+{"text": "Groonga is a full text search engine."}
+]
+[[0,0.0,0.0],3]
+table_create RaBitQ TABLE_HASH_KEY ShortBinary   --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/all-MiniLM-L6-v2-Q4_K_M-GGUF",                                              "code_column", "rabitq_code")'
+[[0,0.0,0.0],true]
+column_create RaBitQ data_text COLUMN_INDEX Data text
+[[0,0.0,0.0],true]
+#|w| load: special_eos_id is not in special_eog_ids - the tokenizer config may be incorrect
+select Data   --filter 'language_model_knn(text, "male child", { "k" : 0 })'   --output_columns text
```
otegami, line 22: What does "over" mean here?

abetomo: test/command/suite/language_model/options/k/over.test

Is this the one?
It meant a value over the number of records.

otegami: I'm so sorry I added the comment to the wrong file.
Thank you so much. I understood it.

otegami: How about specifying the `4` as a border value ?

otegami (approved):

test/command/suite/language_model/options/empty.test
```diff
@@ -0,0 +1,28 @@
...
+
+load --table Data
+[
+{"text": "I am a boy."},
+{"text": "This is an apple."},
+{"text": "Groonga is a full text search engine."}
+]
+
+table_create RaBitQ TABLE_HASH_KEY ShortBinary \
+  --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/all-MiniLM-L6-v2-Q4_K_M-GGUF", \
```
kou: Could you use `groonga/multilingual-e5-base-Q4_K_M-GGUF` for new tests?

kou: https://github.com/groonga/groonga/pull/2634

test/command/suite/language_model/options/k/minus/one.test
```diff
@@ -0,0 +1,28 @@
...
+
+table_create RaBitQ TABLE_HASH_KEY ShortBinary \
+  --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/all-MiniLM-L6-v2-Q4_K_M-GGUF", \
+                                             "code_column", "rabitq_code")'
+# This is for waiting for downloading model.
+#@timeout 300
+column_create RaBitQ data_text COLUMN_INDEX Data text
+
+select Data \
+  --filter 'language_model_knn(text, "male child", { "k" : -1 })' \
```
kou, line 38: `-1` is ignored, right?

abetomo: What behavior does 'ignored' refer to? 
Does it mean the default value of `k`, which is 10?
Or does it mean fetching all records like `--limit`?

In the current implementation, specifying `-1` fetches all records.

kou: > Does it mean the default value of `k`, which is 10?

This. Because `GRN_PROC_OPTION_VALUE_UINT32` is used.

> In the current implementation, specifying `-1` fetches all records.

If we test this, we need 11 or more records.

abetomo: With the current implementation, if there are 11 records, specifying `-1` sets `k` to the total number of records, which is 11.
`select` result will only display the default `limit` value of 10 records.

Since that doesn't seem to be as expected, I will change it so that when `k = -1`, it becomes `k = 10`.

plugins/language_model/knn.cpp
```diff
@@ -1072,9 +1073,25 @@ namespace {
...
+                                         GRN_PROC_OPTION_VALUE_UINT32,
+                                         &k,
+                                         NULL);
+      if (rc != GRN_SUCCESS) {
+        return ctx->rc;
+      }
+      if (k == 0) {
+        return GRN_SUCCESS;
+      }
+      if (k >= UINT32_MAX) {
```
kou: Hmm. Should we use `GRN_PROC_OPTION_VALUE_INT32` not `UINT32`? Or should we check overflow in `grn_proc_get_value_uint32()`?

test/command/suite/language_model/options/k/minus/one.test
kou: Hmm. How about aligning with `--limit` behavior?

`-1` == the number of records in the target table and `-2` == the number of records in the target table - 1.

plugins/language_model/knn.cpp
```diff
@@ -1072,9 +1072,29 @@ namespace {
...
+                                         NULL);
+      if (rc != GRN_SUCCESS) {
+        return ctx->rc;
+      }
+      if (arg_k == 0) {
+        return GRN_SUCCESS;
+      } else if (arg_k > 0) {
+        k = static_cast<uint32_t>(arg_k);
+      } else {
+        k = grn_table_size(ctx, table);
```
kou: Hmm. If we have 100 records in `table` and `"k": -4`, are 100 records not 97 records processed?

plugins/language_model/knn.cpp
```diff
@@ -1072,9 +1072,29 @@ namespace {
...
+        if (k + arg_k <= 0) {
```
kou: Does this work with `grn_table_size() == 1` + `"k": -1` case?

plugins/language_model/knn.cpp
```diff
@@ -1106,7 +1126,12 @@ namespace {
     }
 
     auto candidates = *maybe_candidates;
-    k = std::min(k, static_cast<uint32_t>(candidates.size()));
+    if (arg_k > 0) {
+      k = std::min(k, static_cast<uint32_t>(candidates.size()));
+    } else {
+      k = static_cast<uint32_t>(
+        std::max(0, static_cast<int32_t>(candidates.size() + arg_k + 1)));
+    }
```
kou: Hmm. Should we use `arg_k` here too...?

abetomo: Thanks for your comment. It really helped me organize things.

test/command/suite/language_model/options/invalid/unknown_option.test
```diff
@@ -0,0 +1,28 @@
...
+load --table Data
+[
+{"text": "I am a boy."},
+{"text": "This is an apple."},
+{"text": "Groonga is a full text search engine."}
+]
+
+table_create RaBitQ TABLE_HASH_KEY ShortBinary \
+  --default_tokenizer 'TokenLanguageModelKNN("model", "hf:///groonga/multilingual-e5-base-Q4_K_M-GGUF", \
+                                             "code_column", "rabitq_code")'
```
kou: Could you also specify `"passage_prefix", "passage: ", "query_prefix", "query: "`? They are needed for multilingual-e5-base.

PR #2645, opened 2025-11-11
ci cmake: Add MSYS2 ARM64 job

otegami:
GitHub: GH-2646

Note: This is still work in progress. CI is currently failing, but we will fix these errors step by step.


kou:
We should merge this before #2647, right?

otegami:
Yes. I want to do it before #2647.

Issue #2646, opened 2025-11-11
ci windows: support Windows ARM64


otegami:
### What is your problem?

Groonga fails to build on Windows ARM64 environment with MSYS2. This issue was discovered when attempting to update Groonga to version 15.2.0 in the MINGW-packages repository for ARM64 Windows.

The build failure was not detected earlier because Groonga's CI pipeline currently does not include ARM64 Windows environment testing. With the increasing adoption of ARM64 Windows devices, it's important to ensure Groonga builds and runs correctly on this platform.

```
  CMAKE_BUILD_TYPE=Release
  -- Found Git: C:/M/msys64/usr/bin/git.exe (found version "2.51.2")
  -- The ASM compiler identification is Clang with GNU-like command-line
  -- Found assembler: C:/M/msys64/clangarm64/bin/clang.exe
  -- Setting GGML_NATIVE_DEFAULT to OFF
  -- Warning: ccache not found - consider installing it for faster compilation or disable this warning with GGML_CCACHE=OFF
  -- CMAKE_SYSTEM_PROCESSOR: ARM64
  -- GGML_SYSTEM_ARCH: ARM
  -- Including CPU backend
  CMake Error at C:/_/B/src/build-CLANGARM64/_deps/llama_cpp-src/ggml/src/CMakeLists.txt:363 (message):
    Unsupported ARM target OS: Windows
  
  
  -- Configuring incomplete, errors occurred!
  ==> ERROR: A failure occurred in build().
      Aborting...
  ==> Removing installed dependencies...
  checking dependencies...
```

### How to reproduce it

Attempt to build Groonga 15.2.0 using the MINGW-packages build scripts on Windows ARM64 environemnt.

otegami:
The follwoing error is raised now. This error has already fixed at upstream. So we will update the submodule version of mruby.

```
C:/a/groonga/groonga/ci/msys2/src/groonga-15.2.1/vendor/mruby-source/src/error.c:201:3: error: __builtin_longjmp is not supported for the current target
  201 |   MRB_THROW(mrb->jmp);
```

ref: https://github.com/mruby/mruby/issues/6637

otegami:
When we upgrade the mruby, the following error is raised.
```
rake aborted!
Can't find /home/otegami/work/c/groonga/vendor/mruby-source/mrbgems/mruby-print/mrbgem.rake
[...]

PR #2647, opened 2025-11-11
cmake: disable GGML_CPU_ALL_VARIANTS on Windows ARM64

otegami:
GitHub: GH-2646

## Issue

When building with `GGML_CPU_ALL_VARIANTS=ON` on Windows ARM64, the build fails with:

```
CMake Error at ggml/src/CMakeLists.txt:367 (message):
  Unsupported ARM target OS: Windows
```

## Cause

llama.cpp's GGML_CPU_ALL_VARIANTS feature doesn't
support Windows ARM64 yet. It only supports ARM on Linux, Android, and macOS.
Even llama.cpp's own CI explicitly disables
`GGML_CPU_ALL_VARIANTS` for Windows ARM64 builds too.

ref: https://github.com/ggml-org/llama.cpp/blob/ece0f5c1771f1835e66900d4168233f0430d819d/ggml/src/CMakeLists.txt#L343-L368
ref: https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/release.yml#L288

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1743,7 +1743,12 @@ if(NOT "${GRN_WITH_LLAMA_CPP}" STREQUAL "no")
       string(SUBSTRING "${GRN_LLAMA_CPP_BUNDLED_VERSION}" 1 -1
                        GGML_BUILD_NUMBER)
       set(GGML_CPU_ALL_VARIANTS ON)
-      if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+      if(WIN32 AND GRN_SYSTEM_PROCESSOR STREQUAL "arm64")
+        # GGML_CPU_ALL_VARIANTS doesn't support Windows ARM64 yet
+        #
+        # https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/release.yml#L288
+        set(GGML_CPU_ALL_VARIANTS OFF)
+      elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         if(GRN_SYSTEM_PROCESSOR STREQUAL "x86_64")
           # GCC 11.5 or later is needed for "-mavxvnni" and so on
           if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.5")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1783,6 +1783,13 @@ if(NOT "${GRN_WITH_LLAMA_CPP}" STREQUAL "no")
         else()
           set(GGML_CPU_ALL_VARIANTS OFF)
         endif()
+      elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
kou: Is this check needed? We need to use `OFF` even when GCC + Windows + arm64, right?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e8b2c0c62..cb5f7568b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1743,7 +1743,12 @@ if(NOT "${GRN_WITH_LLAMA_CPP}" STREQUAL "no")
       string(SUBSTRING "${GRN_LLAMA_CPP_BUNDLED_VERSION}" 1 -1
                        GGML_BUILD_NUMBER)
       set(GGML_CPU_ALL_VARIANTS ON)
-      if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+      if(WIN32 AND GRN_SYSTEM_PROCESSOR STREQUAL "arm64")
+        # GGML_CPU_ALL_VARIANTS doesn't support Windows ARM64 yet
+        #
+        # https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/release.yml#L288
+        set(GGML_CPU_ALL_VARIANTS OFF)
+      elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         if(GRN_SYSTEM_PROCESSOR STREQUAL "x86_64")
           # GCC 11.5 or later is needed for "-mavxvnni" and so on
           if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.5")
```

otegami: fix: [00ba172](https://github.com/groonga/groonga/pull/2647/commits/00ba172dc00520339b2b0b591283d3c194bd064e) Yes, we don't need to check compiler here. I added it because I focus on the CLANGARM64 environment.

otegami:
Thanks for the reviews! I've addressed your comments so far.

otegami:
Done: ~~I will rebase it now.~~

PR #2653, opened 2025-11-25
windows: fix compilation errors on ARM64

otegami:
## Issue

Compilation fails on ARM64 Windows CI with the following error:

```
lib/windows.c:408:4: error: "Intel x86, Intel Itanium and x64 are only supported architectures"
  408 | #  error "Intel x86, Intel Itanium and x64 are only supported architectures"
      |    ^
1 error generated.
```

## Cause

The stack walking code in `lib/windows.c` only supports Intel x86, Intel Itanium, and x64 architectures. ARM64 architecture was not recognized.

## Solution

Add support for the `_M_ARM64` compiler macro and
configure the StackWalk64 API with ARM64-specific
registers. msys2 packages use the same approach:

- `Pc` (Program Counter)
- `Fp` (Frame Pointer)
- `Sp` (Stack Pointer)

ref: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-groonga/001-aarch64.patch
ref: https://developer.arm.com/documentation/102374/0103/Registers-in-AArch64---other-registers
ref: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-arm64_nt_context

Also update the error message to include ARM64 in the list of supported architectures.

### DIFF
--- a/lib/windows.c
+++ b/lib/windows.c
@@ -404,8 +404,13 @@ grn_windows_log_back_trace(grn_ctx *ctx,
   frame.AddrPC.Offset = context->Rip;
   frame.AddrFrame.Offset = context->Rbp;
   frame.AddrStack.Offset = context->Rsp;
+# elif defined(_M_ARM64)
+  machine_type = IMAGE_FILE_MACHINE_ARM64;
+  frame.AddrPC.Offset = context->Pc;
+  frame.AddrFrame.Offset = context->Fp;
+  frame.AddrStack.Offset = context->Sp;
 # else /* _M_IX86 */
-#  error "Intel x86, Intel Itanium and x64 are only supported architectures"
+#  error "Intel x86, Intel Itanium, x64 and ARM64 are only supported architectures"
 # endif /* _M_IX86 */
 
   DWORD64 previous_address = 0;


kou:
Can we remove https://github.com/groonga/groonga/blob/7b72335d0414e7120c96e1b91a12d0472b81ccc7/.github/workflows/cmake.yml#L624 and so on?

komainu8:
Could you confirm error( https://github.com/groonga/groonga/actions/runs/19659505671/job/56302929369?pr=2653 )?

otegami:
We couldn't install gem gobject-introspection because `mingw-w64-clang-aarch64-` prefix is required for the package name.
ref: https://packages.msys2.org/packages/mingw-w64-clang-aarch64-gobject-introspection
ref: https://github.com/ruby-gnome/native-package-installer/blob/main/lib/native-package-installer/platform/msys2.rb#L42-L54

```console
MAKEFLAGS=-j$(nproc) gem install \
    fiddle \
    grntest \
    pkg-config \
    red-arrow
  ruby -e "require 'arrow'"
ERROR:  Error installing red-arrow:
	ERROR: Failed to build gem native extension.

    current directory: C:/a/_temp/msys64/clangarm64/lib/ruby/gems/3.4.0/gems/gobject-introspection-4.3.3/ext/gobject-introspection
C:/a/_temp/msys64/clangarm64/bin/ruby.exe extconf.rb
...
checking for gobject-introspection-1.0... no (nonexistent)
installing 'gobject-introspection' native package... failed
Failed to run 'pacman -S --noconfirm gobject-introspection'.
```

kou:
We should improve native-package-installer.

otegami:
We will fix the following error in the following PR.
So I want this PR focus on this build error.
https://github.com/groonga/groonga/pull/2653#issuecomment-3574104758

kou:
You just confirmed that this is buildable and you haven't confirmed that this change generates backtrace properly on ARM64, right?

otegami:
> You just confirmed that this is buildable and you haven't confirmed that this change generates backtrace properly on ARM64, right?

Yes.

PR #2665, opened 2025-11-28
doc TokenLanguageModelKNN: Add centroid_column option

abetomo:


komainu8 (approved):

doc/source/reference/tokenizers/token_language_model_knn.md
```diff
@@ -187,6 +195,36 @@ Some models such as multilingual-e5 require prefix used for search-target texts
...
+#### `centroid_column`
+
+```{versionadded} 15.2.1
+
+```
+
+This option is for large embeddings with more than 1025 dimensions (more than 4100 bytes).
+Groonga table keys must be 4 KiB or smaller. Embeddings larger than 4 KiB cannot be stored as keys.
+
+You can use this option to store large embeddings in column values instead of table keys.
```
otegami: [nits]
```suggestion
You can use this option to store large embeddings in a column instead of in the table key.
```

PR #2672, opened 2025-12-04
Add parsed JSON format specification

kou:


tools/parsed-json.rb
```diff
@@ -0,0 +1,712 @@
...
+  def append_string(string)
+    offset = @string_offsets.bytesize - 4 # sizeof(uint32_t)
+    @string_values << string
+    last_string_offset = @string_offsets.unpack1("L", offset: offset)
+    @string_offsets << [last_string_offset + string.bytesize].pack("L")
+    offset
+  end
+
+  def append_int8(int8)
+    offset = @int8_values.bytesize
```
otegami: The instance variable `@int8_values` might be undefined in this class.

tools/parsed-json.rb
```diff
@@ -0,0 +1,712 @@
...
+
+  def test_double_positive
+    assert_roundtrip(2.9)
+  end
+
+  def test_double_nan
+    assert_roundtrip(Float::NAN)
+  end
+
+  def test_double_inifinity_positive
```
otegami:
```suggestion
  def test_double_infinity_positive
```

tools/parsed-json.rb
kou: Good catch! I forgot to remove finally unused codes.

tools/parsed-json.rb
kou: Good catch!

PR #2677, opened 2025-12-10
compressor: add skeleton support for OpenZL

otegami:
This adds the skeleton code for OpenZL support
compression support.

```console
$ groonga --version | grep openzl
Groonga 15.2.1-6-g5bff1f5 [Linux,x86_64,utf8,match-escalation-threshold=0,nfkc,mecab,message-pack,onigmo,zlib,lz4,zstandard,epoll,xxhash,h3,simdjson,llama.cpp,faiss,openzl]
```

We'll implement actual compression in the following PR.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -53,15 +53,15 @@ if(GRN_C_COMPILER_GNU_LIKE)
     SOURCE distance.cpp
     APPEND_STRING
     PROPERTY COMPILE_FLAGS " -Wno-unknown-pragmas -Wno-float-equal")
-  if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
-    # TODO: Remove this when OpenZL fixes duplicate typedef definitions
-    # in their headers.
-    # ref: https://github.com/facebook/openzl/pull/293
-    set_property(
-      SOURCE compressor.c
-      APPEND_STRING
-      PROPERTY COMPILE_FLAGS " -Wno-typedef-redefinition")
-  endif()
+endif()
+if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+  # TODO: Remove this when OpenZL fixes duplicate typedef definitions
+  # in their headers.
+  # ref: https://github.com/facebook/openzl/pull/293
+  set_property(
+    SOURCE compressor.c
+    APPEND_STRING
+    PROPERTY COMPILE_FLAGS " -Wno-typedef-redefinition")
 endif()
 if(GRN_WITH_XSIMD)
   if(GRN_WITH_SIMD_AVX)


include/groonga/groonga.h
```diff
@@ -757,6 +758,11 @@ typedef uint32_t grn_column_flags;
  * Indicates that the objects are compressed using Zstandard.
  */
 #define GRN_OBJ_COMPRESS_ZSTD (0x03 << 4)
+/**
+ * OpenZL compression.
+ * Indicates that the objects are compressed using OpenZL.
```
kou, line 763:
```suggestion
 * Indicates that the objects are compressed using OpenZL.
 * 
 * \since 15.2.2
```

lib/CMakeLists.txt
```diff
@@ -53,6 +53,14 @@ if(GRN_C_COMPILER_GNU_LIKE)
     SOURCE distance.cpp
     APPEND_STRING
     PROPERTY COMPILE_FLAGS " -Wno-unknown-pragmas -Wno-float-equal")
+  if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    # TODO: Remove this when OpenZL fixes duplicate typedef definitions
+    # in their headers (zl_opaque_types.h and other headers).
```
kou: Have you report this to OpenZL?
Could you add issue or PR URL?

lib/CMakeLists.txt
```diff
@@ -53,6 +53,14 @@ if(GRN_C_COMPILER_GNU_LIKE)
     SOURCE distance.cpp
     APPEND_STRING
     PROPERTY COMPILE_FLAGS " -Wno-unknown-pragmas -Wno-float-equal")
+  if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
```
kou: Do we need this check?
If this is needed, could you move this block to out of the `if(GRN_C_COMPILER_GNU_LIKE)` block?

lib/CMakeLists.txt
otegami: fix: [e353d23](https://github.com/groonga/groonga/pull/2677/commits/e353d232af3799e08e6e3bfe1fb328840d776236) I've reported it to the upstream and added the PR URL.

include/groonga/groonga.h
otegami: fix: [e55f696](https://github.com/groonga/groonga/pull/2677/commits/e55f69667abaff99237e52c9c4f7930113917761)

lib/CMakeLists.txt
otegami: fix: [368a89e](https://github.com/groonga/groonga/pull/2677/commits/368a89e1d57d5fc6aea61b8e258d367efd8fdaf8) We need it because the `typedef-redefinition` is the Clang specific flag. I've moved it to the out of this block.

otegami:
Thanks you for reviews, I've just addressed all of review comments, so far.

PR #2694, opened 2025-12-16
doc: add `doxygen-awesome-css` theme for C API documentation

otegami:
If doxygen-awesome-css is found, use it instead of the default theme. This provides a modern, clean
appearance with the following features:

- Dark mode toggle button
- Code fragment copy button
- Paragraph link buttons for easy linking
- Interactive table of contents

ref: https://github.com/jothepro/doxygen-awesome-css
ref: https://github.com/jothepro/doxygen-awesome-css/blob/main/docs/extensions.md

Doxyfile is renamed to Doxyfile.in to support CMake variable substitution for paths.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -49,16 +49,19 @@ if(NOT EXISTS "${GRN_DOC_PRE_BUILD_STAMP}")
             HINTS /usr/share/doxygen-awesome-css)
   if(DOXYGEN_AWESOME_CSS_DIR)
     message(STATUS "Found doxygen-awesome-css: ${DOXYGEN_AWESOME_CSS_DIR}")
+    set(DOXYGEN_HTML_HEADER "doc/header.html")
+    set(DOXYGEN_HTML_EXTRA_STYLESHEET
+        "${DOXYGEN_AWESOME_CSS_DIR}/doxygen-awesome.css")
+    set(DOXYGEN_HTML_EXTRA_FILES
+        "${DOXYGEN_AWESOME_CSS_DIR}/doxygen-awesome-darkmode-toggle.js \
+         ${DOXYGEN_AWESOME_CSS_DIR}/doxygen-awesome-fragment-copy-button.js \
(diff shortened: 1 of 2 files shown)


doc/CMakeLists.txt
```diff
@@ -44,15 +44,23 @@ endforeach()
...
 # TODO: When building packages in `packages/`, the documentation generated by Doxygen should be included.
 if(NOT EXISTS "${GRN_DOC_PRE_BUILD_STAMP}")
+  include(FetchContent)
+  fetchcontent_declare(
+    doxygen-awesome-css
+    URL https://github.com/jothepro/doxygen-awesome-css/archive/refs/heads/main.zip
+  )
+  fetchcontent_makeavailable(doxygen-awesome-css)
+  fetchcontent_getproperties(doxygen-awesome-css
+                             SOURCE_DIR DOXYGEN_AWESOME_CSS_DIR)
```
abetomo: https://github.com/groonga/groonga/tree/main/data/html/admin/assets contains files that have been fetched and committed.

Should we commit this instead of fetching it dynamically?

kou: Can we use system's one?
It seems that Debian provides a package for it: https://packages.debian.org/search?keywords=doxygen-awesome-css

otegami: Although I added the code to use system's one, some platforms don't provides the doxygen-awesome-css, so I still keep download logic here.

- Support doxygen-awesome-css packge
  - Ubuntu noble
  - Debian
- No support  doxygen-awesome-css packge
  - Ubuntu jammy 
  - The other platforms like AlmaLinux, macOS and so on.

kou: Can we use doxygen-awesome-css only when system doxygen-awesome-css exists?

otegami: fix: [ccf871f](https://github.com/groonga/groonga/pull/2694/commits/ccf871f8c649a0fc6ebd77ea3b6adfdecd91d8d6) Sure. If there is no package, we use the default theme instead of doxygen-awesome-theme.

PR #2709, opened 2026-01-03
cmake libedit: include cmake file in release tar

cho-m:
Currently, the release tarball (e.g. https://packages.groonga.org/source/groonga/groonga-15.2.1.tar.gz) cannot detect libedit due to missing CMake file.

I think this is due to CMake file not included in EXTRA_DIST.


kou:
Good catch!

PR #2724, opened 2026-01-30
Fix a compatibility bug of ShortBinary/Binary/LongBinary/JSON

kou:
If we upgrade a DB that was created by Groonga 15.2.2 or earlier to Groonga 15.2.3 or 15.2.4, the "schema" command crashed. Because ShortBinary/Binary/LongBinary/JSON type aren't registered to the DB yet.

We need to register them to a DB if they don't exit.

### DIFF
--- a/lib/db.c
+++ b/lib/db.c
@@ -452,7 +452,7 @@ grn_db_open_ensure_float32(grn_ctx *ctx, grn_db *db)
 {
   return grn_db_open_ensure_type(ctx,
                                  db,
-                                 GRN_DB_FLAOT32,
+                                 GRN_DB_FLOAT32,
                                  GRN_TYPE_FLOAT32_NAME,
                                  GRN_TYPE_FLOAT32_NAME_LEN,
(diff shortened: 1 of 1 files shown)


lib/db.c
```diff
@@ -416,38 +416,47 @@ grn_db_create(grn_ctx *ctx, const char *path, grn_db_create_optarg *optarg)
...
+#define GRN_TYPE_FLOAT32_NAME_LEN (sizeof(GRN_TYPE_FLOAT32_NAME) - 1)
+#define GRN_TYPE_FLOAT32_FLAGS    GRN_OBJ_KEY_FLOAT
+#define GRN_TYPE_FLOAT32_SIZE     sizeof(float)
+
+static bool
+grn_db_open_ensure_float32(grn_ctx *ctx, grn_db *db)
+{
+  return grn_db_open_ensure_type(ctx,
+                                 db,
+                                 GRN_DB_FLAOT32,
```
abetomo:
```suggestion
                                 GRN_DB_FLOAT32,
```

PR #2725, opened 2026-01-30
db: Add ensuring new types

abetomo:
Since databases created before adding a new type do not contain it, we will add a process to ensure the new type is added.


abetomo:
#2724

PR #2737, opened 2026-02-19
compressor openzl: extract the body-only compression case into a separate function

komainu8:
Extracted shared logic for header/body/footer compression.
Separated the header/body/footer and body-only compression paths into different functions to reduce complexity.

### DIFF
--- a/lib/compressor.c
+++ b/lib/compressor.c
@@ -942,18 +942,15 @@ openzl_compress_only_body(grn_ctx *ctx,
     return;
   }
   /*
-   * The body object is the same for both ZL_CCtx_compressMultiTypedRef()
-   * and ZL_TypedRef_free(). However, the two functions expect different
-   * types (const vs. non-const), so the same variable cannot be used
-   * directly for both.
-   * Ref:
-   *  - https://openzl.org/api/c/compress/#ZL_CCtx_compressMultiTypedRef
-   *  - https://openzl.org/api/c/compress/#ZL_TypedRef_free
+   * "ZL_CCtx_compressMultiTypedRef()" requires a value of type "const
+   * ZL_TypedRef *[]" as its fourth argument.
+   *   Ref: https://openzl.org/api/c/compress/#ZL_CCtx_compressMultiTypedRef
+   * Therefore, we prepare a "const ZL_TypedRef *[]" value here and pass it to
+   * "openzl_compress()".
+   * "ZL_CCtx_compressMultiTypedRef()" is called inside "openzl_compress()".
    *
-   * Therefore, we assign it to a "const ZL_TypedRef *[]" variable for
-   * ZL_CCtx_compressMultiTypedRef().
-   * Since the OpenZL tests also repack the values before using them,
-   * We followed the same approach here.
+   * Since the OpenZL tests uses the same approach, we are following it here as
+   * well.
    * (Ref:
    * https://github.com/facebook/openzl/blob/dev/tests/round_trip/test_n_to_n.cpp#L89-L109)
    */


lib/compressor.c
```diff
@@ -881,6 +881,86 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+  if (ZL_isError(zl_report)) {
+    ERR(GRN_OPENZL_ERROR,
+        "%s failed to compress data: %s",
+        tag,
+        ZL_ErrorCode_toString(ZL_errorCode(zl_report)));
+    return;
+  }
+  size_t zl_compressed_size = ZL_validResult(zl_report);
+  data->compressed_value_len = COMPRESSED_VALUE_LEN(zl_compressed_size);
+}
```
abetomo, line 917: Please insert a line break.

abetomo (approved):

lib/compressor.c
```diff
@@ -881,6 +881,88 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+   */
+  const ZL_TypedRef *input_temporary[] = {input[0]};
+  openzl_compress(ctx,
+                  tag,
+                  data,
+                  zl_cctx,
+                  zl_compressor,
+                  zl_value,
+                  zl_value_len_max,
+                  input_temporary,
```
kou: Can we use `&input[0]` here?


```suggestion
                  &input[0],
```

komainu8: Is this comment suggesting that we make a change like the one below?

```diff
diff --git a/lib/compressor.c b/lib/compressor.c
index 43ff83159..1b28dc4b1 100644
--- a/lib/compressor.c
+++ b/lib/compressor.c
@@ -889,7 +889,7 @@ openzl_compress(grn_ctx *ctx,
                 ZL_Compressor *zl_compressor,
                 void *zl_value,
                 const size_t zl_value_len_max,
-                const ZL_TypedRef *inputs[],
+                ZL_TypedRef *inputs[],
                 size_t n_inputs)
 {
   ZL_Report zl_report = ZL_CCtx_refCompressor(zl_cctx, zl_compressor);
@@ -941,15 +941,6 @@ openzl_compress_only_body(grn_ctx *ctx,
     ZL_TypedRef_free(input[0]);
     return;
   }
-  /*
-   * The input object is the same for both ZL_CCtx_compressMultiTypedRef()
-   * and ZL_TypedRef_free(). However, the two functions expect different
-   * types (const vs. non-const), so the same variable cannot be used
-   * directly for both.
-   *
-   * Therefore, we temporarily assign it to a "const ZL_TypedRef *" variable
-   * only when calling ZL_CCtx_compressMultiTypedRef().
-   */
   const ZL_TypedRef *input_temporary[] = {input[0]};
   openzl_compress(ctx,
                   tag,
@@ -958,7 +949,7 @@ openzl_compress_only_body(grn_ctx *ctx,
                   zl_compressor,
                   zl_value,
                   zl_value_len_max,
-                  input_temporary,
+                  &input[0],
                   1);
   ZL_TypedRef_free(input[0]);
 }
```

If so, this change would cause the warning described as below.

```
[1/21] Building C object lib/CMakeFiles/libgroonga.dir/compressor.c.o
/groonga/lib/compressor.c: In function ‘openzl_compress’:
/groonga/lib/compressor.c:906:45: warning: passing argument 4 of ‘ZL_CCtx_compressMultiTypedRef’ from incompatible pointer type [-Wincompatible-pointer-types]
  906 |                                             inputs,
      |                                             ^~~~~~
      |                                             |
      |                                             ZL_TypedRef ** {aka struct ZL_Input_s **}
In file included from /build-dir/groonga/_deps/openzl-src/include/openzl/openzl.h:8,
                 from /groonga/lib/compressor.c:46:
/build-dir/groonga/_deps/openzl-src/include/openzl/zl_compress.h:299:28: note: expected ‘const ZL_TypedRef **’ {aka ‘const struct ZL_Input_s **’} but argument is of type ‘ZL_TypedRef **’ {aka ‘struct ZL_Input_s **’}
  299 |         const ZL_TypedRef* inputs[],
      |         ~~~~~~~~~~~~~~~~~~~^~~~~~~~
/groonga/lib/compressor.c: In function ‘openzl_compress_only_body’:
/groonga/lib/compressor.c:944:22: warning: unused variable ‘input_temporary’ [-Wunused-variable]
  944 |   const ZL_TypedRef *input_temporary[] = {input[0]};
      |                      ^~~~~~~~~~~~~~~
[21/21] Linking C executable src/groonga
```

To prevent this warning, we need to repack the values into `const ZL_TypedRef *input_temporary[]` as in the existing implementation.

Or are you suggesting that we move the repacking logic into function `openzl_compress()`?

kou: Why do we need to remove `const` from `const ZL_TypedRef *inputs[],`? If we keep the `const`, does it generate another warning?

komainu8: Yes. If we keep const, a warning will occur in another place.

Variable `input` is passed as an argument to the following two functions:

(1). The 4th argument of `ZL_CCtx_compressMultiTypedRef()`
(2). The 1st argument of `ZL_TypedRef_free()`

The type required by (1) is `const ZL_TypedRef* inputs` (with const).
(Ref: https://openzl.org/api/c/compress/#ZL_CCtx_compressMultiTypedRef)

The type required by (2) is `ZL_TypedRef* tref` (without const).
(Ref: https://openzl.org/api/c/compress/#ZL_TypedRef_free)


Therefore, if we keep const, a warning will occur at the call site of function (2).
If we remove const, a warning will occur at the call site of function (1).

Since the OpenZL tests also repack the values before using them, I followed the same approach here.
(Ref: https://github.com/facebook/openzl/blob/dev/tests/round_trip/test_n_to_n.cpp#L89-L109)

kou: OK. Based on the OpenZL test code, we should not define `inputs` as `ZL_TypedRef *[]`:

```c
ZL_TypedRef *body = ZL_TypedRef_createSerial(data->body, data->body_len);
...
const ZL_TypedRef *inputs[] = {body};
ZL_CCtx_compressMultiTypedRef(..., inputs, ...);
ZL_TypedRef_free(body);
```

komainu8: I see. I've fixed it.

lib/compressor.c
```diff
@@ -881,6 +881,95 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+        "%s failed to select starting graph ID: %s",
+        tag,
+        ZL_ErrorCode_toString(ZL_errorCode(zl_report)));
+    ZL_TypedRef_free(body);
+    return;
+  }
+  /*
+   * The body object is the same for both ZL_CCtx_compressMultiTypedRef()
+   * and ZL_TypedRef_free(). However, the two functions expect different
+   * types (const vs. non-const), so the same variable cannot be used
```
kou: `ZL_CCtx_compressMultiTypedRef()` uses `const ZL_TypedRef *[]` but `ZL_TypedRef_free()` uses `ZL_TypedRef *`. `const` and `[]` are different.

komainu8: Ah...
I see. I've fixed it.

Issue #2650, opened 2025-11-18
language_model_vectorize: add `prefix` option


kou:
### What is your problem?

Some language models require prefix for input.

For example, https://huggingface.co/intfloat/multilingual-e5-base requires `passage: ` for search target texts and `query: ` for query texts.

`TokenLanguageModelKNN` has `passage_prefix` and `query_prefix` options. `language_model_vectorize()` should have `prefix` option like `TokenLanguageModelKNN` does.

PR #2738, opened 2026-02-23
language_model_vectorize: add `prefix` option

abetomo:
GitHub fixes GH-2650

### DIFF
--- a/plugins/functions/language_model.c
+++ b/plugins/functions/language_model.c
@@ -135,22 +135,28 @@ func_language_model_vectorize(grn_ctx *ctx,
     return NULL;
   }
 
-  grn_obj prefixed_text;
-  GRN_TEXT_INIT(&prefixed_text, 0);
+  grn_rc rc;
   if (prefix) {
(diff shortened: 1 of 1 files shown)


plugins/functions/language_model.c
```diff
@@ -106,11 +134,24 @@ func_language_model_vectorize(grn_ctx *ctx,
...
+
+  grn_obj prefixed_text;
+  GRN_TEXT_INIT(&prefixed_text, 0);
+  if (prefix) {
+    GRN_TEXT_PUT(ctx,
+                 &prefixed_text,
+                 GRN_TEXT_VALUE(prefix),
+                 GRN_TEXT_LEN(prefix));
+  }
+  GRN_TEXT_PUT(ctx, &prefixed_text, GRN_TEXT_VALUE(text), GRN_TEXT_LEN(text));
```
kou: Can we avoid copying the original text when `prefix` isn't specified?

komainu8 (approved):

PR #2739, opened 2026-02-25
compress openzl: add support for Float32 array compression with a preset dedicated to Float32

komainu8:
This modification implement the following features.

- compress with float32 preset

This modification doesn't implement the following features.

- compress with int preset
- learning input data and store learned model
- compress with learned model
- compress vector columns

---

The compression of float32 array results for the following data:

- The number of array elements is 40,960.
- The number of records is 10,000.

Using Float32 preset:

Compress with OpenZL: 1.3GB
Compress with ZSTD: 1.5GB
No compress: 1.6GB

Using the default preset, equivalent to ZSTD compression:

Compress with OpenZL: 1.5GB
Compress with ZSTD: 1.5GB
No compress: 1.6GB

### DIFF
--- a/lib/compressor.c
+++ b/lib/compressor.c
@@ -1011,9 +1011,9 @@ openzl_compress_only_body(grn_ctx *ctx,
   ZL_GraphID zl_graph_id = openzl_build_graph_id(zl_compressor, data);
   if (!ZL_GraphID_isValid(zl_graph_id)) {
     ERR(GRN_OPENZL_ERROR,
-        "%s Invalid API parameter passing for input data type. Input data type "
-        "is %d",
+        "%s Invalid API parameter passing for input data type: %s(%u)",
         tag,
+        grn_type_id_to_string_builtin(ctx, data->body_range),
         data->body_range);
     return;
   }


lib/compressor.c
```diff
@@ -881,6 +881,28 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
 #endif
 
 #ifdef GRN_WITH_OPENZL
+static inline ZL_GraphID
+openzl_build_graphid(ZL_Compressor *zl_compressor, grn_id body_range)
+{
+  switch (body_range) {
+  case GRN_DB_FLOAT32:
+    const ZL_GraphID successors_float32[] = {ZL_GRAPH_STORE, ZL_GRAPH_FSE};
+    return ZL_Compressor_registerStaticGraph_fromNode(
```
abetomo: https://openzl.org/api/c/compressor/#ZL_Compressor_registerStaticGraph_fromNode
> The newly created graph or ZL_GRAPH_ILLEGAL on error. The user may check for errors using ZL_GraphID_isValid().

It might be `ZL_GRAPH_ILLEGAL`, so better to check?

komainu8: Ah.
I've added it.

lib/compressor.c
```diff
@@ -916,6 +942,39 @@ openzl_compress(grn_ctx *ctx,
...
+    return ZL_TypedRef_createSerial(data->body, data->body_len);
+  }
+  if (data->body_len == 0) {
+    ERR(GRN_OPENZL_ERROR, "%s invalid input data", tag);
+    return NULL;
+  }
+  if ((data->body_column_flags & GRN_OBJ_COLUMN_TYPE_MASK) ==
+      GRN_OBJ_COLUMN_VECTOR) {
+    switch (data->body_range) {
+    case GRN_DB_FLOAT32:
```
abetomo, line 981: https://github.com/groonga/groonga/blob/8d1cbc4fa401fa9960ac0ffaab7147ee1509d290/lib/obj.c#L86-L91

こういう感じで `{ }` で囲んで書けますよ。

komainu8: おぉ。ありがとうございます。`{ }`で囲む書き方にしました。

lib/compressor.c
```diff
@@ -881,6 +881,32 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
 #endif
 
 #ifdef GRN_WITH_OPENZL
+static inline ZL_GraphID
+build_float32_graphid(ZL_Compressor *zl_compressor)
```
kou: Can we add `openzl_` prefix for all OpenZL related functions?

kou: Could you use `graph_id` not `graphid` for readability?

lib/compressor.c
```diff
@@ -881,6 +881,32 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+{
+  const ZL_GraphID successors[] = {ZL_GRAPH_STORE, ZL_GRAPH_FSE};
```
kou: Why do you use `successors` as variable name here?

It seems that `ZL_Compressor_registerStaticGraph_fromNode()` uses `dstGraphs` not `successors`: https://github.com/facebook/openzl/blob/e2842db6d082db8493c096eb3e7cfade61f22d29/include/openzl/zl_compressor.h#L264-L285

lib/compressor.c
```diff
@@ -881,6 +881,32 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+  return ZL_Compressor_registerStaticGraph_fromNode(
+    zl_compressor,
+    ZL_NODE_FLOAT32_DECONSTRUCT,
+    successors,
+    2); /*
+          This parameter must be equal to the number of outputs of
+          ZL_NODE_FLOAT32_DECONSTRUCT.
+          It is 2. Ref: openzl/include/openzl/codecs/zl_float_deconstruct.h
+        */
```
kou: This `2` is the number of elements of `successors`, right?

If so, we should add a comment to `ZL_NODE_FLOAT32_DECONSTRUCT`, `successors` and `2` not only `2`:


```suggestion
  /*
     ZL_NODE_FLOAT32_DECONSTRUCT outputs:
      * fixed-size-fields stream containing sign and fraction bits
      * serial stream containing exponent bits
    See also: openzl/include/openzl/codecs/zl_float_deconstruct.h
  */
  const ZL_GraphID successors[] = {ZL_GRAPH_STORE, ZL_GRAPH_FSE};
  const n_successors = sizeof(successors) / sizeof(ZL_GraphID);
  return ZL_Compressor_registerStaticGraph_fromNode(
    zl_compressor,
    ZL_NODE_FLOAT32_DECONSTRUCT,
    successors,
    n_successors);
```

lib/compressor.c
```diff
@@ -881,6 +881,32 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+{
+  const ZL_GraphID successors[] = {ZL_GRAPH_STORE, ZL_GRAPH_FSE};
+  return ZL_Compressor_registerStaticGraph_fromNode(
+    zl_compressor,
+    ZL_NODE_FLOAT32_DECONSTRUCT,
+    successors,
+    2); /*
+          This parameter must be equal to the number of outputs of
+          ZL_NODE_FLOAT32_DECONSTRUCT.
+          It is 2. Ref: openzl/include/openzl/codecs/zl_float_deconstruct.h
```
kou: In general, an URL not a path is easier to refer:

https://github.com/facebook/openzl/blob/25611c7663721aa5ca0c8af302444682820151e5/include/openzl/codecs/zl_float_deconstruct.h#L12-L19

lib/compressor.c
```diff
@@ -916,6 +942,41 @@ openzl_compress(grn_ctx *ctx,
   data->compressed_value_len = COMPRESSED_VALUE_LEN(zl_compressed_size);
 }
 
+static inline ZL_TypedRef *
+openzl_create_typeref(grn_ctx *ctx, const char *tag, grn_compress_data *data)
```
kou: `typed_ref`?

lib/compressor.c
```diff
@@ -916,6 +942,41 @@ openzl_compress(grn_ctx *ctx,
...
+{
+  if (data->header_len > 0 || data->footer_len > 0) {
+    return ZL_TypedRef_createSerial(data->body, data->body_len);
+  }
```
kou: Why do we not use `header` and `footer` when they exist?

lib/compressor.c
```diff
@@ -916,6 +942,41 @@ openzl_compress(grn_ctx *ctx,
...
+  if (data->body_len == 0) {
+    ERR(GRN_OPENZL_ERROR, "%s invalid input data", tag);
```
kou: We should explain why the data is invalid. Users use the error message to fix their data.

See also: https://www.clear-code.com/blog/2009/4/10.html

lib/compressor.c
```diff
@@ -916,6 +942,41 @@ openzl_compress(grn_ctx *ctx,
...
+  }
+  if ((data->body_column_flags & GRN_OBJ_COLUMN_TYPE_MASK) ==
+      GRN_OBJ_COLUMN_VECTOR) {
+    switch (data->body_range) {
+    case GRN_DB_FLOAT32:
+      {
+        size_t width = sizeof(float);
+        if (data->body_len % width != 0) {
+          ERR(GRN_OPENZL_ERROR,
+              "%s data->body_len is invalid: %d",
```
kou: Should we use `data->body_len` here...? It's an internal data that users don't know...

lib/compressor.c
```diff
@@ -925,14 +986,18 @@ openzl_compress_only_body(grn_ctx *ctx,
                           void *zl_value,
                           const size_t zl_value_len_max)
 {
-  ZL_TypedRef *body = ZL_TypedRef_createSerial(data->body, data->body_len);
+  ZL_GraphID zl_graphid = openzl_build_graphid(zl_compressor, data->body_range);
+  if (zl_graphid.gid == ZL_GRAPH_ILLEGAL.gid) {
+    ERR(GRN_OPENZL_ERROR, "%s failed to build graphid", tag);
```
kou: Can we show why the build is failed?

```suggestion
    ERR(GRN_OPENZL_ERROR, "%s failed to build graph ID", tag);
```

test/command/suite/select/output/openzl/vector/compressed/float32.expected
kou, line 1: Do we add more `compressed_XXX.{test,expected}`?
If so, we should create `.../vector/compressed/XXX.{test,expected}`.

lib/compressor.c
```diff
@@ -916,6 +942,41 @@ openzl_compress(grn_ctx *ctx,
...
+        if (data->body_len % width != 0) {
+          ERR(GRN_OPENZL_ERROR,
+              "%s data->body_len is invalid: %d",
+              tag,
+              data->body_len);
+          return NULL;
+        }
+        return ZL_TypedRef_createNumeric(data->body,
+                                         width,
+                                         data->body_len / width);
```
kou: It seems that `data->body_n_elements` exists.

lib/compressor.c
```diff
@@ -916,6 +942,41 @@ openzl_compress(grn_ctx *ctx,
...
+  if (data->body_len == 0) {
+    ERR(GRN_OPENZL_ERROR, "%s invalid input data", tag);
+    return NULL;
+  }
+  if ((data->body_column_flags & GRN_OBJ_COLUMN_TYPE_MASK) ==
+      GRN_OBJ_COLUMN_VECTOR) {
+    switch (data->body_range) {
+    case GRN_DB_FLOAT32:
+      {
+        size_t width = sizeof(float);
```
kou: Should we add `data->body_element_size`?

lib/compressor.c
```diff
@@ -925,14 +998,19 @@ openzl_compress_only_body(grn_ctx *ctx,
                           void *zl_value,
                           const size_t zl_value_len_max)
 {
-  ZL_TypedRef *body = ZL_TypedRef_createSerial(data->body, data->body_len);
+  ZL_GraphID zl_graph_id =
+    openzl_build_graph_id(zl_compressor, data->body_range);
+  if (zl_graph_id.gid == ZL_GRAPH_ILLEGAL.gid) {
```
abetomo: Would it be better to use ZL_GraphID_isValid() for the check?

lib/compressor.c
komainu8: Ah.
This is mistake. I've fixed it.

lib/compressor.c
komainu8: > This 2 is the number of elements of successors, right?

Yes.
I've fixed and added comment.

lib/compressor.c
komainu8: I've used URL.

lib/compressor.c
komainu8: Ah...
This is needless currently.
Because this modification compress only body data.
Therefore, I removed this condition.

lib/compressor.c
komainu8: I reviewed the error message for all openzl error.

lib/compressor.c
komainu8: Ah...I see.
I reviewed it.

lib/compressor.c
komainu8: Since the API does not return a message that directly indicates the cause of the error, I now display a message that may help identify the cause within the available information.

test/command/suite/select/output/openzl/vector/compressed/float32.expected
komainu8: > Do we add more compressed_XXX.{test,expected}?

yes.
I created `../vector/compressed/XXX.{test,expected}`.

lib/compressor.c
komainu8: I used it.

lib/compressor.c
komainu8: I added `body_element_size` in `data`.

lib/compressor.c
komainu8: I used it.

lib/compressor.c
```diff
@@ -881,6 +881,44 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+static inline ZL_GraphID
+openzl_build_float32_graph_id(ZL_Compressor *zl_compressor)
+{
+  /*
+    ZL_NODE_FLOAT32_DECONSTRUCT outputs:
+      * fixed-size-fields stream containing sign and fraction bits
+      * serial stream containing exponent bits
+    See also:
+      https://github.com/facebook/openzl/blob/v0.1.0/include/openzl/codecs/zl_float_deconstruct.h#L12-L18
+
```
komainu8:
```suggestion
```

lib/compressor.c
```diff
@@ -925,14 +995,22 @@ openzl_compress_only_body(grn_ctx *ctx,
                           void *zl_value,
                           const size_t zl_value_len_max)
 {
-  ZL_TypedRef *body = ZL_TypedRef_createSerial(data->body, data->body_len);
+  ZL_GraphID zl_graph_id =
+    openzl_build_graph_id(zl_compressor, data->body_range);
+  if (!ZL_GraphID_isValid(zl_graph_id)) {
+    ERR(GRN_OPENZL_ERROR,
+        "%s Invalid paramete for input data type. Input data type is %d",
```
komainu8:
```suggestion
        "%s Invalid API parameter passing for input data type. Input data type is %d",
```

lib/compressor.c
```diff
@@ -881,6 +881,43 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+  */
+  const size_t n_dst_graphs = sizeof(dst_graphs) / sizeof(ZL_GraphID);
+  return ZL_Compressor_registerStaticGraph_fromNode(zl_compressor,
+                                                    ZL_NODE_FLOAT32_DECONSTRUCT,
+                                                    dst_graphs,
+                                                    n_dst_graphs);
+}
+
+static inline ZL_GraphID
+openzl_build_graph_id(ZL_Compressor *zl_compressor, grn_id body_range)
```
kou: Is "build" a good name? OpenZL API uses "register" not "build".

lib/compressor.c
```diff
@@ -916,6 +953,51 @@ openzl_compress(grn_ctx *ctx,
...
 
+static inline ZL_TypedRef *
+openzl_create_type_ref(grn_ctx *ctx, const char *tag, grn_compress_data *data)
+{
+  if (data->body_len == 0) {
+    ERR(GRN_OPENZL_ERROR,
+        "%s invalid input data: Compress target size is 0",
+        tag);
+    return NULL;
+  }
```
kou: Do we need this check?
Does callers check this?

lib/compressor.c
```diff
@@ -916,6 +953,51 @@ openzl_compress(grn_ctx *ctx,
...
+    ((data->body_column_flags & GRN_OBJ_WEIGHT_BFLOAT16) != 0);
+
+  if (is_vector_column && !is_with_weight && !is_weight_float32 &&
+      !is_weight_bfloat16) {
+    switch (data->body_range) {
+    case GRN_DB_FLOAT32:
+      {
+        if (data->body_len % data->body_element_size != 0) {
+          ERR(GRN_OPENZL_ERROR,
+              "%s : Data size is invalid. May be this data is broken: "
```
kou:
```suggestion
              "%s Data size is invalid. May be this data is broken: "
```

lib/compressor.c
```diff
@@ -916,6 +953,51 @@ openzl_compress(grn_ctx *ctx,
...
+              "data_size=%" GRN_FMT_INT32U ", element_size=%zu, n_elements=%zu",
+              tag,
+              data->body_len,
+              data->body_element_size,
+              data->body_n_elements);
+          return NULL;
+        }
```
kou: Do we need this check?
May callers pass invalid data?

lib/compressor.c
```diff
@@ -916,6 +953,51 @@ openzl_compress(grn_ctx *ctx,
...
+  bool is_vector_column = ((data->body_column_flags &
+                            GRN_OBJ_COLUMN_TYPE_MASK) == GRN_OBJ_COLUMN_VECTOR);
+  bool is_with_weight = ((data->body_column_flags & GRN_OBJ_WITH_WEIGHT) != 0);
+  bool is_weight_float32 =
+    ((data->body_column_flags & GRN_OBJ_WEIGHT_FLOAT32) != 0);
+  bool is_weight_bfloat16 =
+    ((data->body_column_flags & GRN_OBJ_WEIGHT_BFLOAT16) != 0);
+
+  if (is_vector_column && !is_with_weight && !is_weight_float32 &&
+      !is_weight_bfloat16) {
```
kou: Do we need `!is_weight_float32` and `!is_weight_bfloat16` checks here?
Is `!is_with_weight` enough?

lib/compressor.c
```diff
@@ -881,6 +881,43 @@ grn_compressor_decompress_blosc(grn_ctx *ctx, grn_decompress_data *data)
...
+  /*
+    n_dst_graphs must be equal to the number of ZL_NODE_FLOAT32_DECONSTRUCT
+    output items defined in the API documentation for
+    ZL_Compressor_registerStaticGraph_fromNode().
+
+    See also:
+      https://openzl.org/api/c/compressor/#ZL_Compressor_registerStaticGraph_fromNode
+  */
+  const size_t n_dst_graphs = sizeof(dst_graphs) / sizeof(ZL_GraphID);
+  return ZL_Compressor_registerStaticGraph_fromNode(zl_compressor,
```
abetomo: https://github.com/facebook/openzl/blob/67ce126ee45476ea11adce3d627447430875b162/include/openzl/zl_compressor.h#L171-L324

` * @{`  ~ ` * @}` というコメントで囲まれていて、該当の関数（`ZL_Compressor_registerStaticGraph_fromNode` とか）と離れた場所にコメントが書いてあって、ちょっとわかりにくいのだけど、`ZL_Compressor_registerStaticGraph_fromNode` はそのうち消えるかもだから、`ZL_Compressor_buildStaticGraph` を使ったほうが良いよ！って話？

lib/compressor.c
abetomo: According to a comment from https://github.com/facebook/openzl/blob/67ce126ee45476ea11adce3d627447430875b162/include/openzl/zl_compressor.h#L171-L324 , it's better to use ZL_Compressor_buildStaticGraph, so we used ZL_Compressor_buildStaticGraph.
Therefore, its name is kept as "build".

lib/compressor.c
```diff
@@ -925,20 +1008,29 @@ openzl_compress_only_body(grn_ctx *ctx,
...
-  ZL_TypedRef *body = ZL_TypedRef_createSerial(data->body, data->body_len);
-  if (!body) {
-    ERR(GRN_OPENZL_ERROR, "%s failed to allocate input buffer", tag);
+  ZL_GraphID zl_graph_id = openzl_build_graph_id(zl_compressor, data);
+  if (!ZL_GraphID_isValid(zl_graph_id)) {
+    ERR(GRN_OPENZL_ERROR,
+        "%s Invalid API parameter passing for input data type. Input data type "
+        "is %d",
+        tag,
+        data->body_range);
```
kou:
```suggestion
        "%s Invalid API parameter passing for input data type: %s(%u)",
        tag,
        grn_type_id_to_string_builtin(ctx, data->body_range),
        data->body_range);
```

PR #2748, opened 2026-03-11
json: add support for true/false/null/number in array

kou:
String, object and array aren't supported yet.


kou:
I'll refactor this before I merge this.

PR #2749, opened 2026-03-13
groonga: ensure writing data

kou:
If output (e.g. pipe) may not have enough buffer, `fwrite()` may not write all data. We need to call multiple `fwrite()`s to ensure writing all data.

In my experience, non blocking output may not work well. `fwrite()` against non blocking output may not return `-1` with `EAGAIN` and reported written data may not be written.

So, output should be blocking.

See also: https://github.com/groonga/grntest/pull/31

### DIFF
--- a/src/groonga.c
+++ b/src/groonga.c
@@ -348,6 +348,21 @@ output_envelope(
   grn_output_envelope(ctx, rc, head, body, foot, input_path, number_of_lines);
 }
 
+static void
+ensure_write(FILE *output, const char *data, size_t size)
+{
+  const char *rest_data = data;
+  size_t rest_size = size;
+  while (rest_size > 0) {
+    size_t written = fwrite(rest_data, 1, rest_size, output);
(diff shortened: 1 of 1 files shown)


src/groonga.c
```diff
@@ -348,6 +348,25 @@ output_envelope(
...
+ensure_write(FILE *output, const char *data, size_t size)
+{
+  const char *rest_data = data;
+  size_t rest_size = size;
+  while (rest_size > 0) {
+    struct timespec before;
+    clock_gettime(CLOCK_REALTIME, &before);
+    size_t written = fwrite(rest_data, 1, rest_size, output);
+    struct timespec after;
+    clock_gettime(CLOCK_REALTIME, &after);
```
abetomo: Since we won't be using `before` and `after` anymore, could we remove them?

kou: Oh, sorry. This was for a debug print.

PR #2750, opened 2026-03-16
clang-format: add `SeparateDefinitionBlocks: Always`

abetomo:
An empty line is inserted between blocks.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#separatedefinitionblock

The setting is the same as Mroonga: https://github.com/mroonga/mroonga/pull/1070

### DIFF
--- a/lib/grn_ii.h
+++ b/lib/grn_ii.h
@@ -42,8 +42,9 @@ struct _grn_ii {
   grn_table_flags lflags;
   grn_encoding encoding; /* Character encoding */
                          /* This member is used for matching */
-  uint32_t n_elements;   /* Number of elements in postings */
-                         /* rid, [sid], tf, [weight] and [pos] */
+  uint32_t n_elements;   /* Number of elements in postings. rid, [sid], tf,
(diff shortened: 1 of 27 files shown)


lib/grn_ii.h
```diff
@@ -42,8 +42,9 @@ struct _grn_ii {
   grn_table_flags lflags;
   grn_encoding encoding; /* Character encoding */
                          /* This member is used for matching */
-  uint32_t n_elements;   /* Number of elements in postings */
-                         /* rid, [sid], tf, [weight] and [pos] */
+  uint32_t n_elements;   /* Number of elements in postings. rid, [sid], tf,
+                            [weight] and [pos] */
+
```
abetomo, line 47: Since the line breaks weren't inserted as expected, I manually edited the comments.

abetomo:
Unlike Mroonga, there are a lot of differences.
Should we include this rule in Groonga as well?

kou:
Hmm. We don't need this for Groonga...

Issue #1941, opened 2024-09-03
Add TokenCharabia


kou:
### What is your problem?

https://github.com/meilisearch/charabia

It seems that Charabia detects the input language and selects suitable module automatically.

The current morphological analyzer based tokenizer such as `TokenMecab` isn't multilingual at a time. If Charabia works on multiple languages in a text, we can support multiple languages in a column value.

PR #2780, opened 2026-04-03
tokenizer: add TokenMecabKo tokenizer plugin for Korean morphological analysis

ji-heo111:
Add a new tokenizer plugin "TokenMecabKo" based on mecab-ko, a Korean fork of MeCab. This tokenizer enables Korean full-text search in Groonga by leveraging mecab-ko and mecab-ko-dic for Korean morphological analysis.

Key features:
- Registered as "TokenMecabKo" to coexist with the existing "TokenMecab" for Japanese
- Feature locations adjusted for mecab-ko dictionary format: POS, semantic class, jongseong (final consonant), reading, type, start POS, end POS, expression
- Korean and fullwidth delimiter characters support
- Chunked tokenization support for large texts
- Environment variables: GRN_MECAB_KO_CHUNKED_TOKENIZE_ENABLED, GRN_MECAB_KO_CHUNK_SIZE_THRESHOLD
- Tokenizer options: chunked_tokenize, chunk_size_threshold, include_class, include_reading, include_form, use_reading, target_class

Since mecab-ko is API-compatible with MeCab (shares libmecab), the plugin builds under the existing GRN_WITH_MECAB CMake option with no additional dependencies.

New files:
- plugins/tokenizers/mecab-ko.c
- plugins/tokenizers/mecab_ko_sources.am

Modified files:
- plugins/tokenizers/CMakeLists.txt
- plugins/tokenizers/Makefile.am

Usage:
  SELECT pgroonga_command('plugin_register tokenizers/mecab-ko'); -- Then use 'TokenMecabKo' as the tokenizer

### DIFF
--- a/packages/yum/groonga.spec.in
+++ b/packages/yum/groonga.spec.in
@@ -494,6 +494,7 @@ exit 0
 %files tokenizer-mecab
 %defattr(-,root,root,-)
 %{_libdir}/groonga/plugins/tokenizers/mecab.so
+%{_libdir}/groonga/plugins/tokenizers/mecab-ko.so
 %endif
 
 %files token-filter-stem


abetomo:
Thanks for working on Korean tokenizer support!

There are several repositories named `mecab-ko`.
Which one does this plugin target? Is it https://bitbucket.org/eunjeon/mecab-ko/ ?

It seems that https://github.com/bab2min/Kiwi has become the de facto standard for Korean morphological analysis recently.
If we're adding Korean tokenizer support to Groonga now, Kiwi might be a better choice.
What do you think?

ji-heo111:
Yes, it targets https://bitbucket.org/eunjeon/mecab-ko/.

Thank you for suggesting Kiwi. I agree that Kiwi is a better choice — it has the same LGPL 2.1 license, provides a C API, bundles its own model (no separate dictionary installation needed), and is actively maintained. I'll work on a TokenKiwi implementation instead.

ji-heo111:
@abetomo 
I’ve taken a deeper look into Kiwi, and here is a detailed comparison with mecab-ko regarding their integration as a Groonga tokenizer.

Where Kiwi excels:
* Analysis accuracy: Kiwi uses n-gram language models (Knlm/CoNgram) for context-aware disambiguation, achieving ~94% accuracy on written text vs mecab-ko's simpler bigram cost matrix approach.
* Typo tolerance: Kiwi has a built-in TypoTransformer that generates typo correction candidates with cost-based scoring. This is valuable for user-facing search queries.
* Unknown word handling: Kiwi uses a character-level CoNgramModel for OOV words, which is more sophisticated than mecab-ko's character-type-based grouping.

Where mecab-ko is better suited for FTS:
* Throughput: mecab-ko's pipeline is minimal — DARTS trie lookup + Viterbi with a simple cost matrix. Kiwi adds language model scoring, rule-based penalties, morpheme decomposition, and SIMD-accelerated search on top of its Viterbi. For FTS indexing where millions of documents need to be tokenized, this difference in per-token cost accumulates significantly.
* Memory efficiency: mecab-ko's dictionary is ~50-200MB (mmap'd, shared across processes). Kiwi's model (dictionary + language model) is ~278MB per instance. This matters for concurrent indexing/search threads.
* Thread safety model: mecab-ko allows sharing a single Model object across threads — each thread only needs its own lightweight Tagger + Lattice. The dictionary is immutable after loading, so no read locks are needed. Kiwi does not support sharing a single instance across threads. You either need per-thread instances (~278MB each) or use its built-in ThreadPool with asyncAnalyze(), which is less flexible for integration into Groonga's own threading model.
* Build simplicity: mecab-ko requires only C++98 with no external dependencies. Kiwi requires C++17 and optionally depends on mimalloc, cpuinfo, and Eigen. This affects portability across the environments Groonga supports.
* API surface for integration: mecab-ko's API is minimal — Model::createTagger(), tagger->parseToNode(), iterate nodes. Writing a thin Groonga tokenizer wrapper is straightforward. Kiwi's API involves a KiwiBuilder → Kiwi → analyze() → TokenResult pipeline with richer but more complex integration points.
* Dictionary customization: mecab-ko supports user dictionaries via simple CSV files compiled into binary format. This is well-documented and easy for users to add domain-specific terms. Kiwi requires programmatic morpheme addition through KiwiBuilder.

In a Full-Text Search context, consistency (producing the same tokens at both index and query time) is often more critical than absolute linguistic accuracy. Given its speed and lightweight integration, mecab-ko seems like the more pragmatic starting point.
However, Kiwi’s ability to handle typos and neologisms is definitely superior for modern user-generated content.
What do you think about this? I’d love to hear your thoughts on which direction fits Groonga’s current priorities better.

kou:
It seems that mecab-ko isn't maintained long time. So I'm negative to add a separated tokenizer for mecab-ko.
BTW, can we use the original MeCab (not mecab-ko) for Korean with a Korean dictionary? If we can use the original MeCab, we don't need to add a separated tokenizer for Korean.

If we need a separated tokenizer for mecab-ko, we can implement it as a plugin like https://github.com/groonga/groonga-tokenizer-friso .

We can also implement a tokenizer for Kiwi as a plugin. If you're interesting it, I can create https://github.com/groonga/groonga-tokenizer-kiwi and we can maintain it together.

kou:
Related: #1941

Charabia https://github.com/meilisearch/charabia based tokenizer may fit your use case. We can use Lindera https://github.com/lindera/lindera through Charabia.

ji-heo111:
@kou
Thank you for the detailed suggestions!
I investigated each option:
**Original MeCab + Korean dictionary: Not feasible**
mecab-ko has an engine-level modification — `left-space-penalty-factor` in `connector.h`/`connector.cpp` (~60 lines changed). This adds a cost penalty when a token with specific POS tags (e.g., postpositions) appears right after a space. Korean uses spaces between words unlike Japanese, so without this, MeCab produces incorrect segmentation:
```
Input: "화학 이외의 것"
Original MeCab:  화학 / 이 / 외 / 의 / 것    ← incorrect: splits "이외" into "이"+"외"
mecab-ko:        화학 / 이외 / 의 / 것         ← correct
```
This is configured via `left-space-penalty-factor` in dicrc, but original MeCab simply ignores this parameter. Since mecab-ko-dic depends on this feature, we cannot use original MeCab with the Korean dictionary.

**Kiwi: Performance concerns**
Kiwi is significantly slower than MeCab and has more complex memory management, which makes it less suitable for a search engine tokenizer that needs to handle high throughput.

**My proposal — two approaches in priority order:**
1. **`groonga-tokenizer-mecab-ko` plugin** (first priority)
   - Same approach as groonga-tokenizer-friso
   - mecab-ko is widely adopted in the Korean NLP community (used by Elasticsearch's nori analyzer, etc.)
   - The code diff from original MeCab is minimal, so maintenance cost is low
   - I can start working on this
2. **Charabia/Lindera based tokenizer** (second priority)
   - Charabia supports Korean through Lindera
   - Worth exploring as a longer-term, Rust-based alternative

Would you be willing to create the `groonga-tokenizer-mecab-ko` repository so we can start collaborating on it?

kou:
OK. I'll create `groonga-tokenizer-mecab-ko` and prepare the first content based on the existing `TokenMecab`.

kou:
Created and prepared: https://github.com/groonga/groonga-tokenizer-mecab-ko

But CI is still broken. Let's fix it as the first step.

Let's discuss in the repository for further development.

PR #2799, opened 2026-05-04
cmake blosc: fix Blosc2 version check to allow compatible major versions

PPN-SD:
Blosc2's CMake file only accepts a requested version when the major version matches the installed one. Requesting 2.10.0 fails even if 3.X system version is installed and compiled fine. Rework the detection logic.


CMakeLists.txt
```diff
@@ -1446,16 +1446,15 @@ if(NOT "${GRN_WITH_BLOSC}" STREQUAL "no")
   if("${GRN_WITH_BLOSC}" STREQUAL "bundled")
     set(Blosc2_FOUND FALSE)
   else()
-    if("${GRN_WITH_BLOSC}" STREQUAL "system")
-      find_package(Blosc2 ${GRN_BLOSC_REQUIRED_VERSION} REQUIRED)
-    else()
-      find_package(Blosc2 ${GRN_BLOSC_REQUIRED_VERSION})
-    endif()
+    find_package(Blosc2 QUIET)
```
kou: I want to check the Blosc version in this clause something like the following:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 26f261387..edd0b514d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1431,10 +1431,14 @@ if(NOT "${GRN_WITH_BLOSC}" STREQUAL "no")
   if("${GRN_WITH_BLOSC}" STREQUAL "bundled")
     set(Blosc2_FOUND FALSE)
   else()
-    if("${GRN_WITH_BLOSC}" STREQUAL "system")
-      find_package(Blosc2 ${GRN_BLOSC_REQUIRED_VERSION} REQUIRED)
-    else()
-      find_package(Blosc2 ${GRN_BLOSC_REQUIRED_VERSION})
+    find_package(Blosc2)
+    if(Blosc2_FOUND AND Blosc2_VERSION VERSION_LESS GRN_BLOSC_REQUIRED_VERSION)
+      if("${GRN_WITH_BLOSC}" STREQUAL "system")
+        message(
+          FATAL_ERROR
+            "Blosc is old: ${Blosc2_VERSION} < ${GRN_BLOSC_REQUIRED_VERSION}")
+      endif()
+      set(Blosc2_FOUND FALSE)
     endif()
   endif()
   if(Blosc2_FOUND)
```

PPN-SD: Ok, slightly adapted so it'll fail as before with "system" if blosc is not found.

CMakeLists.txt
```diff
@@ -1446,10 +1446,14 @@ if(NOT "${GRN_WITH_BLOSC}" STREQUAL "no")
...
-    else()
-      find_package(Blosc2 ${GRN_BLOSC_REQUIRED_VERSION})
+    find_package(Blosc2)
+    if(Blosc2_FOUND AND Blosc2_VERSION VERSION_LESS GRN_BLOSC_REQUIRED_VERSION)
+      set(Blosc2_FOUND FALSE)
+    endif()
+    if(NOT Blosc2_FOUND AND "${GRN_WITH_BLOSC}" STREQUAL "system")
+      message(
+        FATAL_ERROR
+          "Blosc >= ${GRN_BLOSC_REQUIRED_VERSION} not found")
```
kou:
```suggestion
      message(FATAL_ERROR "Blosc >= ${GRN_BLOSC_REQUIRED_VERSION} not found")
```

kou:
Thanks.

PR #2829, opened 2026-06-17
packages debian: Drop support for bookworm

kou:
It reached EOL on 2026-06-10.


abetomo:
https://github.com/groonga/groonga/blob/9687edd492099e8ccbcf9e4ac314be3956167e18/packages/apt/test.sh#L90-L93

It would be better to update this part here as well.

kou:
Good catch.

We detect the latest stable Debian code name now:

https://github.com/groonga/groonga/actions/runs/27660740152/job/81805107817?pr=2829#step:14:3949

```text
+ '[' trixie = trixie ']'
```

abetomo (approved):

PR #2852, opened 2026-07-09
ci cmake: stop installing outdated system xsimd

komainu8:
Currently, we install libxsimd-dev in the "Install common build dependencies" step.
However, the system xsimd version in the CI environment is outdated.

As a result, CMake attempts to use the bundled xsimd instead of the system one.
This cause a conflict because the "add_library" target for xsimd already exist.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -127,7 +127,6 @@ jobs:
             build-essential \
             ccache \
             cmake \
-            libxsimd-dev \
             ninja-build
       - uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


komainu8:
The CI is failing as show below, but the failure is unrelated to this change.

https://github.com/groonga/groonga/actions/runs/28984331436/job/86030664859?pr=2852#step:9:3952

```
CMake Error at /usr/lib/cmake/Blosc2/Blosc2Targets.cmake:108 (message):
The imported target "Blosc2::blosc2_static" references the file
"/usr/lib/libblosc2.a"
```

abetomo (approved):

PR #2858, opened 2026-07-13
benchmark: Remove incorrect G_GNUC_CONST usage

### LINKED ISSUE #2857: Remove `G_GNUC_CONST` from `bench_reporter_get_type` declaration
### What happend?

GLib recently removed `G_GNUC_CONST` attributes from its *get_type declarations to avoid miscompilations with GCC trunk:
- https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5223
- https://gitlab.gnome.org/GNOME/glib/-/work_items/3984

I believe a similar change is needed for `bench_reporter_get_type`.

More background reading:
- https://blogs.gnome.org/mcatanzaro/2026/06/29/your-_get_type-function-is-not-g_gnuc_const-part-two/
- https://gcc.gnu.org/onlinedocs/gcc/Common-Attributes.html#index-const

### How to reproduce it

N/A

### Expected behavior

N/A

### Environment

N/A

### Additional context

_No response_

correctmost:
GitHub: Fix GH-2857

`bench_reporter_get_type` has global side effects, which makes it ineligible for the `G_GNUC_CONST` compiler attribute.

See https://gitlab.gnome.org/GNOME/glib/-/commit/016829559 for further information about how incorrect usage can lead to issues with development versions of GCC.

See also:
* https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5223
* https://gitlab.gnome.org/GNOME/glib/-/work_items/3984

### DIFF
--- a/benchmark/lib/bench-reporter.h
+++ b/benchmark/lib/bench-reporter.h
@@ -47,7 +47,7 @@ struct _BenchReporterClass
   GObjectClass parent_class;
 };
 
-GType           bench_reporter_get_type  (void) G_GNUC_CONST;
+GType           bench_reporter_get_type  (void);
 
 BenchReporter  *bench_reporter_new       (void);
 void            bench_reporter_register  (BenchReporter     *reporter,


kou:
Thanks.

PR #2860, opened 2026-07-14
string: implement `string_truncate()`

abetomo:
func_string_truncate() truncates a string to at most `length` characters.
If truncation occurs, the tail of the kept portion is replaced so that the result (kept portion + omission) is `length` characters long, in the same way as Ruby on Rails' String#truncate.
https://api.rubyonrails.org/classes/String.html#method-i-truncate

Note:
If `omission` itself is longer than `length`, the result is `omission` alone.  This is the same as Ruby on Rails' String#truncate.

### DIFF
--- a/plugins/functions/string.c
+++ b/plugins/functions/string.c
@@ -25,6 +25,21 @@
 
 #include <groonga/plugin.h>
 
+static int64_t
+string_n_chars(grn_ctx *ctx, const char *string, size_t string_size)
+{
+  const char *p = string;
+  const char *end = string + string_size;
+  int64_t n_chars = 0;
+  int char_size = 0;
+
+  for (; p < end && (char_size = grn_charlen(ctx, p, end)); p += char_size) {
+    n_chars++;
+  }
+
+  return n_chars;
+}
+
 /*
  * func_string_length() returns the number of characters in a string.
  * If the string contains an invalid byte sequence, this function returns the
@@ -63,16 +78,7 @@ func_string_length(grn_ctx *ctx, int n_args, grn_obj **args,
     return NULL;
   }
 
-  {
-    const char *s = GRN_TEXT_VALUE(target);
-    const char *e = GRN_TEXT_VALUE(target) + GRN_TEXT_LEN(target);
-    const char *p;
-    int cl = 0;
-    for (p = s; p < e && (cl = grn_charlen(ctx, p, e)); p += cl) {
-      length++;
-    }
-  }
-
+  length = string_n_chars(ctx, GRN_TEXT_VALUE(target), GRN_TEXT_LEN(target));
   grn_length = grn_plugin_proc_alloc(ctx, user_data, GRN_DB_UINT32, 0);
   if (!grn_length) {
     return NULL;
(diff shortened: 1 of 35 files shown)


plugins/functions/string.c
```diff
@@ -264,6 +270,136 @@ func_string_substring(grn_ctx *ctx, int n_args, grn_obj **args,
...
+  grn_obj *result = grn_plugin_proc_alloc(ctx, user_data, target->header.domain, 0);
+  if (!result) {
+    return NULL;
+  }
+
+  int64_t max_n_body_chars = max_n_chars - n_omission_chars;
+  if (max_n_body_chars <= 0) {
+    GRN_TEXT_SET(ctx, result, omission_value, omission_size);
+    return result;
+  }
```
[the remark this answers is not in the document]
abetomo:
```
redmine-app(prod):001> "aa".truncate(1)
=> "..."
redmine-app(prod):002> "abc".truncate(2)
=> "..."
```

plugins/functions/string.c
```diff
@@ -264,6 +270,136 @@ func_string_substring(grn_ctx *ctx, int n_args, grn_obj **args,
...
+                         int64_t max_n_chars, int64_t max_n_body_chars, const char **body_end)
+{
+  const char *p = string;
+  const char *end = string + string_size;
+  int64_t n_chars = 0;
+  int char_size = 0;
+
+  if (max_n_chars < 0) {
+    return true;
+  }
```
kou: Should we do this in `func_string_truncate()`?

(Should `max_n_chars > 0` be a caller's responsibility?)

plugins/functions/string.c
```diff
@@ -264,6 +270,136 @@ func_string_substring(grn_ctx *ctx, int n_args, grn_obj **args,
...
+  for (; p < end && (char_size = grn_charlen(ctx, p, end)); p += char_size) {
+    n_chars++;
+    if (n_chars == max_n_body_chars) {
+      *body_end = p + char_size;
+    }
+    if (n_chars > max_n_chars) {
+      return true;
+    }
```
kou, line 290: Can we simplify this?

```suggestion
    } else if (n_chars > max_n_body_chars) {
      return true;
    }
```

plugins/functions/string.c
abetomo: The following check inside func_string_truncate() already covers this case equivalently, so I'll remove it.

```c
  int64_t max_n_body_chars = max_n_chars - n_omission_chars;
  if (max_n_body_chars <= 0) {
```

plugins/functions/string.c
abetomo: Case `string_truncate('target', 8)`

* `max_n_chars` = 8
  * The threshold to truncate the text if it exceeds this number of chars.
* `max_n_body_chars` = 5
  * max_n_chars:8 - omission:3
  * Max number of chars in the text after truncation.
* Since `n_chars` counts the number of chars in `'target'`, the max value is 6.

If we use `n_chars > max_n_body_chars`, it wrongly evaluates to `true` even though no truncation is need.

---

Do you think the variable names are hard to understand?

kou: Yes. Can we use better variable name?

abetomo: I renamed `max_n_body_chars` to `n_body_chars`.
Are there better names for the other variables as well?

kou: Hmm... Do we need to use "body" for kept text...?

`string_truncate_internal()` doesn't truncate. It just detects the end position for kept text and whether the target text should be truncated or not. I feel that the function name doesn't show it.

plugins/functions/string.c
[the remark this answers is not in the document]
kou: We should mention the behavior in the PR description or a comment in source code.

We should document it in our documentation later.

plugins/functions/string.c
```diff
@@ -264,6 +270,146 @@ func_string_substring(grn_ctx *ctx, int n_args, grn_obj **args,
 #undef string_substring_tag
 }
 
+static bool
+string_is_over_n_chars(grn_ctx *ctx, const char *string, size_t string_size,
+                       int64_t max_n_chars, int64_t n_kept_chars, const char **kept_end)
```
kou: Can we use `n_keep_chars` here?

```suggestion
                       int64_t max_n_chars, int64_t n_keep_chars, const char **kept_end)
```

PR #2869, opened 2026-07-28
ci: add "-Wno-error=unused-variable"

komainu8:
Currently, Apache Arrow is built by debug mode to display backtrace when occering error. However, -Werror is specified as g++ flag when using debug mode. As a result, unused-variable is handled as error.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -142,6 +142,7 @@ jobs:
             -S apache-arrow/cpp \
             --preset ninja-debug-minimal \
             -DARROW_COMPUTE=ON \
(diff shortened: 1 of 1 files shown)


kou:
We should fix this in upstream.

kou:
In general, you should share the error log for reference.

komainu8:
> We should fix this in upstream.
> In general, you should share the error log for reference.

Thank you for your comments.
I see. I will report to upstream and I will add error log.

komainu8:
This issue has been already fixed in upstream at https://github.com/apache/arrow/pull/50754.
So, I close this.

PR #2903, opened 2026-09-03
mrb: add `Groonga::ChildContext` and `Groonga::ChildObject`

kou:
`Groonga::ChildContext` pulls a child `grn_ctx` and initializes its mruby.  `Groonga::ChildObject` is a proxy for a value in the child's mruby. We can load code in the child's mruby by
`Groonga::ChildContext#require`, get an object in the child's mruby by `Groonga::ChildContext#[]` and call a method of the object transparently via `Groonga::ChildObject`.

All accesses to the child's mruby are serialized by a per context mutex. This is a preparation for executing a forwarded method call as a `grn::TaskExecutor` task on a worker thread. In this version all calls are run on the caller's thread.

`Groonga::TaskExecutor#execute` runs the given block with a dedicated `Groonga::ChildContext`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -153,10 +153,14 @@ repos:
           ?lib/memory_map\.c|
           ?lib/mrb/mrb_bulk\.c|
           ?lib/mrb/mrb_bulk\.h|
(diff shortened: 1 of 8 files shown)


kou:
I close this in favor of #2904.

PR #2913, opened 2026-09-06
sharding logical_select: process drilldowns in parallel

kou:
Plain drilldown keys and independent labeled drilldowns are processed in parallel with `Groonga::TaskExecutor` when `--n_workers` is 2 or more. A labeled drilldown that depends on another drilldown by `table` is processed after the depended drilldown is finished.

`Groonga::TaskExecutor` changes:

  * `parallel?(n_tasks=nil)` is added. It returns false when `n_tasks` is 1 or less because there is nothing to parallelize.
  * An exception raised in a task uses the rc of its class when it's a `Groonga::GroongaError`. Other exceptions are command errors. This is the same rule as `Groonga::Command#run_internal`. They were unknown errors.

Query log changes:

  * `drilldown(N)` and `drilldowns[LABEL](N)` show their keys like `drilldown(N): KEYS`.
  * The filter operation of a drilldown filter shows its prefix like `drilldowns[LABEL].filter(N): CONDITION`.

Fixed bug:

  * A persistent column specified by `calc_target` was closed after grouping. It's shared with other contexts. It caused a double free in parallel processing.

Tests use `#@sort-query-log-operations shard drilldown` because query log operations of parallel drilldowns aren't ordered. It requires grntest 1.8.7 or later.

### DIFF
--- a/dockerfiles/alpine.dockerfile
+++ b/dockerfiles/alpine.dockerfile
@@ -58,7 +58,7 @@ RUN \
 RUN \
   gem install \
     bundler \
-    "grntest:>=1.8.6" \
+    "grntest:>=1.8.8" \
     groonga-client \
     pkg-config \
     rake
--- a/lib/mrb/mrb_task_executor.cpp
+++ b/lib/mrb/mrb_task_executor.cpp
@@ -148,9 +148,19 @@ namespace {
     return mrb_int_value(mrb, data->executor->get_n_workers());
   }
 
+  /* Groonga::TaskExecutor#parallel?(n_tasks=nil)
+   *
+   * Returns whether tasks are run in parallel. If n_tasks is given,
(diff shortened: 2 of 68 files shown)


kou:
@komainu8 @abetomo This changes query log formats:

* `drilldown(N)` -> `drilldown(N): KEYS`
* `drilldowns[LABEL](N)` -> `drilldowns[LABEL](N): KEYS`
* `filter(N): CONDITION` -> `drilldowns[LABEL].filter(N): CONDITION`

See also: https://github.com/groonga/groonga/pull/2913/changes#diff-fbf4497f150758ac7c3fd28cbdecd8e648596f438eccf31f2d15c62a01f85143

We should mention this in the next release note.

abetomo:
Thanks for the info. I got it.

PR #2914, opened 2026-09-07
doc logical_select: add n_workers

kou:
Shards, plain drilldown keys and independent labeled drilldowns are processed in parallel with n_workers since #2910 and #2913.


kou:
English: https://kou.github.io/groonga/reference/commands/logical_select.html#n-workers
Japanese: https://kou.github.io/groonga/ja/reference/commands/logical_select.html#n-workers

PR #2916, opened 2026-09-07
doc grn_expr_syntax_escape(): don't use backslash in inline code

kou:
Doxygen 1.15.0 or later reports the following warning:

    include/groonga/expr.h:143: warning: The following parameter of grn_expr_syntax_escape(...) is not documented:
      parameter 'escape_character'

Doxygen 1.15.0 or later doesn't close an inline code that has `\"` when there is another inline code that has `\\` in the next `\param` and `EXTRACT_ALL` is `YES`. It's a regression by the fix for https://github.com/doxygen/doxygen/issues/8788 . Doxygen 1.14.0 doesn't report it.


kou:
I'm not sure whether this is a Doxygen bug or Doxygen limitation... I feel that this is a Doxygen limitation...

PR #2924, opened 2026-09-10
doc news: add 16.1.1 entry

abetomo:


komainu8 (approved):

komainu8:
I think we should also include the following changes. What do you think? In particular, since https://github.com/groonga/groonga/commit/e85e2608cd85a07c9fbf38551a0f57876928502c includes changes to the API/ABI, I think it would be worth mentioning that as well.

- https://github.com/groonga/groonga/commit/05f1ebcd1748c628132295228d29e6a411ccf0ea
- https://github.com/groonga/groonga/commit/1032ba8a7f659fba326bb0bc639dd77debfc1608
- https://github.com/groonga/groonga/commit/e85e2608cd85a07c9fbf38551a0f57876928502c

Everything else looks good to me.

kou:
https://github.com/groonga/groonga/pull/2969 restores API/ABI breaking change.

komainu8 (approved):

PR #2927, opened 2026-09-11
select: expose grn_proc_select_create_no_sort_keys_sorted_table()

abetomo:
It creates a table that has only records to be output in the same order as output records.
It's used instead of a sorted table when no sort key is specified.
`logical_select` needs the same table for output stage dynamic columns.

### DIFF
--- a/lib/proc/proc_select.cpp
+++ b/lib/proc/proc_select.cpp
@@ -2463,10 +2463,11 @@ grn_select_create_all_selected_result_table(grn_ctx *ctx, grn_obj *table)
   return result;
 }
 
-static grn_obj *
-grn_select_create_no_sort_keys_sorted_table(grn_ctx *ctx,
-                                            grn_select_data *data,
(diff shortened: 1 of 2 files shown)


lib/proc/proc_select.cpp
```diff
@@ -2463,10 +2463,11 @@ grn_select_create_all_selected_result_table(grn_ctx *ctx, grn_obj *table)
   return result;
 }
 
-static grn_obj *
-grn_select_create_no_sort_keys_sorted_table(grn_ctx *ctx,
-                                            grn_select_data *data,
-                                            grn_obj *table)
+grn_obj *
+grn_proc_select_create_no_sort_keys_sorted_table(grn_ctx *ctx,
```
kou, line 2467: `logical_select`から使うなら`grn_table_*`にしてもっといい感じの名前にしない？

abetomo:
Closing this in favor of #2929.

PR #2955, opened 2026-09-19
json: don't use C++ exception for invalid JSONPath

kou:
Use the std::error_code based jsoncons::jsonpath::make_expression() instead of catching jsoncons::json_exception. This is a preparation for building without C++ exceptions.

jsoncons calls std::terminate() without including `<exception>` when C++ exceptions are disabled. So we include it explicitly.


kou:
Fix in upstream: https://github.com/danielaparker/jsoncons/pull/745

